PR #179, opened 2017-11-13
Additional style options for blockquote component

mikemai2awesome:


mikemai2awesome:
Added invert logo fix, and updated blockquote markdown file. @sghoweri

PR #193, opened 2017-11-30
Feature/bk 66 smooth scroll rebased

remydenton:


remydenton:
One outstanding question on this on my end: should `bolt-navbar` list this as a dependency?

PR #243, opened 2017-12-15
Lower the nav-bar z-index.

charginghawk:
It has to be below the meganav dropdown.


remydenton:
A follow-up to my last comment @mikemai2awesome, Thomas mentioned that `1001` is basically one more than the highest other z-index he found in Bolt.  So, the important part for us is just to keep z-indexes within bolt working relative to each other (ideally using variables) while not going so high as to make overrides for stuff outside of Bolt (e.g. the header navigation) impossible.

remydenton:
Rebased and merged to master.  Opened http://vjira2:8080/browse/BK-98 to follow up on z-index management.

PR #244, opened 2017-12-15
Set smooth scroll offset for nav-bar, body class during scrolling.

charginghawk:
The current method for setting offset (adding class .js-bolt-smooth-scroll-offset to the fixed element) doesn't work for sticky elements. It uses the element's location on the page instead of the element's height when fixed. With a sticky element halfway down the page, it sets the offset to halfway down the page. Instead I'm just setting the offset directly to the nav-bar's height, and we can figure out how to accommodate other sticky elements when we cross that bridge.

This change also sets a class on the body while scrolling, which will allow us to keep the header collapsed. Right now, using the headroom.js, the header is collapsed when scrolling down, but opens when scrolling up, which messes up the offset. Instead of offsetting for the height of the nav-bar, it'd need to offset for the combined height of the header and the navbar. Rather than getting into dynamic offsets for different scenarios, it's easier (and probably more user-friendly) to keep the header collapsed using a scrolling-specific class.

### DIFF
--- a/src/_patterns/02-components/bolt-smooth-scroll/src/smooth-scroll.js
+++ b/src/_patterns/02-components/bolt-smooth-scroll/src/smooth-scroll.js
@@ -2,7 +2,8 @@ import SmoothScroll from 'smooth-scroll';
 
 const scroll = new SmoothScroll();
 
-const defaultScrollOffset = 0;
+// Set offset to 50 for nav-bar.
+const defaultScrollOffset = 50;
 const defaultScrollSpeed = 500;
 
 const customScrollElems = document.querySelectorAll('a[href*="#"]');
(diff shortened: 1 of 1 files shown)


remydenton:
As discussed with @charginghawk:
- Ideally, we don't want to hard-code a non-zero offset in the component-- this should be configurable from the Drupal end.  I'm going to work on a solution that handles this.
- Thomas realized during our call that the body js classes for this won't work as-is (there's some more complexity to get headroom.js to still hide the header when scrolling stops).  He's going to work on setting this up the way he wants in Drupal, and we'll use that as a template for getting the flexible parts back into Bolt.
- In order to more effectively collaborate, we want to have unminified js available in Drupal (in addition to minified js which should be used normally).  I'll address that in #247

remydenton:
Closing this out.  Our conclusion was that Bolt couldn't support the extra stuff the www site needed.  The hope is that once we get the header built out in Bolt, we'll be able to add the necessary support to smooth scroll (and sticky).  For now, Drupal is handling this itself for the www site.

PR #245, opened 2017-12-15
Feature/bk 26 background shapes fixes

remydenton:
- Most importantly, this resolves the errors in Drupal where the svgs could not be found using their relative paths (such a simple fix, doh!)
- It also fixes some variable name typos (also doh!)

@charginghawk @barnettrob, once this is merged, we should be able to update Drupal so that it properly pulls the background shapes pattern from Bolt rather than duplicating it.


barnettrob:
@remydenton tested this on the Drupal side and works!

PR #258, opened 2017-12-19
Fix publicpath() getting called in Drupal

christophersmith262:
This throws warnings in Drupal as we are not currently "officially supporting" publicpath, even though the extension is technically there.

@charginghawk


charginghawk:
See also: http://vjira2:8080/browse/WWWD-864

Issue #309, opened 2018-01-08
Update components to compile using new build tools


EvanLovely:
Each of the components in `packages/components/*` needs to be able to compile using the new build tools.

### What to do for each component

1. In `*.scss` remove any `@import` statement such as `@import "@bolt/settings-all"` and leaving just `@import "@bolt/core"`.
1. Ensure `package.json` has these keys that point at a **single** asset each (all are optional)
    - `style` key pointing to a scss file
    - `main` pointing towards a js files
    - `twig` pointing towards a twig file
1. In `package.json`, adjust any dependencies as you see fit 
    - this is were most troubleshooting happened for me 
    - Sometimes the package version was not exactly what is the repo, so Lerna won't symlink it for you
    - Sometimes the dependency isn't a package in 1.x, but was in 0.x - time to prune
    - Also we're cleaning up any `scripts` related to building the individual components; we do that in bundles now 
1. Add component name to top level package.json under `workspaces` 
1. Add component to Pattern Lab - so if it was the `@bolt/components-button` package and it's version was ~`0.5.1`~ `0.7.0-beta.5`, in the file `apps/pattern-lab--workshop/package.json` add it as a dependency:
    ```diff
   "dependencies": {
    +    "@bolt/components-button": "0.7.0-beta.5"
    }
    ```
1. Run `npm run bootstrap` at root of repo to symlink the new component up properly
1. Add component name to `.boltrc.js` in `apps/pattern-lab--workshop`.
1. Compile by running `npm run build` in `apps/pattern-lab--workshop`.

### List of components to update

- [x] `bolt-button` PR: #312
- [x] `bolt-background-shapes` PR: #316 
- [x] `bolt-background` PR: #320
- [x] `bolt-action-blocks`PR: #320
- [x] `bolt-blockquote` PR: #320
- [x] `bolt-button-group` PR: #320
- [x] `bolt-chip` PR: #320
- [x] `bolt-chip-list` PR: #320
- [x] `bolt-figure` PR: #320
- [x] `bolt-link` PR: #320
- [x] `bolt-ordered-list` PR: #320
- [x] `bolt-unordered-list` PR: #320
- [x] `bolt-teaser` PR: #320
- [x] `bolt-page-footer` PR: #320
- [x] `bolt-page-header` PR: #320
- [x] `bolt-site` PR: #320
- [x] `bolt-card` #343
- [x] `bolt-nav-bar` #346
- [x] `bolt-smooth-scroll` #346
[...]

PR #312, opened 2018-01-09
Update button for new build tools

EvanLovely:
I updated the button! 🎉


packages/components/bolt-button/package.json
```diff
@@ -1,18 +1,8 @@
 {
   "name": "@bolt/components-button",
-  "patternName": "button",
-  "patternType": "components",
-  "patternSubType": "buttons",
```
remydenton: @sghoweri @EvanLovely Are we removing the above lines as well?  I'm not sure what they're meant to do...

remydenton:
I don't see step 5 here from the description in #309:

> 5. Add component to Pattern Lab - so if it was the @bolt/components-button package and it's version was 0.5.1, in the file apps/pattern-lab--workshop/package.json add it as a dependency

Was that unintentionally omitted?  Unnecessary?

packages/components/bolt-button/package.json
EvanLovely: @remydenton We are removing them; they were not used in any way and if we need them, it's easier and more reliable to programmatically figure them out.

EvanLovely:
@remydenton Good catch! Got a PR up for it: https://github.com/bolt-design-system/bolt/pull/315

PR #337, opened 2018-01-16
Re-enabling Strict Schema for Build Tools

EvanLovely:
IMHO, this is really important to keep on. If we don’t, we’ll be lazy with keeping the schema updated. If this build fails, let’s work together on this PR to document the config till it passes.

### DIFF
--- a/apps/pattern-lab--workshop/.boltrc.js
+++ b/apps/pattern-lab--workshop/.boltrc.js
@@ -1,70 +1,12 @@
-// Thanks to StencilJS (https://stenciljs.com/docs/stencil-config) for a little inspiration for some of these configuration names. Naming things is hard!
-
 module.exports = {
   // Environmental variable / preset to use
   env: 'pl',
-
-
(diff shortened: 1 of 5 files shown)


EvanLovely:
OK, I've got all your new config in the schema @sghoweri - the one change I did was from `config.devServer.open` to `config.openServerAtStart` b/c you can't use Env Vars to toggle that (or any config that is nested inside object) (well, we *could*, but converting flat env vars to nested data isn't something I think we should or want to tackle right now). Also, we're using `Object.assign()` to merge and that's shallow... and again, I know we *could* get that in, but that's not something I want to do quite yet.

EvanLovely:
Found out why you were having all those extra items in your config! Turns out that this becomes an EnvVar:

    npm_package_dependencies__bolt_components_button=0.7.0-beta.5

My logic for merging env vars like `bolt_verbosity=5` is now updated to ensure it starts with `bolt_`

PR #346, opened 2018-01-18
#309 update js packages

remydenton:
Please update checkboxes on #309 when merging

Update most of the straightforward packages that run js (those with extra complexity are on separate PRs).

Some of the recurring themes here that deserve a sanity check:
 - Removal of the entire `"scripts": { ... }` and `"devDependencies": {...}` lines in package.json
 - in many cases, the `"main"` key in package.json was previously pointing to the dist folder.  In all cases, I updated it to point to the source js file, though I'm unsure if that was the right thing to do.


EvanLovely:
Looks good to me; let's get @sghoweri review on it before merging though.

PR #348, opened 2018-01-18
Netlify Hosting

EvanLovely:
Couldn't get the Netlify CI to work for use (ImageMagick, see 4bc588e for how I was able to get composer working), so this just deploys to it after Travis builds. At the bottom of the Travis logs is the URL for the build. I've got it so deploys on `epic/refactor` always deploy to http://bolt-design-system.netlify.com and others deploy to URLs like http://5a61317cdf995357b92f3cf3.bolt-design-system.netlify.com - I wasn't able to get the deploy preview to work which results to URLs like http://deploy-preview--feature-netlify.bolt-design-system.netlify.com but oh well.

This also takes care of a good amount of the Travis enhancements concerning Yarn as well.

Also, we can run `npm run deploy` to deploy from local anytime we want as well.


theSadowski:
Does this change what URL we go to view current boot components on netlify?

EvanLovely:
@theSadowski What you've all memorized and know and love ( http://bolt.netlify.com ) won't change, this is actually a 2nd one that we're temporally using ( http://bolt-design-system.netlify.com ) while we work on 1.x, it's mainly just for the devs to make PR reviews easier. Once we launch 1.x, we'll point to http://bolt.netlify.com again

- 0.x : http://bolt.netlify.com
- 1.x : http://bolt-design-system.netlify.com (temp URL)

sghoweri:
@EvanLovely I'd still love to shoot for getting the Slack / Github PR status updates added back in eventually (similar to what we had before in the [1st version](https://github.com/bolt-design-system/bolt/blob/release/0.x/build-tools/gulp-tasks/build-slack/index.js) of our build tools) however this is still a step in the right direction- thanks!!

PR #354, opened 2018-01-19
Feature/build assets tests

EvanLovely:
This enhancement relates to all components that registered in `.boltrc.js` and then looked at to grab their scss, js, and twig files for webpack to compile: now each file reference is checked for it's existence. 

This is a good thing b/c the tests caught two errors we had: 8cef100 & 6ab2d4e

@remydenton Can you confirm that the Bolt Page Header should not have `src/page-header.twig` in it? You can see my change in 6ab2d4e

Most of the changes are renaming `getAssets()` to `getPkgInfo()` and making it so that function returns this format for each package:

```js
const pkgInfo = {
  name: '@bolt/components-button',
  basicName: 'bolt-components-button',
  assets: {
        style: '/Users/Evan/dev/clients/pega-bolt/bolt/packages/components/button/src/button.scss',
        main: '/Users/Evan/dev/clients/pega-bolt/bolt/packages/components/button/src/button.js',
        twig: [
           '/Users/Evan/dev/clients/pega-bolt/bolt/packages/components/button/src/button.twig'
        ]
   }
}
```

Before, `style`, `main`, and `twig` were at the top level and not under `assets` (b/c it was `getAssets()`.

And if you're wondering why it's and absolute path, it's cause that is all in memory and unique to each environment and compile and just makes things generally easier.

### DIFF
--- a/packages/build-tools/utils/assets.js
+++ b/packages/build-tools/utils/assets.js
@@ -1,26 +1,55 @@
+const fs = require('fs');
 const path = require('path');
 const log = require('./log');
+const config = require('./config-store').getConfig();
+
+/**
+ * Ensure a file exists
+ * We don't want to do anything with the file now, we just want to provide an early error if a path is wrong.
+ * This is called async and by the time an error is thrown, we may be several steps ahead with WebPack probably already trying to start - that's ok, we don't want to hold up the process everytime things are correct.
+ * @param filePath {string} - Path to file to ensure exists
+ */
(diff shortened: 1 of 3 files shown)


EvanLovely:
I say we work on adding all components so we know that all can compile, which would complete #309 and then if there are any components that aren't "real" or "ready" or whatever, then we just remove them from `.boltrc.js` in PL

PR #372, opened 2018-01-22
Manifest setup

EvanLovely:
After building, look in `apps/pattern-lab--workshop/www/build/data`.

### DIFF
--- a/apps/pattern-lab--workshop/.boltrc.js
+++ b/apps/pattern-lab--workshop/.boltrc.js
@@ -23,6 +23,7 @@ module.exports = {
       '@bolt/components-action-blocks',
       '@bolt/components-background',
       '@bolt/components-background-shapes',
+      '@bolt/components-band',
       '@bolt/components-blockquote',
       '@bolt/components-button',
       '@bolt/components-button-group',
--- a/apps/pattern-lab--workshop/.boltrc.js
+++ b/apps/pattern-lab--workshop/.boltrc.js
@@ -6,6 +6,16 @@ module.exports = {
   wwwDir: 'www',
   plConfigFile: './config/config.yml',
   verbosity: 1,
+  plTwigNamespaces: {
+    'bolt-assets': {
+      recursive: true,
+      paths: ['www/build'],
+    },
+    utils: {
+      recursive: true,
+      paths: ['src/_includes'],
+    },
+  },
   components: {
     global: [
       '@bolt/core',
@@ -30,7 +40,7 @@ module.exports = {
       '@bolt/components-page-header',
       '@bolt/components-site',
       '@bolt/components-teaser',
-      '@bolt/components-unordered-list'
+      '@bolt/components-unordered-list',
     ],
     individual: [
     ],
--- a/packages/build-tools/cli.js
+++ b/packages/build-tools/cli.js
@@ -2,9 +2,6 @@ const path = require('path');
 const program = require('commander');
 const packageJson = require('./package.json');
 const configStore = require('./utils/config-store');
-const log = require('./utils/log');
-
-log.intro();
 
 // global `bolt` cli options & meta
 program
@@ -12,10 +9,13 @@ program
   .option('-C, --config-file <path>', 'Pass in a specific config file instead of default of ".boltrc.js/json".')
   .option('-v, --verbosity <amount>', 'How much output do you want? 0-5', parseInt);
 
-
-// defaults to `.boltrc` so either `.boltrc.js` or `.boltrc.json` works
+// We need to initialize config as early as possible
 const configFilePath = path.resolve(process.cwd(), program.configFile || '.boltrc');
 configStore.init(require(configFilePath));
+// Now that config is initilized, we can start requiring other things
+
+const { buildBoltManifest } = require('./utils/manifest');
+const log = require('./utils/log');
 
 /**
  * Update config with all options flags
@@ -43,10 +43,21 @@ function updateConfig(options, programInstance) {
 
   const config = configStore.getConfig();
   log.dim(`Verbosity: ${config.verbosity}`);
-  log.dim(`Opening browser: ${config.openServerAtStart}`);
-  log.dim(`Quick mode: ${config.quick}`);
+  if (config.verbosity > 2){
+    log.dim(`Opening browser: ${config.openServerAtStart}`);
+    log.dim(`Quick mode: ${config.quick}`);
+    log.dim(`buildDir: ${config.buildDir}`);
+    log.dim(`dataDir: ${config.dataDir}`);
+    log.dim(`wwwDir: ${config.wwwDir}`);
+  }
+
+  // Basically at this point, the cli is bootstrapped and ready to go.
+  // Let's build the core bolt manifest
+  buildBoltManifest();
   return config;
 }
+
+log.intro();
 
 // `bolt build`
 program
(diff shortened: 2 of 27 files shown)


packages/build-tools/tasks/internal-tasks.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Makes all directories in config
+ * @async
+ * @returns {Promise}
+ */
+async function mkDirs() {
+  try {
+    return Promise.all([
+      mkdirp(config.wwwDir),
```
sghoweri: Thoughts on killing off PL doing this for us (via a patch maybe?)

apps/pattern-lab--workshop/.boltrc.js
```diff
@@ -6,13 +6,24 @@ module.exports = {
   wwwDir: 'www',
   plConfigFile: './config/config.yml',
   verbosity: 1,
+  plTwigNamespaces: {
```
sghoweri: This smells like a Pattern Lab-ism 😉

Why not call this "namespaces" or "twigNamespaces" and call it a day?

packages/build-tools/cli.js
```diff
@@ -43,11 +43,22 @@ function updateConfig(options, programInstance) {
...
+  // Basically at this point, the cli is bootstrapped and ready to go.
+  // Let's build the core bolt manifest
+  buildBoltManifest();
   return config;
 }
 
+log.intro();
+
 // `bolt build`
 program
```
sghoweri: Thoughts on (eventually) moving each of these commands into a separate file in a `/commands` folder or something?

I'm still mulling on this - I just feel like a repeating recipe for something like this starts to feel like something you'd piece together via includes...

packages/build-tools/tasks/pattern-lab-tasks.js
```diff
@@ -37,6 +35,65 @@ const plPublic = path.join(plRoot, plConfig.publicDir);
...
+    });
+  });
+
+  const namespaceConfigFile = Object.assign({
+    // Can hit anything with `@bolt`
+    bolt: {
+      recursive: true,
+      paths: [
+        plSource,
+        ...allDirs,
```
sghoweri: +1

packages/build-tools/tasks/pattern-lab-tasks.js
```diff
@@ -2,21 +2,20 @@ const chalk = require('chalk');
...
 const del = require('del');
 const debounce = require('lodash.debounce');
 const log = require('../utils/log');
 const { getConfig } = require('../utils/config-store');
 const ora = require('ora');
+const {getBoltManifest} = require('../utils/manifest');
 
 const config = Object.assign({
   plConfigFile: 'config/config.yml',
-  watchedPkgExtensions: [
```
sghoweri: Q. How does this handle watching Twig PHP extensions?

packages/build-tools/utils/manifest.js
```diff
@@ -0,0 +1,53 @@
...
+const writeFile = promisify(fs.writeFile);
+const path = require('path');
+const config = require('./config-store').getConfig();
+
+let boltManifest = {
+  name: 'Bolt Manifest',
+  components: {
+    global: {
+      src: [],
+      dist: {},
```
sghoweri: Dist?

packages/build-tools/utils/sh.js
```diff
@@ -31,7 +31,7 @@ async function sh(cmd, exitOnError, streamOutput) {
       if (code > 0) {
         if (exitOnError) {
             process.exitCode = 1;
-            reject(new Error(`Error with code ${code} after running: ${cmd}`));
+            reject(new Error(`Error with code ${code} after running: ${cmd}\n ${output}`));
```
sghoweri: +1

packages/twig-extensions/filters/custom-filters/webpack.filter.php
```diff
@@ -1,13 +1,13 @@
 <?php
 
-$filter = new Twig_SimpleFilter('webpack', function ($name, $manifestFile = '/dist/bolt-manifest.json', $removeQueryString = true) {
-  
+$filter = new Twig_SimpleFilter('webpack', function ($name, $manifestFile = '/www/build/bolt-webpack-manifest.json', $removeQueryString = true) {
```
sghoweri: TODO: not have this rely on a hard coded path like this, ESPECIALLY given the number of times we've had to update this path + this default wouldn't necessarily work in any other environment or config other than our PL setup...

packages/components/color-swatch/package.json
```diff
@@ -6,5 +6,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "style": "_color-swatch.scss"
-}
\ No newline at end of file
+  "style": "_color-swatch.scss",
```
sghoweri: todo: underscores vs no underscores... and if the answer is both, when do we stick with one over the orher

packages/components/bolt-band/package.json
```diff
@@ -1,17 +1,17 @@
 {
   "name": "@bolt/components-band",
+  "version": "0.6.0-beta.5",
```
sghoweri: @EvanLovely are bands showing up / compiling with these updates?

packages/build-tools/utils/manifest.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    individual: {
+      src: [],
+      dist: {},
+    },
+  },
+};
+
+function buildBoltManifest() {
+  if (config.components.global) {
```
sghoweri: Remind me again what the difference is between individual vs global here? I get what the theoretical difference is but in practice, what's the difference here?

The duplicate logic with just the key name differing... *shrug*

packages/build-tools/utils/manifest.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return boltManifest;
+}
+
+function getBoltManifest() {
+  return boltManifest;
+}
+
+async function writeBoltManifest() {
```
sghoweri: Magic string? Where else are we referencing bolt-full-manifest?

packages/build-tools/tasks/pattern-lab-tasks.js
```diff
@@ -37,6 +35,65 @@ const plPublic = path.join(plRoot, plConfig.publicDir);
...
+      recursive: true,
+      paths: [
+        plSource,
+        ...allDirs,
+      ],
+    }
+  }, namespaces, config.plTwigNamespaces || {});
+
+  await writeFile(
+    path.join(config.dataDir, 'twig-namespaces.json'),
```
sghoweri: I'd suggest we stick to using `bolt-` filename prefixes for consistency

EvanLovely: done

packages/build-tools/utils/manifest.js
EvanLovely: Was thinking that it could hold the info that the WebPack manifest has now: what these all compiled to.

packages/build-tools/utils/manifest.js
EvanLovely: All global components turn into `bolt-global.css` and `bolt-global.js`, each component in `individual` gets it's own `PKG-NAME.js` and `PKG-NAME.css` (`@bolt/button` makes `bolt-button.css`).

EvanLovely: Useful for if/when Drupal themes want to `attach_libraries()` for individual components; like sliders.

packages/build-tools/utils/manifest.js
EvanLovely: Yep; for now. Referencing in PL Twig files.

packages/components/bolt-band/package.json
EvanLovely: Yep! But now it's this:

screenshot_2018-01-22_05-54-29_pm

sghoweri:
Works great locally - thanks @EvanLovely!!

PR #376, opened 2018-01-24
BK-149 video player close button

remydenton:


remydenton:
I'm going to merge this myself now, but @sghoweri, you should review when you get back.  It doesn't feel like the most robust way of handling this, but the best I could do under the circumstances.

PR #384, opened 2018-01-29
Refactoring pattern_template Twig function 

EvanLovely:
This removes need for old Bolt Manifest - temp workaround till templates can be refactored to not use approach like this:

```twig
{% include pattern_template('teaser') %}
```

An advantage of not refactoring Twig templates like this:

```diff
- {% include pattern_template('teaser') %}
+ {% include "@bolt/teaser.twig" %}
```

is that we'll be able to update 1.x templates from 0.x by simply copying them in. We'll want to do that refactor after we are well into 1.x

/cc @sghoweri


christophersmith262:
@remydenton @sghoweri @EvanLovely @charginghawk 

Can we talk about this one?  It's fine if we remove `pattern_template`, but I want to make sure we have an alternative way to solve the problems it was introduced to solve (chiefly, allowing Drupal to alter composed components).

Do we have a consolidated list of things like this in 1.0?  I'd like to give a once-over to let you guys know what might be a breaking change from a Drupal perspective.

EvanLovely:
@christophersmith262 Sure thing! Are you referring to integration with the UI Patterns Drupal module? Mind giving me a little background? 

Also, it'd be easy to get this going again as we've got a much better and more robust approach to Manifests that have been introduced specifically to help integration into different environments; Drupal being chief among them.

christophersmith262:
Sure @EvanLovely, it has to do with the UI patterns implementation:

Take the teaser component for example...  We currently have an adapter twig template sitting in front of the bolt twig that lets us inject extra content:
* http://vstash:7990/projects/PM/repos/pega_bolt/browse/features/pega_bolt_teaser/templates/teaser-adapter.twig?at=feature/product-pages
* http://vstash:7990/projects/PM/repos/pega_bolt/browse/features/pega_bolt_teaser/pega_bolt_teaser.ui_patterns.yml?at=feature/product-pages#4

The teaser gets composed inside other components (cards, bands, etc.) within bolt.  Bolt calling `pattern_template` allows it to render / extend the Drupal adapter instead of only dealing directly with bolt twig files.

I use this example because it was used to solve a common workflow issue, where a particular feature (the inclusion of metadata in this case) could not be completed in time within the design system, so we had to cobble it together in Drupal.

While all options would ideally be directly implemented in the design system, it hasn't always been feasible, and having the ability to override composed patterns has given us the tools to keep the lights going.

I'm open to alternatives, but I'd like to still have the ability to override bolt twig from a practical standpoint.

EvanLovely:
So in a nutshell you want to be able to override a particular template, kinda like taking a Twig file out of Drupal core and placing it in your theme?

christophersmith262:
I think that's a fairly good analogy, but to me it's more akin to how you would override a base theme's template with a sub-theme template that extends the original, rather than forking and hacking.

christophersmith262:
Maybe a good alternative would be to create a twig extension that registers the `@bolt` namespace to dynamically resolve the twig file rather than using https://www.drupal.org/project/components and forcing bolt to implement a drupalism?

remydenton:
Thanks for flagging this @christophersmith262.  I agree it's helpful to have a clean way for Drupal to extend a Bolt pattern-- could be used to inject all sorts of Drupalisms like contextual links, for example.

Can you explain more what you mean by "a twig extension that registers the @bolt namespace"?  Isn't that pretty similar to what `pattern_template()` is doing?

christophersmith262:
@remydenton I mean we could create a twig plugin like:
http://cgit.drupalcode.org/components/tree/src/Template/Loader/ComponentLibraryLoader.php?id=57259ca#n20

That way drupal and bolt can compose things by using standard `{% include '@bolt/...'; }` and in the background, we can have the twig plugin dynamically lookup where `@bolt/...` should resolve to.  If there is a template override in drupal, we would have it resolve there otherwise we would resolve it to what's in `node_modules`.

I'm not even sure if this is possible, just an idea.

EvanLovely:
We should talk out what the process looks like for a dev to override/extend a Bolt template, but know that I feel that all the tools are present in 1.x to do this flexibly. 

I don’t think using this syntax: 

```
{% include pattern_template('teaser') %}
```

over this syntax:

```
{% include "@bolt/teaser.twig" %}
```

is the right way to do it though. So I propose we merge this PR in and start up a new issue that outlines the use case & requirements so we can discuss there.

krlucas:
@EvanLovely That's reasonable (but I defer to Charginghawk, Remy and Chris) but let's first create the follow-up issue and make sure it's marked as a 1.x release blocker, then merge the PR. Also can we just mark `pattern_template()` as deprecated instead of removing it altogether? Bolt 0.x code is in prod on www now, and will soon be on PDN; documenting the ways 1.x will break that code is important to ever getting it integrated.

(I hope I don't need to do this in github but I will do it anyway @remydenton @christophersmith262 @charginghawk @sghoweri)

EvanLovely:
Got the issue created here: #386 - @christophersmith262 can you fill out the use case on that ticket please? Feel free to edit the post.

> Also can we just mark pattern_template() as deprecated instead of removing it altogether?

This PR doesn't remove `pattern_template()` - it just simplifies it's logic. I originally proposed planning to remove it in the future, but since you all have a need for it; we can ensure it works. 

What was happening before is that it used the old Bolt Manifest to resolve a short name (like `button_group`) to a path with a Twig Namespace (`@bolt/button-group.twig`). I found what was using it and simplified it to a simply `switch` statement. It'd be easy to integrate it with the new Manifest if we believe this to be the solution. I feel that our solution to the challenge mentioned doesn't use `pattern_template()`, but involves the Manifest and how we register the Twig Namespaces paths.

Also, documenting the upgrade path from 0.x to 1.x is something we're aware we need to do.

krlucas:
Awesome. Thanks @EvanLovely!

PR #390, opened 2018-01-30
Foundation for Image Build Tasks

EvanLovely:
This sets up the very basics of the image tasks for #364 No resizing/minification nor watch. Does have support for size suffixes. When prod it copies (will resize/min later), and when not on prod it symlinks.


EvanLovely:
Discussed this with @sghoweri whom gave it his 👌

PR #391, opened 2018-01-31
Resizing images when prod

EvanLovely:
OK, so `sharp` is awesome! This PR will resize the images when it's a production build. I opened all the files it resized and ran it through my own copy of ImageOptim and got an average savings of 3.5%, so it's looking like it's doing some nice minification for us.

What do we do when an image is smaller than the size we are resizing to? So when the original is 700px wide, and we're dealing with the 1024px size step. Do we not make it? Do we make a 700px wide image with a `-1024` filename suffix?

Also: this sets the ground for removing the need for people to install ImageMagick!


EvanLovely:
For #364

PR #397, opened 2018-02-01
refactor: Updating breadcrumb pattern to not have nested patterns

remydenton:
A tweak to the breadcrumbs component that allows breadcrumb content to be flexible.

The goal is to make the breadcrumbs pattern only concerned with the layout and separators of items and agnostic to what those items actually are.  This means:
  - A consumer could potentially have each breadcrumb be something besides links, such as unlinked text or even icons
  - A consumer like Drupal can potentially render its own links without using this pattern to do so.


EvanLovely:
👍

PR #410, opened 2018-02-07
Fix for css class names with escaped characters

EvanLovely:
OK, this should fix the situation where css classes like `.u-hide\@large` get turned into `.u-hide-large` until [the bug in `css-loader`](https://github.com/webpack-contrib/css-loader/issues/578) can get fixed. /cc @sghoweri 

Jake, can you check to see if this fixes the issue, then merge if it does?

### DIFF
--- a/packages/build-tools/create-webpack-config.js
+++ b/packages/build-tools/create-webpack-config.js
@@ -97,36 +97,37 @@ function createConfig(config) {
   }
 
 
-  // @todo Remove workaround for getting classes with `\@` to compile correctly
+  // This workaround has been disabled for now as setting `modules: false` on `css-loader` fixes it; see https://github.com/bolt-design-system/bolt/pull/410
+  // Workaround for getting classes with `\@` to compile correctly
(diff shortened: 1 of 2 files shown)


EvanLovely:
Well, crap; getting this build error:

```
Error: /home/travis/build/bolt-design-system/bolt/packages/global/styles/index.scss
Module build failed: ModuleBuildError: Module build failed: Unknown word (1:1)
> 1 | -theAtSymbol-charset "UTF-8";
    | ^
  2 | @font-face {
  3 |   font-family: OpenSansSubset;
  4 |   src: url(/fonts/opensans-subset.woff2) format("woff2"),url(/fonts/opensans-subset.woff) format("woff");
```

My initial tests were just on my small use case, not on the whole thing.

EvanLovely:
I've done some more research on the core issue and posted some findings in comments there. Also, I've found that disabling CSS Modules with `modules: false` also gets the behavior we want w/o this workaround. Are we using them?

PR #421, opened 2018-02-09
Color Swatch - moved assets over to PL utils and out of components

rockymountainhigh1943:
@sghoweri -- We decided to move color swatches into the PL utils since it's really not a component and makes more sense to live there. Have a look and if you're good with it - merge away and we can cross another one off!


rockymountainhigh1943:
Hi @sghoweri -- the color swatches are still used within visual styles > color palette, we've just moved the color swatch twig template into utils since it isn't really a reusable component; only a utility offered by the system.

It only makes sense then that the sass should get pulled back into PL leaving us with an empty package. This pointing to your other thought about unpublishing the package -- I'm all for lean and mean so if we don't need it lets clean house. 

Let me know what you think when you have a sec.

Thanks!
Jake

rockymountainhigh1943:
Hi @sghoweri -- this one should be all set based on your feedback from the other day. Please have a quick look and merge it in if we're solid.

Thanks!
Jake

PR #424, opened 2018-02-09
Twig Extension Migration

EvanLovely:
Let's move all the Twig Extension from `_twig-components` into the proper Twig Extensions found in `packages/core-php/src/TwigExtensions/`. All can be placed in *either* the `BoltCore.php` or `BoltExtras.php` files found in that directory. We're done when the build stops failing.


sghoweri:
@EvanLovely there's a couple things to chat about (and related to) this work*, but for the most part, Pattern Lab is now fully switched over to using the new Class-based extensions architecture** and is compiling without any errors - huzzah!!

I freakin LOVE how much cleaner and much more manageable this new setup is netting out. Having the ability to have reusable functions / helper utils across your Twig extension logic ROCKS! 🤘


*Things to chat / follow up on: `autoload.php` for class name resolution and the hairier parts of PHP Class namespaces, getting a little help to get some of the manifest file path resolution logic rolled into some of these, questions about the current state of the low quality base64 placeholder logic + image srcset logic in PL, the ratio object appearing to be missing in a few places, and the Bolt Icon component names being hard coded >_<

**There's a couple extensions I've kept around that I suspect we _might_ need for a few missing features (like globbing files on the file system) -- want to double check on these first before removing.

EvanLovely:
So this error:

```
PHP Warning:  require(/home/travis/build/bolt-design-system/bolt/packages/core-php/src/../vendor/autoload.php): failed to open stream: No such file or directory in /home/travis/build/bolt-design-system/bolt/packages/core-php/src/TwigFunctions.php on line 5
```

was caused by the presence of `require __DIR__ . '/../vendor/autoload.php';` in a package that is symlinked (due to being a custom `path` repo in composer.json `repositories` in PL. The folder `core-php/vendor` doesn't exist. I initially moved `\Bolt\Layout` from psr-0 to psr-4 and also removed those `require` statements. I'm getting a new error now:

```
Uncaught Error: Class 'Bolt\Drupal\Core\Template\Attribute' not found in /home/travis/build/bolt-design-system/bolt/packages/core-php/src/TwigFunctions.php:17
```

I see those are also psr-0, I'm wondering if moving them to psr-4 will help.

EvanLovely:
Wahoo! We got it!

EvanLovely:
You'll see on my commit b646bec (Adding specific version number to core-php so composer.lock doesn't have the current git branch as version number) that I've added this to `core-php/composer.json`:

```diff
+"version": "0.9.0",
```

The reason is b/c `composer.lock` would have the version number as the current branch name... which, well... yeah: you get it.

screenshot_2018-02-12_02-40-51_pm

Just wanted to call this out so you were aware.

EvanLovely:
> how does PL know about our local extension in this new setup?

Via [`twigExtensions` in PL's `config.yml`](https://github.com/bolt-design-system/bolt/blob/epic/refactor/apps/pattern-lab--workshop/config/config.yml#L52-L55) - and since those packages are `require`-ed in `composer.json`, it can call that public static function and get back a Twig Extension.

sghoweri:
Right right right - was just confirming that https://github.com/bolt-design-system/bolt/blob/011ee14a017b360624986e6bc4b5300249c36c1f/apps/pattern-lab--workshop/composer.json#L39 also still existed so the symlink in the vendor folder was still happening. 

The updated config.yml in PL ties in these new extensions being symlinked in the monorepo...

EvanLovely:
Yep, we're still symlinked!

EvanLovely:
I'm merging this in as the remaining questions we have are all outside the scope of this PR.

PR #427, opened 2018-02-10
headline schema

joekarasek:
I worked backwards from the code to figure out all the parameters, converted the demos to iterate over the schema, created a reference to the icon schema, found one more variation to demo (quoted headlines), and added schema validation to the twig template. There are errors generated from the schema validation, but these seem to be due to internal mechanism of the component and I didn't want to change anything in headline yet.


joekarasek:
This ticket is partially blocked by pr 425. The proper handling of true and false in the schema breaks the schema-doc twig template, but that is being updated in https://github.com/bolt-design-system/bolt/pull/425

PR #431, opened 2018-02-13
WIP logo schema

joekarasek:
This was an interesting one...

With other schema, parameters for included components are wrapped in their own object. For example, the block quote includes an image. The data passed to blockquote is structured thus...
```
text: "This is a quote.",
image: {
    src: "http://google.com",
    alt: "this is an image"
}
```

The logo component seems to act more like an extension of the image component. It inherits the properties of image.
```
invert: true,
src: "http://google.com",
alt: "this is an image"
```

I was able to get this to work from a json schema validation standpoint, but I don't think our current doc generation will work with this format. I'm also thinking that changing the logo component to behave in a way that is consistent with the other components would be a good ticket for a future release.

Thoughts?


sghoweri:
@joekarasek @EvanLovely I don't want this to die on the vine here -- what's the next step to getting this out the door (besides fixing the merge conflict ;) )?

sghoweri:
@joekarasek @rockymountainhigh1943 @EvanLovely @mikemai2awesome @remydenton @theSadowski just a friendly reminder to test your code before putting in a PR (even if it's just a docs-specific or PL specific change) -- pretty please :)

For example, forgetting to include the `only` attribute in your Twig include in Pattern Lab can break a bunch of stuff if we're not careful (and checking our work):

image

PR #465, opened 2018-02-21
pre publish tests

EvanLovely:
This branch is based off same branch used for https://github.com/bolt-design-system/bolt/pull/464 so please merge that first. This adds `npm run` commands to top level:

- `npm run build` - runs a prod build for each folder in `apps/` - this ensures a compile works and will expose any errors from compiles (and linting in the future)
- `npm run composer` - this goes into each `apps/` directory and `rm -rf composer.lock vendor` and then runs `compose install`. This is triggered as part of `npm run setup`

Additionally, the `npm run build` is ran before `npm run publish` to provide testing assurances, and before `npm run deploy` to ensure a site is a production build before deploying as non-prod builds use symlinks for images that don't work in a deploy.

Also, I gotta say: blank lines between hunks of related scripts works well for readability! If you're curious why I went series instead of parallel, it's all about readability in the terminal.

### DIFF
--- a/apps/bolt-site--storefront/package.json
+++ b/apps/bolt-site--storefront/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "bolt build",
     "build:noisy": "bolt build --verbosity 5",
+    "build:prod": "bolt build --prod",
     "serve": "bolt serve",
     "watch": "bolt watch",
     "start": "bolt start"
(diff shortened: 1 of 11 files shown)


EvanLovely:
Oh, also `npm start` outputs this:

```
To start up a server and watches, head into any of the folders in the apps/ directory, then run npm start there.
If you want to start Pattern Lab, run this:

cd apps/pattern-lab--workshop
npm start
```

PR #474, opened 2018-02-22
Feature/testing

joekarasek:
Alright, this went smoothly except I had a few small issues around testing the writing of files (namely, clean up). I tried to used the built in methods with jest to manage this and I got the right effects (files were cleaned up after testing, tests passed) but it caused the overall test to exit with an error. Still looking into this.


joekarasek:
Issue was resolved with 'decrapated unresolved promise' node warning by using async/await.

This work is being moved to a future PR with testing for both js and phpunit.

PR #475, opened 2018-02-23
migration of button

joekarasek:
Looks like the javascript for the button had a call to a bolt utility that's no longer in 1.x.

Module not found: Can't resolve '@bolt/utilities-visuallyhidden/_utilities.visuallyhidden.scss' in '/Users/karasek/Dev/work/bolt-1.x/packages/components/button/src/button.standalone.js' Line 13


joekarasek:
@EvanLovely Thanks!

PR #496, opened 2018-03-02
Shadow

joekarasek:
Basic functionality is there. Could use a once over on refactors as well as documentation on the bolt docs site.

Did we want utility classes for this one? I made them just in case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 php:
-  - 7.0
+  - 7.1
 
 addons:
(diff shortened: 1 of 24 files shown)


joekarasek:
@theSadowski 
Check this out... https://5a999339efbe5d46e64228f7--bolt-design-system.netlify.com/pattern-lab/?p=styleguide-shadow
This is all dynamically generated from one settings file, so it'll be easy to add, remove, edit, and rename these.

PR #503, opened 2018-03-06
PHP Package Publishing

EvanLovely:
Triggering read-only repo pushes and providing some test content.


EvanLovely:
OK, so that worked - the read only repo was not updated. Merging.

PR #506, opened 2018-03-06
Manifest Improvements

EvanLovely:
Can now access schemas with this:

```twig
{% set schema = bolt.data.components['@bolt-components-action-blocks'].schema %}
```

instead of:

```twig
{% set schema = get_data('@bolt-components-action-blocks/action-blocks.schema.yml') %}
```

So this should **drastically** reduce the times we access the file system and do Yaml parsing.

I've updated the Action Blocks to use the new method only: 136959a


EvanLovely:
Don't merge yet; I want to look at the deploy first.

EvanLovely:
OK, this is looking good! Let's discuss soon @sghoweri

PR #530, opened 2018-03-19
Decreasing the level of shadow on form inputs from 30 to 20

mikemai2awesome:
The shadow is HUGE! This decreases it down to a similar level as what the design mockup had.


remydenton:
@mikemai2awesome, as a general rule, include devs who have worked or are working on the code you're modifying as reviewers.  I had actually independently made this exact same change as part of BK-313-- it would have saved me that time if I'd seen this.

PR #531, opened 2018-03-20
Feature/bk 313 form styling updates

remydenton:
See http://vjira2:8080/browse/BK-313

@mikemai2awesome, most of this work is updates to CSS that you originally wrote.  Can you code review to make sure it all makes sense in the context of your original work?

@sghoweri, the one request from the original ticket I couldn't figure out was the following.  Could you elaborate a little (could be on a call)?
> Floating label size is correct, but the default size should be the one that displays when a form input is active, not an already scaled size


remydenton:
One other question for @mikemai2awesome -- I was unsure how best to create inline form elements.  I have a shameful workaround that can be seen in [the inline form demo](https://5ab148cbdf99532bf7bac8e5--bolt-design-system.netlify.com/pattern-lab/?p=components-form-full-inline).  Any thoughts about how to do that better?

sghoweri:
@remydenton were you going to be updating the form full inline demo per our chat earlier in the week? Alternatively perhaps just disable that demo in PL for now (add a dash before the filename) and we handle that separately?

image

sghoweri:
Another question for @remydenton, @theSadowski and @mikemai2awesome actually - how does someone know if an input field is or isn't required?

Input Not Required:
input-not-required

Input Required:
input-required

remydenton:
Yes, my plan is to update the inline demo based on our discussion @sghoweri.  It'll probably be next week before I have time to wrap that up though.

Re: indication for optional form elements, here's the design spec: http://vjira2:8080/browse/WWWD-1236.  You, @theSadowski, and I had previously decided to handle that on the consumer end (Drupal) rather than with logic in Bolt.

remydenton:
@sghoweri, this is now stable and ready to merge if all looks good.  I've got a couple separate follow-on issues that will have their own PRs, if necessary.

PR #533, opened 2018-03-20
Block List, Tooltip, and Copy to Clipboard Components

rockymountainhigh1943:
This PR contains:
- Completed Block List Component
- Completed Tooltip Component
- Completed Copy to Clipboard component
- WIP Share Component (only the base, nothing to get to crazy over yet)


apps/pattern-lab/src/_patterns/02-components/share/00-share-docs.twig
```diff
@@ -0,0 +1,18 @@
+{% set schema = get_data('@bolt-components-share/share.schema.yml') %}
+{% set code %}
+  {% verbatim %}
+    {% include '@bolt-components-share/share.twig' with {
+      sources: ['facebook','twitter','linkedin'],
+      email: true,
+      copy: true,
+      visibility: 'visible',
```
sghoweri: What does the `visibility` param do here? What happens if that isn't passed in?

apps/bolt-site/content/20-docs/40-guides/15-Adding-a-new-component.md
```diff
@@ -0,0 +1,128 @@
...
+Below you'll find a step-by-step guide to walk through the process of adding a new component to the Bolt ecosystem.
+
+**<u>Step 1</u>:**  Create a new dir under `/packages/components/[bolt-NEW-COMPONENT]`
+
+**<u>Step 2</u>:** `cd` into the new dir and create a new `package.json` in the new dir:
+```
+{
+  "name": "@bolt/[NEW-COMPONENT]",
+  "description": "BLANK Component in Bolt",
+  "version": "1.0.0-rc.3",
```
sghoweri: Update version to "0.0.0" (which also means the PL package.json file reference under dependencies needs to be pointing at v0.0.0 as well).

Think about it like this: the initial version we are adding hasn't yet been published so it technically can't have a version number (version number = version someone should expect to see available on NPM). 

Also keep in mind that Lerna's publish command will go in and bump it to the latest version being published so we don't need to thrash around trying to make sure the version we're adding is actually the one we need!

apps/bolt-site/content/20-docs/40-guides/15-Adding-a-new-component.md
```diff
@@ -0,0 +1,128 @@
...
+    "type": "git",
+    "url": "https://github.com/bolt-design-system/bolt/"
+  },
+  "bugs": {
+    "url": "https://github.com/bolt-design-system/bolt/issues"
+  },
+  "style": "src/[COMPONENT].scss",
+  "main": "src/[COMPONENT].js",
+  "twig": "src/[COMPONENT].twig",
+  "publishConfig": {
```
sghoweri: FYI this publishConfig option here is SUPER important! 

This is what tells NPM that our package is OK to publish as a public (not as a private) NPM organization (private orga require a paid account). Without this, the entire publish script would fail!!**

**Then you'd have to go in, fix the package.json file(s), republish with new version #s, and have to live with yourself knowing that v1.x.y exists out of the wild that nobody can use! 😊

apps/bolt-site/content/20-docs/40-guides/15-Adding-a-new-component.md
```diff
@@ -0,0 +1,128 @@
...
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/bolt-design-system/bolt/"
+  },
+  "bugs": {
+    "url": "https://github.com/bolt-design-system/bolt/issues"
+  },
+  "style": "src/[COMPONENT].scss",
+  "main": "src/[COMPONENT].js",
```
sghoweri: Q. Is it worth mentioning to omit any of these if your new component doesn't require them?

apps/bolt-site/content/20-docs/40-guides/15-Adding-a-new-component.md
```diff
@@ -0,0 +1,128 @@
...
+
+**<u>Step 1</u>:**  Create a new dir under `/packages/components/[bolt-NEW-COMPONENT]`
+
+**<u>Step 2</u>:** `cd` into the new dir and create a new `package.json` in the new dir:
+```
+{
+  "name": "@bolt/[NEW-COMPONENT]",
+  "description": "BLANK Component in Bolt",
+  "version": "1.0.0-rc.3",
+  "homepage": "https://boltdesignsystem.com",
```
sghoweri: Technically, yes, it's supposed to be `https://boltdesignsystem.com` however at the moment we're having to use my backup `https://bolt-design-system.com` domain since the DNS for the non-dashed one still needs to be updated to point to Netlify, otherwise this URL would be broken. I'm thinking update it to use the dashed version till then (then we can just to a global find and replace for this?)

CC @theSadowski

apps/bolt-site/content/20-docs/40-guides/15-Adding-a-new-component.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Be sure to follow the general coding style and practices of the Bolt design system (review other components for reference)
+
+**<u>Step 5</u>:** Add a new `README.md` at the root of your component, next to `package.json` (with pertinent details)
+
+- A description of the component's functionality should reside at the very top of the readme
+- Be sure to review an existing readme and include the "Install via NPM" instructions
+- Consider adding a "description" section as well as a "best practices" bullet list
+
+**<u>Step 6</u>:** Add component to `workspaces` array within the `package.json` located at the root of the repo
```
sghoweri: > **<u>Step 6</u>:** Add component to `workspaces` array within the `package.json` located at the root of the repo

"In the package.json file in the root of the Bolt repo, add the new component's folder path to the `workspaces` key."

How does ^ sound?

apps/pattern-lab/src/_patterns/02-components/block-list/00-block-list-docs.twig
```diff
@@ -0,0 +1,41 @@
+{% set schema = get_data('@bolt-components-block-list/block-list.schema.yml') %}
+{% set code %}{% verbatim %}
+{% include "@bolt-components-block-list/block-list.twig" with {
+  "contentItems" : [
```
sghoweri: Thoughts on us ripping the bandage off and switching this over to either be a generic `children` variable (following the React / Preact / VueJS naming convention) or `items` if we're explicitly talking about an array of things that we need to iterate over.

@remydenton @rockymountainhigh1943 @joekarasek @EvanLovely thoughts?

apps/pattern-lab/src/_patterns/02-components/share/05-share.twig
```diff
@@ -0,0 +1,7 @@
+{% include '@bolt-components-share/share.twig' with {
+  sources: ['facebook','twitter','linkedin'],
+  email: true,
+  copy: true,
+  visibility: 'visible',
+  inline: false
```
sghoweri: Question on the default for this: if by default we don't want this to be inline (which off the top of my head, we don't), can we flip this and by default not require the inline param and then only require it when we DO want this to be an inline layout... make sense?

apps/bolt-site/templates/site.scss
```diff
@@ -193,6 +193,12 @@ $storefront-sidebar-width: 320px;
   padding: 0 0 0 0;
   transition: margin .4s;
 
+  a {
+    border-bottom: 1px dotted;
+    color: bolt-color(indigo);
+    font-weight: 600;
```
sghoweri: FYI we should use the `@bolt-font-weight(semibold|bold|regular)` mixin since that directly ties in the available font stacks in the system with the weights we're using in our Sass

apps/bolt-site/templates/site.scss
```diff
@@ -193,6 +193,12 @@ $storefront-sidebar-width: 320px;
   padding: 0 0 0 0;
   transition: margin .4s;
 
+  a {
```
sghoweri: Two quick things on this:

1. What was this change specifically meant for just so I can better understand what we're trying to do here and if there's a different way we can go about it?
2. Did you notice the bunch of visual regressions this change introduced largely because of the element selector (why we try to use class selectors for everything that's not a _very_ light handed element default)

Action Blocks on Docs Homepage Color Changed:
image

Double Underlines on Homepage "View Source" Link:
image

Manually setting a hard-coded color on a link breaks the theming system:
image

One other separate thing (probably due to when this branch was made since we had to revert this on master): the regression relating to the headline schema updates causing "headlines" to output as "text" components instead is causing the margin bottom on all headlines to be missing:
image

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.js
```diff
@@ -0,0 +1,30 @@
+import ClipboardJS from 'clipboard';
+
+let theClipRef = document.querySelector('.c-bolt-copy-to-clipboard__default .c-bolt-link');
```
sghoweri: For any non-Web Component JS in Bolt that need to target bits of the DOM, please make sure we're only targeting using `js-bolt` classes like you're doing down below.

- [ ] Update selectors to reference `js-bolt` classes

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.js
```diff
@@ -0,0 +1,30 @@
...
+    event.preventDefault(); // Prevent the default link behavior
+
+    let clip = new ClipboardJS('.c-bolt-copy-to-clipboard__default .c-bolt-link');
+
+    /**
+     * [1] Adds a class onClick after successful copy and enables the first set of animations
+     * [2] Waits until the first set of animations complete and adds the last class for last animations
+     **/
+    clip.on("success", function() {
+      document.querySelector('.js-bolt-copy-to-clipboard').classList.add('copied'); // [1]
```
sghoweri: - [ ] Please  make sure to us `is-` prefixes on class names that get toggled via JS. Ex. `is-copied`, `is-transitioning`, etc

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.js
```diff
@@ -0,0 +1,30 @@
+import ClipboardJS from 'clipboard';
+
+let theClipRef = document.querySelector('.c-bolt-copy-to-clipboard__default .c-bolt-link');
+let theClipCopied = document.querySelector('.c-bolt-copy-to-clipboard__copied .c-bolt-link');
```
sghoweri: - [ ] Please make sure we're handling situations where more than one copy to clipboard instance exists on the page (ex. a page where a user could share something on the top and on the bottom of the page) - otherwise this would only select the very first of these found on the page (one of the freebies we get by using Web Components since each instance is self contained and gets its own lifecycle of events when getting added / removed)

If we're not using a WC here, make sure to do something similar to how we had to do the smooth scroll querySelectorAll forEach loop: https://github.com/bolt-design-system/bolt/blob/master/packages/components/bolt-smooth-scroll/src/smooth-scroll.js
- [ ] Update query selector to only target `js-` selectors

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.js
```diff
@@ -0,0 +1,30 @@
...
+  theClipRef.addEventListener("click", function(event) {
+    event.preventDefault(); // Prevent the default link behavior
+
+    let clip = new ClipboardJS('.c-bolt-copy-to-clipboard__default .c-bolt-link');
+
+    /**
+     * [1] Adds a class onClick after successful copy and enables the first set of animations
+     * [2] Waits until the first set of animations complete and adds the last class for last animations
+     **/
+    clip.on("success", function() {
```
sghoweri: - [ ] Single quotes, not double quotes

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.js
```diff
@@ -0,0 +1,30 @@
+import ClipboardJS from 'clipboard';
+
+let theClipRef = document.querySelector('.c-bolt-copy-to-clipboard__default .c-bolt-link');
+let theClipCopied = document.querySelector('.c-bolt-copy-to-clipboard__copied .c-bolt-link');
+
+if (theClipRef !== null) {
+  theClipRef.addEventListener("click", function(event) {
```
sghoweri: - [ ] Single quotes not double quotes

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.js
```diff
@@ -0,0 +1,30 @@
...
+    clip.on("success", function() {
+      document.querySelector('.js-bolt-copy-to-clipboard').classList.add('copied'); // [1]
+      setTimeout(function(){ // [2]
+         document.querySelector('.js-bolt-copy-to-clipboard').classList.add('transitioning');
+      }, 2000);
+    });
+  }, false);
+}
+
+// We don't want the other link clickable either
```
sghoweri: The other link clickable? What other link? I might be missing something here...

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.scss
```diff
@@ -0,0 +1,67 @@
+/* ------------------------------------ *\
+   Copy to Clipboard
+\* ------------------------------------ */
+
+@import '@bolt/core';
+
+.js-bolt-copy-to-clipboard {
```
sghoweri: - [ ] As a rule of thumb, please don't use `js-` class selectors in our Sass files.

`c-bolt-` class prefix = Component class in a `.scss` file
`js-bolt-` JS selector in a `.js` file

Harry Roberts has a good article running through a lot of the core naming convention concepts that have heavily influenced how the design system's CSS Architecture tackles naming things.
https://csswizardry.com/2015/08/bemit-taking-the-bem-naming-convention-a-step-further/

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.scss
```diff
@@ -0,0 +1,67 @@
...
+  display: inline-block;
+  position: relative;
+
+  .c-bolt-copy-to-clipboard__icon,
```
sghoweri: - [ ] Why are these CSS selectors nested? Using BEM virtually eliminates the need for this in most instances! ;)

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.scss
```diff
@@ -0,0 +1,67 @@
...
+    position: absolute;
+    top: 0;
+    left: 0;
+
+    .c-bolt-link__text {
+      color: bolt-color(teal, dark);
+    }
+  }
+
+  &.copied {
```
sghoweri: - [ ] See previous comment on using `is-` Sass selectors if we need styling based on JS state changes

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.scss
```diff
@@ -0,0 +1,67 @@
...
+      top: 0;
+      bottom: 0;
+      margin: auto;
+    }
+  }
+
+  &.copied.transitioning {
+    .c-bolt-copy-to-clipboard__icon {
+      opacity: 0;
+      transition: opacity 1s ease;
```
sghoweri: Q. Are we using our new transition / animation values that @theSadowski had been putting together?

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.twig
```diff
@@ -0,0 +1,50 @@
...
+{% set attributes = create_attribute(attributes|default({})) %}
+
+{% if text %}
+  {% set attributes = attributes.setAttribute("text", text) %}
+{% else %}
+  {% set attributes = attributes.setAttribute("text", "Copy Link") %}
+  {% set text = "Copy Link" %}
+{% endif %}
+
+{% if copiedText %}
```
sghoweri: So if copiedText isn't defined, we want to set a default copied-text value, yes?

Any reason why you didn't use the `default` Twig filter to save a couple lines of code?

```
{# How are we handling translation with this? #}
{% set defaultCopiedText = "Copied" %}

{% set copiedText = copiedText | default(defaultCopiedText) %}
{% set attributes = attributes.setAttribute("copied-text", copiedText) %}
```

sghoweri: ^ Same approach could be applied with any other instances where we want to set a default value but also handle more specific overrides when a component is getting pulled in.

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.twig
```diff
@@ -0,0 +1,50 @@
...
+{% else %}
+  {% set attributes = attributes.setAttribute("copied-text", "Copied!") %}
+  {% set copiedText = "Copied!" %}
+{% endif %}
+
+{% if url %}
+  {% set attributes = attributes.setAttribute("url", url) %}
+{% endif %}
+
+<bolt-{{ componentName }} bolt-component class="c-bolt-copy-to-clipboard js-bolt-copy-to-clipboard" {{ attributes }}>
```
sghoweri: - [ ] Web Component `<custom-element>` but non-Web Component JS?

packages/components/bolt-share/package.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "style": "src/share.scss",
+  "main": "src/share.js",
+  "twig": "src/share.twig",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@bolt/core": "^1.0.0-rc.9",
+    "@bolt/components-link": "^1.0.0-rc.9",
```
sghoweri: +1 good catch to include pulling this in as a dependency

packages/components/bolt-share/src/share.standalone.js
```diff
@@ -0,0 +1 @@
+// console.log('Share');
```
sghoweri: - [ ] Make sure to remove once Share component work is wrapped up :)

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+   Tooltip
+\* ------------------------------------ */
+@import '@bolt/core';
+
+$bolt-tooltip-bubble-triangle-width: 8px;
+$bolt-tooltip-bubble-border-color: bolt-color(gray, light);
+$bolt-tooltip-transition: $bolt-transition;
+
+// Aria tags for accessibility.
+[aria-hidden="true"] {
```
sghoweri: - [ ] Please please please make sure to update these two `[aria-]` Sass selectors to only directly target this component (ideally the thing we actually want to be hidden and nothing more) - the early version of this added to Drupal ended up causing the exact same visual regressions to the video component that I'm seeing on this staged Netlify instance:
https://5ab36c7548769b0e32c5c376--bolt-design-system.netlify.com/pattern-lab/?p=viewall-components-video

Note the video play button icon is missing with this Sass as is:
image

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+  opacity: 0;
+  pointer-events: none;
+}
+
+[aria-hidden="false"] {
+  opacity: 1;
+  pointer-events: auto;
+}
+
+@keyframes bounceDown {
```
sghoweri: - [ ] Question: is this Sass something we could (or rather, should) reuse in other places? If not, definitly keep it here, otherwise I'd recommend we add this up to the generic ITCSS level for reuse across multiple components.

CC @mikemai2awesome @theSadowski @joekarasek @EvanLovely @remydenton

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+
+.c-bolt-tooltip--action {
+  .c-bolt-tooltip__trigger {
+    cursor: pointer;
+    user-select: none;
+  }
+}
+
+
+// Spacing options
```
sghoweri: I'm ok lettings these go for not but this is the sort of thing where ideally we'd set the size of something at the top level of a component and have the different sub-parts separately worry about adding modifier classes as needed.
 It's also one of the few use cases where I feel having nested BEM &-- modifier selectors make a lot more sense to keep things coupled together.

Ex.
```
.c-bolt-tooltip__content-bubble {
    ...

    &--xsmall {

   }
   
    &--small {

   }
}
```

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+    overflow: hidden;
+
+    &:focus {
+      outline: none;
+    }
+  }
+
+  bolt-icon {
+    padding-right: 5px;
+    width: auto; // IE & Firefox fix
```
sghoweri: But only in the context of this component??

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+    height: auto; // IE & Firefox fix
+  }
+
+  button .toggle--closed,
```
sghoweri: - [ ] Q. do we really need a vanilla `button` element selector here? Can't we use a class for this?

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+
+.c-bolt-tooltip__content-bubble {
+  @include bolt-padding(xsmall, squished);
+  @include bolt-shadow('level-20');
+
+  display: block;
+  position: relative;
+  width: 300px;
+  border: 1px solid $bolt-tooltip-bubble-border-color;
+  border-radius: 3px;
```
sghoweri: - [ ] Swap for global bolt border radius variable

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+  }
+}
+
+.c-bolt-tooltip__content-bubble {
+  @include bolt-padding(xsmall, squished);
+  @include bolt-shadow('level-20');
+
+  display: block;
+  position: relative;
+  width: 300px;
```
sghoweri: Q. why 300px? Where's this magic number coming from?

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+    content: '';
+    pointer-events: none;
+  }
+}
+
+
+// Show and hide effects
+.c-bolt-tooltip:not(:hover) > .c-bolt-tooltip__content,
+.c-bolt-tooltip__trigger + .c-bolt-tooltip__content {
+  @extend [aria-hidden="true"];
```
sghoweri: See comment above on these. 

- [ ] I'd suggest just make the original `[aria-hidden="true"]` and `[aria-hidden="false"]` selectors to be mixins that get used here and not output as standalone attribute selectors

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+}
+
+.c-bolt-tooltip:hover > .c-bolt-tooltip__content,
+.c-bolt-tooltip__trigger.is-active + .c-bolt-tooltip__content {
+  @extend [aria-hidden="false"];
```
sghoweri: Same as with https://github.com/bolt-design-system/bolt/pull/533/files#r176771933

packages/components/bolt-tooltip/src/tooltip.standalone.js
```diff
@@ -0,0 +1,182 @@
...
+            type={data.triggerType}
+            transform={data.triggerTransform}
+            icon={data.triggerIconName}
+            size={data.triggerIconSize}
+            toggle-text={data.triggerToggleText}
+            toggle-icon={data.triggerToggleIcon}
+            trigger-id={data.triggerId}
+          />
+          <TooltipContent type={data.triggerType} trigger-id={data.triggerId}>
+            <span dangerouslySetInnerHTML={{__html: data.content}} />
```
sghoweri: What's going on here? Why are we having to use `dangerouslySetInnerHTML`?

packages/components/bolt-tooltip/src/tooltip.standalone.js
```diff
@@ -0,0 +1,182 @@
...
+} from '@bolt/core';
+
+import styles from './tooltip.scss';
+import button from '@bolt/components-button/src/button.scss';
+
+@define
+export class BoltTooltip extends withPreact(withComponent()) {
+  static is = 'bolt-tooltip';
+
+  static props = {
```
sghoweri: @rockymountainhigh1943 @mikemai2awesome @theSadowski @remydenton @joekarasek @EvanLovely 

I'm still wrapping my head around this Tooltip component but my gut tells me that there's probably either way too many available options for configuring this from a content author's POV and/or things that we should be handling internally / automatically to not make someone think (a good example that comes to mind is `triggerId` since that could just be a randomly generated UUID I think)

packages/components/bolt-tooltip/src/tooltip.standalone.js
```diff
@@ -0,0 +1,182 @@
...
+            {data.text}
+          </span>
+          }
+        </span>
+      </span>
+    );
+  }
+}
+
+const TooltipContent = (props) => {
```
sghoweri: 💯

packages/components/bolt-tooltip/tooltip.schema.yml
```diff
@@ -0,0 +1,89 @@
...
+      - action
+      - help
+  position:
+    type: string
+    description: Should tooltip push up or down?
+    default: down
+    enum:
+      - up
+      - down
+  noWrap:
```
sghoweri: Q. is noWrap really a necessary option? Do we have examples of using / not using this?

CC @mikemai2awesome @rockymountainhigh1943

apps/pattern-lab/src/_patterns/02-components/share/00-share-docs.twig
rockymountainhigh1943: This relates to the use case outlined in the original ticket where on the platform it's being implemented on, it might be hidden until scroll happens. In this case the `visibility` param would be set to `scroll`. The other use case, which has not been fully vetted yet, is to set this to `timed` where after a specific amount of time (undecided yet) the component would fade in.

We can rip this out if you feel the component should always be available and the previously above discussed functionality would be on the implementor to handle all aspects.

apps/bolt-site/content/20-docs/40-guides/15-Adding-a-new-component.md
rockymountainhigh1943: Yes! Added :-)

apps/bolt-site/content/20-docs/40-guides/15-Adding-a-new-component.md
rockymountainhigh1943: Sounds good 👍

apps/bolt-site/content/20-docs/40-guides/15-Adding-a-new-component.md
rockymountainhigh1943: Excellent! Replacing what I have with this!

apps/pattern-lab/src/_patterns/02-components/share/05-share.twig
rockymountainhigh1943: Yup! I'll remove that param from the docs example as well as the demo.

apps/bolt-site/templates/site.scss
rockymountainhigh1943: Thanks for catching this! I was intending to style all links within the documentation section and failed to see this was effecting the homepage. I will do better in checking this should I need to make any such adjustments in the future. At the time of adding those rules the links were very bland and I wanted to spice things up a bit (Emeril Lagasse style BANG!)

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.js
rockymountainhigh1943: We discussed this earlier -- provides a fallback if JS craps the bed haha

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.scss
rockymountainhigh1943: Sir, yes sir! Was moving fast and missed that. Will keep an extra eye on that moving forward! :-D

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.scss
rockymountainhigh1943: Got me there!

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.scss
rockymountainhigh1943: I have not seen these new transition animation values yet. Have they been merged in yet?

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.twig
rockymountainhigh1943: Think of it like a developer easter egg! Was just trying to keep you on your toes! ;-)

packages/components/bolt-copy-to-clipboard/src/copy-to-clipboard.twig
rockymountainhigh1943: Hoping we can chat about this one on Monday.

packages/components/bolt-tooltip/src/tooltip.scss
rockymountainhigh1943: Great catch!! I grabbed a good portion of this SCSS from Mike's demo and should have noticed that sooner. Thank you!

packages/components/bolt-tooltip/src/tooltip.scss
rockymountainhigh1943: My personal opinion is that this is a custom use case and I am not aware of a situation this would be used elsewhere. But -- if anyone else has any feelings about it I'll be happy to move it.

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -0,0 +1,472 @@
...
+.c-bolt-tooltip--help {
+  cursor: help;
+
+  .c-bolt-tooltip__trigger button {
+    cursor: help;
+  }
+}
+
+.c-bolt-tooltip--action {
+  .c-bolt-tooltip__trigger {
```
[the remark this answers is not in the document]
rockymountainhigh1943: Yes, I would say so. The "--action" class is a top level class and we are only wanting to target the trigger within. (If we want to run through any of these on screen share Monday I am totally down with that).

packages/components/bolt-tooltip/src/tooltip.scss
rockymountainhigh1943: I agree. We are for sure going to do more work on this component in the very near future and we should totally refactor some of this for that.

packages/components/bolt-tooltip/src/tooltip.scss
rockymountainhigh1943: I just did a lot of testing in Firefox and it only appears to be an issue in context of this component. I time boxed attempting to see what was specifically happening with this context and couldn't rule out anything specific. Would you prefer I keep digging into this more?

packages/components/bolt-tooltip/src/tooltip.scss
rockymountainhigh1943: Good question! I grabbed this from the example provided -- but four share component I would need even narrower. Should we add a param for this? Something like `width` and go with `small`, `medium`, `large`?

packages/components/bolt-tooltip/tooltip.schema.yml
rockymountainhigh1943: This is for when the tooltip only shows one line of text and does not need a "full bubble"

packages/components/bolt-tooltip/src/tooltip.standalone.js
rockymountainhigh1943: 👍

apps/pattern-lab/src/_patterns/02-components/block-list/00-block-list-docs.twig
remydenton: I'm a little late to the game here, but I think changing `contentItems` to `Items` actually isn't ripping off a band aid at all.  `contentItems` comes with a lot of baggage from previous patterns about what format it should actually take-- see [ui-list](https://github.com/bolt-design-system/bolt/blob/master/packages/global/styles/05-objects/objects-ui-list/src/ui-list.twig) for one example (is it an array of objects with `text` keys? an array of objects with `pattern` keys? who knows...).

This pattern handles it the way we should do it going forward, which is accepting an array of renderable items-- simple and flexible.  If we could rename it `items`, I think that would help re-enforce this and speed the deprecation of the old `contentItems` pattern.

packages/components/bolt-tooltip/src/tooltip.standalone.js
rockymountainhigh1943: Removed some options

packages/components/bolt-block-list/src/block-list.twig
```diff
@@ -0,0 +1,28 @@
...
+
+{% set classes = [
+  baseClass
+] %}
+
+<bolt-{{ componentName }} bolt-component>
+  <ul {{ attributes.addClass(classes) }}>
+    {% for listItem in contentItems %}
+      <li class="{{ "#{baseClass}__item" }}">
+        {% if listItem.text %}
```
remydenton: Is this necessary?  Can't we just assume that listItem is renderable (i.e. either text or something that is otherwise ready to print, like a pattern)?

apps/pattern-lab/src/_patterns/02-components/block-list/00-block-list-docs.twig
charginghawk: I should note that this is tangential to one source of friction in Bolt integration. In Drupal, there's a strong separation between entities and their fields, most particularly in that they always have different templates. Any iteration, anything using an array of "items", happens in the field sub-template.

Bolt doesn't have this separation of components and fields - there's a lot of iterating over items in the component template (where in Drupal that'd only happen in a sub-template). This makes it difficult to map components to data entities on a 1-to-1 basis, because it incurs some minimal preprocess logic per component that builds up over time as we add more.

However, with regards to this discussion, I agree with @remydenton's assertion about how that iteration should look, wherever it is. Inconsistency re: item iteration within Bolt is a bigger source of friction than where that iteration happens.

sghoweri: @remydenton @charginghawk I had a couple ideas on what ^ might look like. I'm gonna push to have us tackle that separately since we need to get this mega PR merged in but I definitely want to make sure we get these pain points addressed across the board

#consistancy

PR #580, opened 2018-04-04
Fixing Silent Errors

EvanLovely:
This is addressing a few things that create silent errors (things that do not exit non-0, so Travis does not fail, nor do the tools show errors).

- Invalid JSON Schemas. This introduces validation of those via the JSON Schema Schema :)
- Invalid YAML. Adding Yaml Linting.
- Array to String notices that happen when one uses an array as a var in PL. I have added an error handler in PL for this.

Next up is fixing the invalid schemas:

- [x] `block-list.schema.yml`
- [x] `button.schema.yml`
- [x] `device-viewer.schema.yml`
- [x] `headline.schema.yml`

### DIFF
--- a/packages/components/bolt-button/src/button.twig
+++ b/packages/components/bolt-button/src/button.twig
@@ -16,23 +16,21 @@
 {% endif %}
 
 {# set up psuedo self-validation by limiting param values to what's specifically allowed in the component schema #}
-{% set alignOptions = schema.properties.align.enum|remove_null %}
+{% set alignOptions = schema.properties.align.enum %}
 {% set roundedOptions = schema.properties.rounded.enum %}
-{% set sizeOptions = schema.properties.size.enum|remove_null %}
-{% set styleOptions = schema.properties.style.enum|remove_null %}
-{% set tagOptions = schema.properties.tag.enum|remove_null %}
-{% set transformOptions = schema.properties.transform.enum|remove_null %}
-{% set widthOptions = schema.properties.width.enum|remove_null %}
+{% set sizeOptions = schema.properties.size.enum %}
+{% set styleOptions = schema.properties.style.enum %}
+{% set tagOptions = schema.properties.tag.enum %}
+{% set transformOptions = schema.properties.transform.enum %}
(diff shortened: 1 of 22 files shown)


EvanLovely:
I'm seeing this mistake in a few places:

```yml
var:
  type:
    anyOf:
      - string
      - object
```

When it should be:

```yml
var:
  anyOf:
    - type: string
    - type: object
```

The `- type: string` can be a little deceiving in Yaml, as that translates to:

```json
{
  "var": {
    "anyOf": [
      {
        "type": "string"
      },
      {
        "type": "object"
      }
    ]
  }
}
```

`anyOf` (or `oneOf`) is an array of whole schemas.

EvanLovely:
Sweet; the PL error handler that catches the Array to String conversion will output the twig file and line number it is happening at!!

EvanLovely:
The other schemas are a little tricky as they all involve `null`. We can't have it be an option in `enum` as it is not a string:

screenshot_2018-04-04_04-15-09_pm

[source](https://spacetelescope.github.io/understanding-json-schema/reference/generic.html#enumerated-values)

packages/components/bolt-button/src/button.twig
```diff
@@ -16,23 +16,21 @@
...
+{% set sizeOptions = schema.properties.size.enum %}
+{% set styleOptions = schema.properties.style.enum %}
+{% set tagOptions = schema.properties.tag.enum %}
+{% set transformOptions = schema.properties.transform.enum %}
+{% set widthOptions = schema.properties.width.enum %}
 
 {# check if the value set to a prop is allowed or defined. if not, default to the default value specified in the component's schema (if one exists) #}
 {% set align =  align in alignOptions ? align : schema.properties.align.default %}
 {% set iconOnly = iconOnly | default(false) %}
-{% set rounded = rounded in roundedOptions ? rounded : schema.properties.rounded.default %}
```
joekarasek: rounded is only used once to define classes, and at that point it is tested against roundedOptions. This is therefore redundant. Also, if the default is undefined, the default should not be defined and called (thats the pattern for the rest of the schema).

PR #589, opened 2018-04-06
Bolt Site Mobile Menu Fix & Schema Table Android Scroll Fix

rockymountainhigh1943:
**This PR contains two important fixes:**
- Bolt Site > Removed unnecessary mobile menu item
- Schema Table > Fixed issue where schema table was not scrollable on Android


apps/bolt-site/templates/site.twig
```diff
@@ -70,14 +69,6 @@
         <a href="/pattern-lab" target="_blank">Pattern Lab</a>
       {% endif %}
     {% endfor %}
-
-    <a href="#menu" class="storefront-docs__hamburger u-bolt-hide@medium">
```
sghoweri: @rockymountainhigh1943 A for effort, but I gotta ask: have you checked out the site on your phone after making this update? 

I totally agree with removing this on smaller screens while viewing the homepage buuuut, the docs site on any other page is totally unusable as a result of the off-canvas menu getting axed:

b6766669-2c84-4c19-a003-306fb7e68e71

rockymountainhigh1943: @sghoweri -- I absolutely tested this on my Android and both iPhoneX and iPhone8. There is no issues with mobile menus on any of these devices:

image

Can you send me an example of the issue you're talking about? I don't see any problem with any other page's navigation at mobile, tablet, or desktop.

Thanks dude!

rockymountainhigh1943: Ok, so after looking at the doc site -- I don't see the docs navigation that was on the left being able to be triggered anymore.. whoops!! 👎

rockymountainhigh1943: @sghoweri -- this has now been fixed and I added a few more adjustments that should really make this sing on mobile.

Check it out when you can and if it's good lets merge! :-)

PR #598, opened 2018-04-09
Bolt Dropdown Cleanup

joekarasek:
@sghoweri Two quick things...
1) There were two copies of the guts of bolt-dropdown, one in the root of the component and one in a `src` directory. I removed the src directory as it appeared to be the older of the two (and had unresolved merge conflicts in it).
2) Webpack failed on dropdown.scss due to a custom property having `!important` on it. Best I can understand it, this important gets striped out anyway. https://www.w3.org/TR/css-variables-1/#syntax


joekarasek:
@sghoweri Just saw that you already fixed the css var in the previous still open PR

PR #602, opened 2018-04-11
WIP Dropdown Refactor

joekarasek:
1) Remove duplicate demo files from PL, rename and reorder
2) Fix example from docs so that it works
3)


sghoweri:
@joekarasek woah - hold up! 

Why are we removing all the custom element demos here? How would we know if the `<bolt-dropdown>` component works through a vanilla Custom Element tag (not getting included via a Twig template directly)

sghoweri:
I'm not saying I don't disagree with the PR here - that's just a super important use case I want to make sure we are actively checking for moving forward (even if we go about it in a different way like two Twig includes -- one via custom element tag and another which pulls in the guts)

joekarasek:
@sghoweri 
That makes sense. Let's go ahead and clearly document both strategies, and then demo both strategies. 
Regardless, we shouldn't have templates that compete for namespace in PL. `01-dropdown-custom-element--collapse--xdark.twig` and `45-dropdown-custom-element--collapse--xdark.twig` are in the PL, but only one of them actually renders. Same with most of the patterns.

joekarasek:
image

However, I don't like that the nav component is used and referenced in the dropdown docs, but there is zero public facing documentation on what bolt-nav is and how to use it... If I were a user who had no knowledge of the inner workings, I would assume this is either a typo or that the docs are out of date.

sghoweri:
@joekarasek I think I need to think this through a bit (both in terms of docs but also regarding nested dependencies of Twig templates and how we can prevent gotchas like this)

FWIW, I'm liberally borrowing some of the ideas the React Bootstrap project is using (as far as component breakdown and composition is concerned):

```
 <Navbar>
  <Navbar.Header>
    <Navbar.Brand>
      <a href="#home">React-Bootstrap</a>
    </Navbar.Brand>
  </Navbar.Header>
  <Nav>
    <NavItem eventKey={1} href="#">
      Link
    </NavItem>
    <NavItem eventKey={2} href="#">
      Link
    </NavItem>
    <NavDropdown eventKey={3} title="Dropdown" id="basic-nav-dropdown">
      <MenuItem eventKey={3.1}>Action</MenuItem>
      <MenuItem eventKey={3.2}>Another action</MenuItem>
      <MenuItem eventKey={3.3}>Something else here</MenuItem>
      <MenuItem divider />
      <MenuItem eventKey={3.4}>Separated link</MenuItem>
    </NavDropdown>
  </Nav>
</Navbar>;
```

PR #609, opened 2018-04-12
Fix "Call to undefined function Bolt\Layout\createAttribute()" error.

charginghawk:


EvanLovely:
Thanks! Got it merged and just released v1.1.3 with it!

PR #639, opened 2018-04-24
Update sticky component to support advanced use cases.

charginghawk:


charginghawk:
Closed in favor of #644.

PR #663, opened 2018-04-30
Hotfix/firefox rendering fractional pixels

joekarasek:
See http://vjira2:8080/browse/WWWD-1936
A small white line is appearing between divs at specific viewport widths (dependent upon screen resolution). This is an issue with how firefox renders fractional pixels caused by using rem units (TL:DR, they need to be half a pixel closer to each other so no gap ever appears).

@sghoweri @mikemai2awesome I wasn't sure where the most prudent place to put this fix was, so I put it on the full bleed mixin.


sghoweri:
image

sghoweri:
image

sghoweri:
Edit: this is definitely a strange rendering quick in Firefox related to position sticky + having the band component fill the viewport width + `transform: translate3d(x, y, z)` (also `transform: translateZ()`)...

Updating this hack to `margin-top: -0.5px;` and `margin-bottom: -0.5px;` does seem to do the trick -- just seeing if there's any way we can limit the scope in any way before merging this in.

PR #683, opened 2018-05-08
Close Button on Firefox

joekarasek:
Very small icon on firefox would not fire event upon being clicked. Adding point-event:none addresses this.
In hotfix for immediate release in PR https://github.com/bolt-design-system/bolt/pull/685


EvanLovely:
Merged already via #685 and then merging that branch back into master.

PR #701, opened 2018-05-16
Fix: themed card to have solid colored background

mikemai2awesome:
This enables individually themed cards to have solid colored backgrounds, no transparency.


remydenton:
Wow, very clever solution @mikemai2awesome.

To summarize for anyone else who's curious:

- If you set a theme on the card, the background will be opaque and will solve the original issue.  However, if you _don't_ set a theme on the card, you can still get a slightly transparent background as shown in the last example at (because the color/opacity is applied to `-c-bolt-card::before`) https://bolt-design-system-emmjqwcren.now.sh/pattern-lab/?p=components-card-band-variation
- The card background colors are close (not exact) to what they were previously because there's now a mostly transparent white background on `-c-bolt-card::after`.  Without that, the colors appear much more saturated than they were previously.

PR #709, opened 2018-05-23
Update share inline component 

theSadowski:
Update share inline component with "share this page" text, color change to dark gray, icon color to indigo, light instead of social brand colors.


mikemai2awesome:
The color and text changes are good to go.

PR #714, opened 2018-05-25
Fix smoothscroll flash in IE.

charginghawk:
This fixes a flashing issue in IE (see WWWD-2151) by instantiating SmoothScroll once instead of attaching a SmoothScroll to every anchor link. In the process it simplifies the code more towards [recommended usage](https://github.com/cferdinandi/smooth-scroll#3-initialize-smooth-scroll)

This negates the ability to set offset on a per-link basis. We weren't using this on the Drupal-side anyway, but we may have to at some point if the navbar height varies. This fix will get us through our next major release, in any case.

Also, this will break the navbar demo within Pattern Lab, eg:

http://localhost:3000/pattern-lab/?p=components-sticky-with-content-example

But the demo will work when viewed on its own page, eg:

http://localhost:3000/pattern-lab/patterns/02-components-sticky--15-sticky-with-content-example/02-components-sticky--15-sticky-with-content-example.html#section-1

Presumably due to a conflict with how Pattern Lab handles anchor links.

### DIFF
--- a/packages/components/bolt-smooth-scroll/src/smooth-scroll.js
+++ b/packages/components/bolt-smooth-scroll/src/smooth-scroll.js
@@ -1,44 +1,21 @@
 import SmoothScroll from 'smooth-scroll';
 
-const scroll = new SmoothScroll();
-
+// In the future, we could add support for links to modify options like scrollOffset, scrollOffset, etc.  However,
+// we should provide options carefully-- only enable these after considering whether the use case that requires them
+// is justified. In particular, this assumes a static navbar height, which is the case is practice, but subject to change.
 const defaultScrollOffset = 45;
 const defaultScrollSpeed = 750;
 
-const customScrollElems = document.querySelectorAll('a[href^="#"]');
-for (var i = 0, len = customScrollElems.length; i < len; i++) {
-  const scrollElem = customScrollElems[i];
-  const scrollSpeed = defaultScrollSpeed;
+var scroll = new SmoothScroll('a[href^="#"]', {
+  ignore: '[data-scroll-ignore]', // Selector for links to ignore (must be a valid CSS selector)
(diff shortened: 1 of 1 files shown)


charginghawk:
I'll take a minute to explain the navbar offsetting ecosystem for posterity.

There's three offsets: the gumshoe offset, the smoothscroll offset, and the navbar offset.

In Drupal, these offsets have to take into account the fixed main header, and then the gumshoe and smoothscroll offsets have to also account for the height of the navbar itself.

The navbar offset works using the css `top`, as is required by `position: sticky` and the polyfill we're using. We add the Drupal-side `top` in the Drupal styling.

The smoothscroll offset uses a selector for fixed headers, `js-bolt-smooth-scroll-offset`, which is on the fixed main header, and a hardcoded default offset set to 45px, the general navbar height.

The gumshoe offset uses the offset attribute on <bolt-nav>, which in Drupal is set to 165px.

remydenton:
Thanks summarizing that @charginghawk-- we should definitely have a unified strategy there.

One drawback to `boltSettings` is that there'd be no way to update it on the fly-- say, if you need to change the offset depending on whether the header has squished down, the navbar is overlaying things, etc.  Is that mostly unimportant?

charginghawk:
Before, the fixed main header was using headroom.js to appear and disappear (and on most pages it still uses headroom.js to grow and shrink), but with Bolt v1.x we disabled that on pages with the navbar to smooth out all this offsetting, so now for navbar purposes the header height is static.

boltSettings probably wouldn't work if you had fixed elements that changed height, so that offsets had to change. It also probably wouldn't work if you needed varying offsets - say you had multiple sticky elements of different height on the same page.

I'd say we should push back on designs that have these requirements. The navbar suffers chronically from too much complexity.

PR #717, opened 2018-05-30
Video Enhancement - Collapse video when clicking away

rockymountainhigh1943:
This works with multiple videos on the page and is cross-browser / device tested.

**Testing Instructions**
1. Boot up PL
1. Navigate to http://localhost:3000/pattern-lab/patterns/02-components-video-video--background-video-with-inline-script-and-external-controls/02-components-video-video--background-video-with-inline-script-and-external-controls.html
1. Click the button to launch video
1. Click anywhere outside bolt-video to see the video close

### DIFF
--- a/packages/components/bolt-video/src/video.standalone.js
+++ b/packages/components/bolt-video/src/video.standalone.js
@@ -96,6 +96,8 @@ class BoltVideo extends withPreact(withComponent()) {
     // This binding is necessary to make `this` work in the callback
     this.handleClose = this.handleClose.bind(this);
 
+    this.collapseOnClickAway = this.collapseOnClickAway.bind(this);
+
     // BoltVideo.globalErrors.forEach(this.props.onError);
 
     this.defaultProps = {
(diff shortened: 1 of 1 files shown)


packages/components/bolt-video/src/video.standalone.js
```diff
@@ -354,13 +356,25 @@ class BoltVideo extends withPreact(withComponent()) {
...
+    }
+  }
 
   _onWindowResize(event) {
     this._calculateIdealVideoSize();
   }
 
+  // If we click outside bolt-video we collapse
+  collapseOnClickAway(event) {
+    const videoTag = document.querySelector('bolt-video');
```
remydenton: Here's what this code is saying:
1) Get [_the first_](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) `<bolt-video>` element on the page and save it as `videoTag`
2) If the collapseOnClickAway event is not happening on `videoTag`, close the video

That'll result in some interesting bugs-- for example, if the background video in question is not the first `<bolt-video>` element on the page, clicking directly on the video (say, the controls...) will actually close it.  This can't be seen in pattern lab only because the background video is the first example-- but you'll see it if you rename `bolt/apps/pattern-lab/src/_patterns/02-components/video/video--background-video-with-inline-script-and-external-controls.twig` to `z-video--background-video-with-inline-script-and-external-controls.twig`).

Instead of using a separate querySelector here to find this video, you'll need to find a way to use the current video object here...

theSadowski: @remydenton @rockymountainhigh1943 — YES. clicking outside the video itself, whether it's the indigo or outside the band altogether, it should collapse.

rockymountainhigh1943:
Hi @remydenton - I think I have both your items above addressed within an adjustment to the selector. Let me know if you find anything else.

PR #719, opened 2018-05-31
Fix/wwwd 2183 chevron logic fix

remydenton:
This PR updates the particularly troublesome logic for when to add the default right chevron icon to a linked headline.  See the inline comment I added to _typography.twig for an explanation of how it now works.

For context, this line has been changed multiple times in Bolt 1.x attempting to fix this very issue:
dabb904
79d012dd
139e41e

### DIFF
--- a/apps/pattern-lab/src/_patterns/02-components/headline/30-headline-link-variations.twig
+++ b/apps/pattern-lab/src/_patterns/02-components/headline/30-headline-link-variations.twig
@@ -4,21 +4,13 @@
 } only %}
 
 {% include "@bolt-components-headline/headline.twig" with {
-  text: "This headline is also a link, still with Default Icon",
-  url: "https://www.google.com",
-  icon: null
-} only %}
-
-{% include "@bolt-components-headline/headline.twig" with {
   text: "This headline is also a link, with Defined Icon",
   url: "https://www.google.com",
-  icon: {
-    name: "search"
-  }
+  icon: "search"
 } only %}
 
 {% include "@bolt-components-headline/headline.twig" with {
   text: "This headline is also a link, without an Icon",
   url: "https://www.google.com",
-  icon: false
+  icon: "none"
 } only %}
--- a/apps/pattern-lab/src/_patterns/02-components/headline/30-headline-link-variations.twig
+++ b/apps/pattern-lab/src/_patterns/02-components/headline/30-headline-link-variations.twig
@@ -4,6 +4,12 @@
 } only %}
 
 {% include "@bolt-components-headline/headline.twig" with {
+  text: "This headline is also a link, still with Default Icon",
+  url: "https://www.google.com",
+  icon: null
+} only %}
+
+{% include "@bolt-components-headline/headline.twig" with {
   text: "This headline is also a link, with Defined Icon",
   url: "https://www.google.com",
   icon: {
--- a/packages/components/bolt-headline/src/_typography.twig
+++ b/packages/components/bolt-headline/src/_typography.twig
@@ -43,7 +43,12 @@
   iconPosition ? baseClass ~ "--icon-position-" ~ iconPosition : ""
 ] %}
 
-{% if url and icon is not defined %}
+{#
+  If this is a link, add a right chevron icon by default.  There are only two ways to opt out:
+  - Set 'icon: false' to explicitly have no icon
+  - Set a different icon to use instead of the default
+#}
+{% if url and icon is not sameas(false) and icon.name is empty %}
   {% set icon = icon | default({}) | merge({
     name: "chevron-right"
   }) %}
--- a/packages/components/bolt-headline/src/_typography.twig
+++ b/packages/components/bolt-headline/src/_typography.twig
@@ -21,10 +21,36 @@
 {% set style = style | default(schema.properties.style.default) %}
 {% set type = type in types ? type : "text" %}
 {% set size = size | default(schema.properties.size.default) %}
-{% if icon %}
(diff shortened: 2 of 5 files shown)


packages/components/bolt-headline/src/_typography.twig
```diff
@@ -43,7 +43,12 @@
   iconPosition ? baseClass ~ "--icon-position-" ~ iconPosition : ""
 ] %}
 
-{% if url and icon is not defined %}
+{#
+  If this is a link, add a right chevron icon by default.  There are only two ways to opt out:
+  - Set 'icon: false' to explicitly have no icon
+  - Set a different icon to use instead of the default
+#}
+{% if url and icon is not sameas(false) and icon.name is empty %}
```
mikemai2awesome: Is this the new way of doing this kind of logic? We should document this and go back to other components and replace the old method with this new method.

apps/pattern-lab/src/_patterns/02-components/headline/30-headline-link-variations.twig
```diff
@@ -3,6 +3,12 @@
   url: "https://www.google.com"
 } only %}
 
+{% include "@bolt-components-headline/headline.twig" with {
+  text: "This headline is also a link, still with Default Icon",
+  url: "https://www.google.com",
+  icon: null
```
mikemai2awesome: Why are we doing null instead of false here?

remydenton: This is showing that if you pass `null`, it's _not_ the same as `false`.  Basically, if you don't want the chevron, you must pass either `false` or some other icon.  Anything else you do (including null) will result in the default chevron.

packages/components/bolt-headline/src/_typography.twig
```diff
@@ -43,7 +43,12 @@
   iconPosition ? baseClass ~ "--icon-position-" ~ iconPosition : ""
 ] %}
 
-{% if url and icon is not defined %}
+{#
+  If this is a link, add a right chevron icon by default.  There are only two ways to opt out:
+  - Set 'icon: false' to explicitly have no icon
```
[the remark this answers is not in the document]
remydenton: The main reason is that in the past, if you passed "null" (which is very close to but not quite the same as passing nothing), you'd get the default chevron icon.  The repeated changes that we've made to this logic (I linked to them in the PR description) caused regressions because they changed it so that passing `null` actually meant you _wouldn't_ get the default chevron, so a bunch of links in Drupal stopped getting the chevron all of a sudden.

The new setup is as simple as I could make it: either pass `false`, or an icon name, or you get the default.  Do you think it should work differently or just needs better documentation?

apps/pattern-lab/src/_patterns/02-components/headline/30-headline-link-variations.twig
mikemai2awesome: That is just super confusing, when I see icon: null, I am assuming it means do not render the icon.

packages/components/bolt-headline/src/_typography.twig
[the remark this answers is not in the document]
mikemai2awesome: The language is confusing. How do we communicate null means default, and is this consistent across all components? Who's been setting null, is that their true intention? Are we sure they are not trying to remove the icon?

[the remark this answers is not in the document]
mikemai2awesome: Is this some kind of drupal default? Null wasn't set intentionally by anyone?

packages/components/bolt-headline/src/_typography.twig
remydenton: Good point.

The only other place I see us setting a default icon is in [card footer](https://github.com/bolt-design-system/bolt/blob/master/packages/components/bolt-card/src/card.twig#L149).  That does not have this logic because, apparently, there hasn't been a use case to ever need to not have an icon.  Then again, it looks like the icon is missing in some cards.... https://demo1.d8.pega.com/events/pegaworld/agenda

remydenton:
Just to document an extensive Slack conversation that @mikemai2awesome and I had, we're hoping to move to a standard of passing the string `none` whenever you need to explicitly reject a default (Mike documented that principle [in the wiki](https://github.com/bolt-design-system/bolt/wiki/Code-Standards-&-Conventions#how-to-handle-value-of-none)).  So I have work to do on that.

@sghoweri, I'm so glad you had a chance to look as well-- let's follow up with a quick conversation after stand up today.

christophersmith262:
@remydenton is the solution we reached with this backwards-compatible with the way we handle headlines now.  Ie. can we expect to have to refactor to achieve the same output or does this update only allow you to opt-out of the current behavior?

remydenton:
@christophersmith262 yes, the solution is backwards compatible, specifically:

1. Passing nothing (undefined) for `icon` will still give you a right chevron if it's a link or no icon if it's not, as it always has
2. Passing `null` for `icon` will give you the same result as 1, which was previously how it worked until the Bolt 1.x regression that I initially set out to fix.
3. Passing `FALSE` for `icon` will still remove the chevron if it's a link, though this is deprecated in favor of passing the string `none`if you want to remove a default icon.

@sghoweri @mikemai2awesome, can you both take another look at my last three commits?  It's a different approach from the first pass.  New things:

- You can now pass either a string (with an icon name) _or_ an icon object for `icon`
- Passing the string `none` is the new official way to remove the default icon, though passing `FALSE` still works for backwards compatibility.
- Passing `null`, or anything else that's falsey besides `FALSE`, will not affect the defaults.

PR #730, opened 2018-06-06
Button Regression Fix

rockymountainhigh1943:
This pull request adds a fix for the broken button demo page. The logic to make this happen is a little cumbersome and might need to be refactored in another ticket.

The issue arose because `sanitizeBoltClasses()` function strips out all `c-bolt-` classes. This makes it so classes being passed, like `c-bolt-button--hover`, never make it onto the button.

**Steps to fix this issue were:**
1. Update PL demo page to use simplified classes that wont be stripped out (e.g. `is-hover`) and removed the "is" attributes
1. Update our button mixin to catch the new class names

### DIFF
--- a/packages/components/bolt-button/src/button.standalone.js
+++ b/packages/components/bolt-button/src/button.standalone.js
@@ -112,9 +112,9 @@ class BoltButton extends BoltComponent() {
       this.props.disabled ? 'c-bolt-button--disabled' : '',
 
       // Test out psuedo states via prop values
-      this.props.isHover ? 'c-bolt-button--hover' : '',
-      this.props.isActive ? 'c-bolt-button--active' : '',
-      this.props.isFocus ? 'c-bolt-button--focus' : '',
+      // this.props.isHover ? 'is-hover' : '',
+      // this.props.isActive ? 'is-active' : '',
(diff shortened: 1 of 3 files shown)


packages/components/bolt-button/src/button.standalone.js
```diff
@@ -112,9 +112,9 @@ class BoltButton extends BoltComponent() {
       this.props.disabled ? 'c-bolt-button--disabled' : '',
 
       // Test out psuedo states via prop values
-      this.props.isHover ? 'c-bolt-button--hover' : '',
-      this.props.isActive ? 'c-bolt-button--active' : '',
-      this.props.isFocus ? 'c-bolt-button--focus' : '',
+      // this.props.isHover ? 'is-hover' : '',
+      // this.props.isActive ? 'is-active' : '',
+      // this.props.isFocus ? 'is-focus' : '',
```
remydenton: These lines should be deleted rather than commented out-- they're no longer used or intended to be used, so just noise.

We should also remove the lines that define the (now unused) properties https://github.com/bolt-design-system/bolt/pull/730/commits/e957fa3a781d5d9991eb2935138083600566f0e7#diff-ad8093a42b9e5fc5b748d20b6e11b078R73

PR #734, opened 2018-06-07
Fix: update PL specific includes

mikemai2awesome:
Certain include paths were meant to be changed to use the `@pl` prefix, this catches the ones that were missed and updates them.


remydenton:
Actually, one more thing I noticed is the references to the demo form elements for facets.  See . http://localhost:3000/pattern-lab/?p=pages-event-landing

PR #747, opened 2018-06-12
WYSIWYG editor elements demo page and research

mikemai2awesome:
This just adds a page showing what we need to do to make CKEditor work with our components.

In the page dropdown in pattern lab, go to Bolt Dev Sandbox and then wysiwyg kitchen sink to see the page.


remydenton:
@sghoweri, I won't merge for now to give you a chance to look at this.

PR #749, opened 2018-06-12
feat: update CTA text to a text string with default "Share this page"…

theSadowski:
…, update the size of CTA text to small, update content hub anchor ribbon example to xlight and add inline share example at the bottom of page


packages/components/bolt-share/src/share.twig
```diff
@@ -9,10 +9,12 @@
 {% set items = [] %}
 
 {% if isFlat is sameas(true) %}
+  {% set shareText = shareText | default(schema.properties.shareText.default) %}
   {% set items = items|merge([include("@bolt-components-headline/text.twig", {
-    text: "Share this page",
-    transform: 'uppercase',
+    text: 'Share this page',
```
remydenton: I think you want this to be `text: shareText` instead of `text: 'Share this page'`, right?

remydenton: If so, I'd recommend also adding at least one example in Pattern Lab (or updating an existing one) that passes something different for `shareText` so we can see it in action.

remydenton: Also, you have to define the variable `schema` at the top of the file to be able to use `schema.something`. 
 See the button component for an example: https://github.com/bolt-design-system/bolt/blob/bd009998a7e9751b6acafb65dcf60af9a5bcf9e1/packages/components/bolt-button/src/button.twig#L1

packages/components/bolt-share/share.schema.yml
```diff
@@ -30,4 +34,4 @@ properties:
   inline:
     type: boolean
     description: 'Display as a flat layout, rather than popping out'
-    default: False
\ No newline at end of file
+    default: False
```
remydenton: +1 for fixing the formatting :)

PR #769, opened 2018-06-19
Fix: doc styles and formats

mikemai2awesome:
I updated all the docs/readmes to have consistent headlines and info architecture. PL specific CSS has been updated as well. Usage section is using the code snippet component through twig.


mikemai2awesome:
@remydenton I moved some stuff around and the template handles the schema part as well.

apps/pattern-lab/src/_includes/utils/docs.twig
```diff
@@ -1,8 +1,15 @@
+{% set schema = bolt.data.components["@bolt-components-#{componentName}"].schema %}
+{% if componentName %}
+  {% set readmeFile = "@bolt-components-#{componentName}/README.md" %}
+{% endif %}
```
remydenton: I think the `set schema` statement should also be moved inside the `{% if componentName %}` statement: 

```
{% if componentName %}
  {% set schema = bolt.data.components["@bolt-components-#{componentName}"].schema %}
  {% set readmeFile = "@bolt-components-#{componentName}/README.md" %}
{% endif %}
```

remydenton:
Other than my last comment (which is minor), this looks really great.  Thanks @mikemai2awesome!

PR #775, opened 2018-06-22
Make smooth scroll offset dynamic.

charginghawk:
Adds support for modifying scroll offset. It looks for the link's first ancestor that has an 'offset' attribute, returning that value. You'll find offset on the nav-indicator, where gumshoejs also gets its offset.

### DIFF
--- a/packages/components/bolt-smooth-scroll/src/smooth-scroll.js
+++ b/packages/components/bolt-smooth-scroll/src/smooth-scroll.js
@@ -3,14 +3,27 @@ import { isValidSelector } from '@bolt/core';
 
 export const smoothScroll = new SmoothScroll();
 
-export const defaultScrollOptions = {
+export const scrollOptions = {
   ignore: '[data-scroll-ignore]', // Selector for links to ignore (must be a valid CSS selector)
   header: '.js-bolt-smooth-scroll-offset', // Selector for fixed headers (must be a valid CSS selector)
 
   // Speed & Easing
   speed: 750, // Integer. How fast to complete the scroll in milliseconds
-  offset: 45, // Integer or Function returning an integer. How far to offset the scrolling anchor location in pixels
   easing: 'easeInOutCubic', // Easing pattern to use
+
+  // Here we have support for modifying scroll offset. It looks for the link's
+  // first ancestor that has an 'offset' attribute, returning that value. You'll
+  // find offset on the nav-indicator, where gumshoejs also gets its offset.
+  // @see nav-indicator.js
+  offset(anchor, toggle) {
+    var offsetElement = toggle.closest('[offset]');
+    if (offsetElement) {
+      return offsetElement.getAttribute('offset');
+    }
+    else {
+      return 0;
+    }
+  },
 
   // Callback API
   updateURL: true, // Update the URL on scroll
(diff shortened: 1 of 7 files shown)


packages/components/bolt-smooth-scroll/src/smooth-scroll.js
```diff
@@ -3,15 +3,27 @@ import { isValidSelector } from '@bolt/core';
 
 export const smoothScroll = new SmoothScroll();
 
-export const defaultScrollOptions = {
+export const scrollOptions = {
```
remydenton: `defaultScrollOptions` is exported for use elsewhere; changing the name means we need to change places it's being consumed as well.

The only place I know of currently is in navlink.js:
https://github.com/bolt-design-system/bolt/blob/master/packages/components/bolt-navlink/navlink.js#L17
https://github.com/bolt-design-system/bolt/blob/master/packages/components/bolt-navlink/navlink.js#L71

packages/components/bolt-smooth-scroll/src/smooth-scroll.js
```diff
@@ -3,15 +3,27 @@ import { isValidSelector } from '@bolt/core';
...
+  // @see nav-indicator.js
+  offset: function (anchor, toggle) {
+      var offsetElement = toggle.closest('[offset]');
+      if (offsetElement) {
+          return offsetElement.getAttribute('offset');
+      } else {
+          return 0;
+      }
+  },
+
```
remydenton: A pretty reasonable approach.  I really do feel like we should make this easier on consumers though-- as of now, they have to know and remember to put an "offset" attribute on every container that contains smooth scroll links.  Great if you want that flexibility, not so much if you don't.

In short, can we make a page-level default?  I guess you could actually do that with this code by setting the attribute on body:
`<body offset="45"> ... </body`.
That just feels a little dirty.

packages/components/bolt-smooth-scroll/src/smooth-scroll.js
charginghawk: Good catch. Updated accordingly.

packages/components/bolt-smooth-scroll/src/smooth-scroll.js
charginghawk: Actually, I think setting offset on the body tag would be the good track. Maybe you could clean it up - offset is probably too generically named, and you could have js go directly to the body instead of traversing ancestry (although you'd lose the flexibility). The BoltSettings dream still lives, though - we could update stuff in pure js instead of having attribute proxies.

In any case, this accomplishes the immediate goal.

remydenton: Yep, agreed.  I think the end game here is a global `BoltSettings` in pure js.  But this nicely accomplishes the goal in the meantime.

Marking this PR as approved.  I'll hold off merging for a bit to give Salem a chance to look, but will go ahead and merge in early next week if he hasn't responded by then.

charginghawk:
>1. Are we accounting for the need to have this offset value change dynamically (ex. if the global header height changes at different screen sizes)?

Yep, smooth scroll checks every time before scrolling, and I believe gumshoe updates automatically too (I believe @remydenton wrote the js so he can confirm). On the Drupal side we update it whenever the window gets resized:

http://vstash:7990/projects/PEGAWWW/repos/pegawwwd8/pull-requests/1337/overview

So we should be covered.

As for 2 and 3, that's not for me to say per sé, but my stated preference is for a boltSettings javascript object that maps with the drupalSettings javascript object:

http://cgit.drupalcode.org/drupal/tree/core/misc/drupalSettingsLoader.es6.js

remydenton:
1. Yes-- scroll links will just read the scroll value at runtime when they're clicked, so any changes made to the scroll attribute value before that moment will affect what happens.
2. At the very least, we should actually update the schema for navbar-priority at bolt/packages/components/bolt-nav-priority/nav-priority.schema.yml.  @charginghawk, can you take a stab at that actually?
3. As I just saw @charginghawk mentioned, and as [we discussed earlier on this PR](https://github.com/bolt-design-system/bolt/pull/775#discussion_r197568762), I think the best solution is a global javascript setting (and a system for handling that) rather than a separate component.

@charginghawk, okay if we consider this a temporary workaround and something we plan to remove soon in favor of the Javascript settings?  I want to unblock you, especially if that doesn't happen as soon as we'd like, but handling this globally really does seem better and more maintainable.

remydenton:
I made some updates to documentation (and a few other random things I saw while working, like translatability).

I'm going to merge this to unblock Thomas, but I'm also going to create a proof-of-concept/RFC branch with Bolt settings for discussion as I think that's the better solution long-term.

PR #778, opened 2018-06-26
fix: Add extra offset to gumshoe to fix perceived bug (WWWD-2278)

remydenton:
Resolves http://vjira2:8080/browse/WWWD-2278


remydenton:
Agree @sghoweri, it is absolutely an arbitrary number.  That is kind of the point though-- it has nothing to do with some accurately determined thing and more just what "feels" right, and for now doesn't bear any relation to other things on the page or design system.

The option to custom configure it could come in #779.  But, I think it's over engineering to set up configuration for something that doesn't have an obvious use case to configure?

In any case, the issues discussed on the Jira issue bother me (i.e. this somehow affects smooth scroll offset too...).  Maybe this is something we consider saying "won't fix" on, at least for now.

PR #792, opened 2018-07-03
Fix: remove align-items prop from text align util

mikemai2awesome:
This fixes the text-align utility class.

# Changes
1. Removed `align-items` prop because that depends on `flex-direction` being declared for desired results, which makes this util has no context of.
2. Added `start` and `end` values for better internationalization support. This allows RTL pages to know the exact intended alignment. (source: https://davidwalsh.name/text-align-start)
3. Added `justify` option.


mikemai2awesome:
@remydenton @sghoweri Added comments to the release PR.

PR #815, opened 2018-07-16
BDS-452 Pagination additional features and improvements

remydenton:
Resolves http://vjira2:8080/browse/BDS-452

Additionally:
- Adds full support for link attributes
- Makes the active item not a link
- Adds align property to center by default

### DIFF
--- a/apps/pattern-lab/src/_patterns/02-components/pagination/15-pagination-count-variations.twig
+++ b/apps/pattern-lab/src/_patterns/02-components/pagination/15-pagination-count-variations.twig
@@ -54,6 +54,38 @@
 
 <hr>
 <div class="u-bolt-margin-bottom-medium">
+  <h3>1 of 7</h3>
+  {% include "@bolt-components-pagination/pagination.twig" with {
+    current: 1,
+    total: 7,
+    pages: {
+      1: {
+        href: '#!'
+      },
+      2: {
+        href: '#!'
+      },
+      3: {
+        href: '#!'
+      },
+      4: {
+        href: '#!'
+      },
+      5: {
+        href: '#!'
+      }
+    },
+    next: {
+      href: '#!'
+    },
+    last: {
+      href: '#!'
+    }
+  } only %}
+</div>
+
+<hr>
+<div class="u-bolt-margin-bottom-medium">
   <h3>4 of 5</h3>
   {% include "@bolt-components-pagination/pagination.twig" with {
     current: 4,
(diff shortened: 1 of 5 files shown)


sghoweri:
Nice @remydenton!

3 quick questions to run by you on this (all pagination related - albeit 2 are just to get me up to speed on pagination in general):

1. Where exactly are these new First and Last links on the pagination section? Not seeing them rendered anywhere on my device... 
https://feature-bds-452-pagination-first-last.bolt-design-system.com/pattern-lab/?p=viewall-components-pagination

1c6e2f1e-335e-4ed1-9bc9-a4936e9fd4c9

4642ef9c-cf8a-482a-a90c-576606f76a36

## More general questions  
2. Why isn’t this component centered by default? I get allowing for different alignments but isn’t it reasonable to assume most of the time this’ll be centered unless otherwise specified?

3. Why does the center “current” link background change color when I tap on it on my phone? As a rule of thumb, non-interactive elements should have pseudo state changes (especially once activated)... bug?

remydenton:
@sghoweri,

1. The first and last links don't actually say "first" and "last"-- they're just the numbers "1" and "10" (or whatever the last page is).  See the mockup image on http://vjira2:8080/browse/BDS-452

Re: 2 and 3, @mikemai2awesome is probably the best guy to talk to.  But here's my take:

2. Good question.  As a rule of thumb, a component should not be responsible for its own alignment.  But I also agree this will almost always be centered.  I'm torn on this one.

3. Not sure I understand.  Did you mean to say "interactive elements" instead of "non-interactive elements"?  Or that "non-interactive elements should NOT have pseudo state changes"?  As built, the current page link is actually a link and therefore interactive, which to me means in _should_ have an interactive state, which it does...

mikemai2awesome:
@remydenton @sghoweri 

2. I can be swayed on this one, but adding center by default will create some undesired results when it is not by itself in a row. In the example of pagination on the left and something on the right, you would have to intentionally make it not center for it to work. So if we insist on it being centered by default, we'll have to make an additional prop to handle the alignment.

3. This is also something I don't like, but I refrained from changing it last time I touched this. Basically the current item shouldn't be a link, it doesn't make sense functionally and for accessibility, a page should not link to itself because that will take the user for a loop. I didn't know how much this change would affect the Drupal end so I didn't update it.

remydenton:
Agree with you both on item 3 -- I just pushed a commit that makes the active item not a link.

Just need to figure out 2 now.  Am I old school or could we just use `text-align: center` and `display: inline-block` instead of flex for the layout, then just use `u-bolt-text-align-left` in the rare occasions where you don't want it centered?

mikemai2awesome:
@remydenton @sghoweri 

Align prop has been added.

sghoweri:
Thanks @remydenton - thanks @mikemai2awesome! 

Sounds like this should be all ready to go and can get merged in, yes?

remydenton:
@mikemai2awesome's improvements look great, I approve those.  Since you guys both gave me verbal approval on the other pieces, I'm merging this in and adding some release notes.

PR #822, opened 2018-08-02
feat: Remove all vertical padding on video share overlay (BDS-495)

remydenton:
## Jira
http://vjira2:8080/browse/BDS-495
http://vjira2:8080/browse/WWWD-2367

## Summary

Prevent scrollbar from appearing on video overlay unless absolutely necessary.

## Details

Stakeholders were pretty adamant about not seeing scrollbars in the video overlay.  This takes previous measures a step further still without hindering usability.

Specifically, this removes the top and bottom padding from the overlay.  It's vertically centered anyway, so this only affects cases where the content is slightly too large to fit vertically.  In those cases, the share content now presses closer to the top and bottom of the video player without activating the scrollbar.  However, the scrollbar still activates if the content would actually be cropped.

## How to test

- Go to the videos page `pattern-lab/?p=viewall-components-video`
- Make the browser width as small as possible in pattern lab
- Play a video, then press the share button icon to pull up the share overlay.
- Attempt to scroll up or down in the video
Previous behavior: the overlay would scroll up and down, even though no content was hidden
New behavior: the overlay does not scroll unless content is actually cropped.

### DIFF
--- a/packages/components/bolt-video/src/_videojs-enhancements.scss
+++ b/packages/components/bolt-video/src/_videojs-enhancements.scss
@@ -143,7 +143,7 @@ $bolt-video-js-button-shadow-level: 'level-40';
     background: rgba(bolt-color(black), 0.9);
 
     .vjs-modal-dialog-content {
-      @include bolt-padding(medium large medium medium);
+      @include bolt-padding(0 large 0 medium);
       @include bolt-font-size(medium);
 
       display: block;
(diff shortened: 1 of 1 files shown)


mikemai2awesome:
The overall font-size and spacing are tighten up. Content only overflows if the video goes below 220px wide, at that point, the brightcove player ui is breaking, too. So a video shouldn't get any smaller than that anyway, otherwise the user wouldn't even be able to interact with it.

remydenton:
Thanks again @mikemai2awesome for the more thorough solution.  Looks great.

PR #836, opened 2018-08-14
Subfeature/bds 336 fix attributes support

remydenton:
## Jira
N/A (see #835)

## Summary

This is a "sub-feature" PR with cleanup/fixes for #835

## Details

In addition to various minor cleanup, this fixes a regression where the ability to pass attributes besides classes to the video component was lost.  To see the regression:

- Go to the `video w inline script and email share` demo on the master branch for #835 (https://bolt-design-system.com/pattern-lab/?p=components-video-w-inline-script-and-email-share).
- Play the video, then click to share.  You should see an email share button based on data attributes passed in.
- Now go to the `video w inline script and email share` demo on the feature branch for #835 (https://fix-bds-336-enable-utilities-on-video.bolt-design-system.com/pattern-lab/?p=components-video-w-inline-script-and-email-share).
- Play the video, then click to share.  You should not see the email share link (this is the regression)

## How to test
- Now go to the `video w inline script and email share` demo from this PR
- Play the video, then click to share.  Confirm that the share details have been populated the same as on the master branch.

### DIFF
--- a/packages/components/bolt-video/src/video.twig
+++ b/packages/components/bolt-video/src/video.twig
@@ -18,32 +18,24 @@
 ] %}
 
 {% if ratio == true %}
-  {% set utilClasses = [] %}
-  {% set videoClasses = [] %}
-
+  {# Move utility classes to a separate attributes object so they can be added to the wrapping ratio object instead #}
+  {% set ratioAttributes = create_attribute({}) %}
   {% for class in attributes["class"] %}
     {% if class starts with "u-" %}
-      {% set utilClasses = utilClasses|merge([class]) %}
-    {% else %}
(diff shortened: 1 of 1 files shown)


remydenton:
That is all it's doing, but It's still an attributes object.  I think the naming works best as-is.  @sghoweri, any thoughts on this?

remydenton:
For what it's worth, I do think any time we're managing classes (not to mention other attributes), we should use attributes objects rather than arrays for consistency.  I suppose that's a part of the cleanup in this PR I didn't explicitly call out.

mikemai2awesome:
@remydenton The reason for that is Ratio still has its own set of attributes, right? Just don't want the name to mislead.

PR #852, opened 2018-08-29
Fix: convert bolt-list to use the inner attributes convention

mikemai2awesome:
## Jira

http://vjira2:8080/browse/BDS-548

## Summary

Replace `parentAttributes` method in bolt-list with `inner_attributes`.

## Details

As discussed with @sghoweri and @remydenton, the `inner_attributes` method is the new standard of handling outer and inner attributes. By default, all attributes go to the parent container, `inner_attributes` only render the component specific class names.

In addition, I fixed some variable names to follow our conventions.

## How to test

Try to pass util classes (`u-bolt-whatever`) or other attributes to the bolt-list component and make sure they only get added to the `<bolt-list>` tag and not anything inside.

### DIFF
--- a/packages/components/bolt-list/src/list.twig
+++ b/packages/components/bolt-list/src/list.twig
@@ -7,6 +7,7 @@
 {# Variables #}
 {% set baseClass = "c-bolt-list" %}
 {% set attributes = create_attribute(attributes|default({})) %}
+{% set innerAttributes = create_attribute({}) %}
 
 {# Set up checks to validate that the component's prop values are allowed, based on the component's schema. #}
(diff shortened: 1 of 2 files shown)


sghoweri:
@mikemai2awesome I just noticed that the List component’s view-all page appears to be scrolling off to the side on smaller screen sizes (overflow issue?): separate unrelated bug, yes? 

https://fix-bolt-list-attributes.bolt-design-system.com/pattern-lab/patterns/02-components-list/index.html

af29c544-033a-4683-a507-cd29bb1b5617

mikemai2awesome:
@sghoweri Separate issue. I need to fix the spacing poking out.

PR #859, opened 2018-09-05
BDS-575 Deprecate itemAlignment parameter in button component

remydenton:
## Jira

http://vjira2:8080/browse/BDS-575

## Summary

- Deprecates `itemAlignment` property of button component.  Use `align` instead.
- Deprecates `left` and `right` values for `align` property.  Use `start` and `end` instead.

## Details

More important than this specific deprecation is that we're defining the workflow for deprecation.  The steps in this PR create the deprecation; after it's merged, I'll create an additional PR against 2.x to remove support for the deprecated parameter.

I've created this page in the wiki to describe the whole process: https://github.com/bolt-design-system/bolt/wiki/Deprecating-component-properties

## How to test

1. Confirm schema shows as invalid if there's any usage of `itemAlignment`
    - Check out this branch
    - Add `itemAlignment: "start"` to button (anywhere it is removed in this PR)
    - Run `cd apps/pattern-lab` and `npm run start`.  The build should fail due to schema errors

2. Confirm the deprecated parameter can still be used
    - Stop pattern lab
    - In `apps/pattern-lab/.boltrc.js`, set `schemaErrorReporting` to `console`
    - Run `cd apps/pattern-lab` and `npm run start`.  The build should work, deprecated parameters should function as they have in the past, and schema validation errors should be shown in the browser console when viewing pattern lab pages that use the deprecated parameter.

### DIFF
--- a/apps/pattern-lab/src/_patterns/02-components/card/00-card-docs.twig
+++ b/apps/pattern-lab/src/_patterns/02-components/card/00-card-docs.twig
@@ -26,7 +26,7 @@
       pattern: "button",
       style: "text",
       width: "full",
-      itemAlignment: "start",
+      align: "left",
       text: "Get the report"
     }
   ]
--- a/apps/pattern-lab/src/_patterns/02-components/card/00-card-docs.twig
+++ b/apps/pattern-lab/src/_patterns/02-components/card/00-card-docs.twig
@@ -24,9 +24,7 @@
     },
     {
       pattern: "button",
-      style: "text",
       width: "full",
-      align: "left",
       text: "Get the report"
     }
   ]
--- a/apps/pattern-lab/src/_patterns/02-components/card/05-card.twig
+++ b/apps/pattern-lab/src/_patterns/02-components/card/05-card.twig
@@ -26,7 +26,7 @@
       pattern: "button",
       style: "text",
       width: "full",
-      itemAlignment: "start",
+      align: "left",
       text: "Get the report"
     }
   ]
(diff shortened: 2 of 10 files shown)


remydenton:
I agree that `start` and `end` are preferable to `right` and `left`.

This deprecation was first introduced in 74da8ee.  Before that, I believe `itemAlignment` did accept `start` and `end` (hard to say for sure just by looking at that commit though).  @sghoweri, was that an intentional choice?

remydenton:
Cool, thanks guys.  I'll go ahead and deprecate the newer properties then as well 😅

remydenton:
@mikemai2awesome, I've updated the buttons to use "start" and "end" for alignment.  Two important notes:

1. I did not actually change `text-align: left;` in the CSS, because there are some rules that apply left and right margins as well which would not work with `start` and `end`.  Should we be updating the `bolt-margin-right()` mixin to be `bolt-margin-start()` as well?  I think that's beyond the scope of this PR, but an important decision to make for the future.

2. I discovered that passing in either `align` or `style` does _nothing_ to buttons included in `contentItems` within cards.  That's because the [code in card.twig](https://github.com/bolt-design-system/bolt/blob/feature/BDS-575-deprecate-itemAlignment/packages/components/bolt-card/src/card.twig#L166) will always force set them to hard coded values (namely, "start" and "text").  This may be intentional (@sghoweri?), and is probably best left as-is anyway since changing it could break things.  In any case, the bigger problem is that our docs show usage examples that are actually wrong, so I fixed that at least.

PR #864, opened 2018-09-12
fix: prevent device viewer from overlapping bg video WWWD-2574

remydenton:
## Jira

http://vjira2:8080/browse/WWWD-2574

## Summary

Fixes z-index regression where device viewer overlaps band background video

## Details

Honestly, it's unclear to me why this works.

### DIFF
--- a/packages/components/bolt-device-viewer/src/device-viewer.scss
+++ b/packages/components/bolt-device-viewer/src/device-viewer.scss
@@ -13,6 +13,7 @@ bolt-device-viewer {
   height: auto; // Based on inner contents + padding;
   margin: 0 auto;
   box-sizing: content-box; //Required for now till device viewer updated to be fluid in size.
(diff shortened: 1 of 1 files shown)


charginghawk:
@remydenton It's because transform styling creates a new "stacking context" that puts children elements into kind of a z-index pocket universe. For the descendants of an element with transform styling, their z-index only applies among the other children - otherwise they have the z-index of the element of the transform styling. 

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

It can cause some headaches (which we've run into a lot with the main nav), which is why I'm generally suspicious of using transform styling when there are alternatives (eg top vs transform is an ongoing debate).

remydenton:
Thanks @charginghawk!  (@sghoweri actually pointed me to similar docs).

I'm going to go ahead and merge this since it's essentially Salem's code, and it now makes sense to me.

PR #879, opened 2018-09-20
BDS-569 refactor copy to clipboard

remydenton:
## Jira

http://vjira2:8080/browse/BDS-569

## Summary

Significantly refactors copy to clipboard component to make it more flexible

## Details

Beyond some general cleanup, the most important change here is that **you can now pass anything (string, rendered pattern, etc) for trigger, spinner, and confirmation**.  The only requirement is that the trigger have a `data-clipboard-text` attribute somewhere that includes the text to copy.

This is based a discussion Mike and I had a week or so ago.  I want a gut check from both @mikemai2awesome and @sghoweri though that this makes sense before I go further.  Specifically:

- Does this solve Mike's use case in share?
- Is this a sensible direction for copy-to-clipboard?

TODO (just FYI):
- add theming support
- ~deprecate/rename variables (most of the existing variables no longer make sense -- "copiedText", for example, will probably rarely be used and users could just change it and more by passing a custom value for "success") . "url" should actually be "clipboard_text" or something similar, because you might want to copy something other than a url to the clipboard).~
- ~continued cleanup (replace success with a non-link, for example) and ensure compatibility with share component.~

### Deprecations:
- The  `iconSize` property is deprecated and will be removed in the next major version of Bolt.  Use the new custom_trigger, custom_transition, and custom_confirmation properties if you need to change icon size.
- The  `copiedText` property is deprecated and will be removed in the next major version of Bolt.  Use the new custom_confirmation property if you need to change the confirmation text.
- The `text` property has been renamed `trigger_text`.  The `text` property is deprecated and will be removed in the next major version of Bolt.
- The `url` property has been renamed `text_to_copy`.  The `url` property is deprecated and will be removed in the next major version of Bolt.

## Testing Instructions
- View the newly added pattern lab demos at https://feature-bds-569-refactor-copy.bolt-design-system.com/pattern-lab/?p=components--custom-copy-to-clipboard


mikemai2awesome:
@remydenton The ability to pass anything for the trigger, spinner and confirmation is definitely awesome!

remydenton:
Thanks @mikemai2awesome.  Regarding updating pre-existing variables, that's where I was going with the second bullet point under "TODO".  To summarize my suggestions:

- Remove `iconSize`.  It should be rarely changed, and could be customized if needed using the new trigger, spinner, and confirmation variables.
- Remove`copiedText`.  Same reason as icon size.
- Keep `text` because it may be used more often (maybe refactor as `trigger_text` for clarity though?)
- Keep `url` but refactor to... `text_to_copy`?  I think it won't work with anything but a text string.

Any objections?

mikemai2awesome:
`trigger_text` and `text_to_copy` make sense.

mikemai2awesome:
@remydenton 

`spinner` needs to be renamed to `transition`. The fading in and out is specific to the component, while how the individual elements animate should belong with each element, this allows scalability for doing separate custom animation for each element.

When we have a animation library, a custom `transition` can use whatever animation is available, it would not be tied to spinning.

PR #895, opened 2018-10-02
Twig Renderer creation & initial Template rendering Jest test

EvanLovely:
## Jira

BDS-585

## Summary

Adds `@bolt/twig-renderer` which exports a `render()` and `renderString()` function that can be passed a template and data then will return an HTML string. Introduces ability to create individual template rendering snapshot testing.

## Details

See `bolt-button/__tests__/button.js` for how tests are created. In a nutshell, it's this:

```js
const { render } = require('@bolt/twig-renderer');
const { readYamlFileSync } = require('@bolt/build-tools/utils/yaml');
const { join } = require('path');
const schema = readYamlFileSync(join(__dirname, '../button.schema.yml'));
const { tag } = schema.properties;

describe('button', async () => {
  test('basic button', async () => {
    const results = await render('@bolt-components-button/button.twig', {
      text: 'Hello World',
    });
    // console.log({ results });
    expect(results.ok).toBe(true);
  });

  tag.enum.forEach(async tagChoice => {
    test(`button tag: ${tagChoice}`, async () => {
      const results = await render('@bolt-components-button/button.twig', {
        text: 'Hello World',
        tag: tagChoice,
      });
      // console.log({ results });
      expect(results.ok).toBe(true);
    });
  });
});
```

## How to test

- A `bolt build` in `apps/bolt-site` must be ran first as this creates all the files in the data directory that the Bolt Core Twig Extension expects to find to fill `bolt.data`. You can see that the `.boltrc.js` at the top of the repo points to that same directory for reading files.
- Run `yarn test:js` (which just runs `jest`) to fire off tests. A couple commands that may be of interest:
  - `jest path/to/dir/` - will just run tests found in that dir; so we could make a `test:components` script with `jest packages/components` if we'd like. Or one could run `yarn test:js packages/components/bolt-button` to just run that single test.
  - `jest -u` will update any out of date snapshots with latest examples.


packages/components/bolt-button/__tests__/button.js
```diff
@@ -0,0 +1,28 @@
...
+  test('basic button', async () => {
+    const results = await render('@bolt-components-button/button.twig', {
+      text: 'Hello World',
+    });
+    // console.log({ results });
+    expect(results.ok).toBe(true);
+    expect(results.html).toMatchSnapshot();
+  });
+
+  tag.enum.forEach(async tagChoice => {
```
sghoweri: +1 👍🏼

sghoweri:
> 1145d5a6-26b9-420f-9a80-2a9ff0a5a583
> Lies! 🍰

@EvanLovely head’s up: our Nightwatch.js safety net that tests to verify Pattern Pab and the Docs sites compile and deploy successfully randomly stopped working very recently... likely due to an API change in Now.sh.

a9589483-ea3d-4d28-a8fc-d18cbbcb2b55

CC @margoromo and @remydenton since I mentioned this Nightwatch check breaking...

sghoweri:
> A bolt build in apps/bolt-site must be ran first as this creates all the files in the data directory that the Bolt Core Twig Extension expects to find to fill bolt.data. You can see that the .boltrc.js at the top of the repo points to that same directory for reading files.

Out of curiosity, what specifically do our Jest tests _actually_ need in order to run? The Twig namespace JSON data (plus maybe some might need the global JSON data coming from our export data Sass function)?

Similar to how I’ve been refactoring much of the build to still work when tasks or functions are run in a different sequence than we originally planned for  (async / await), I’ve been working on reducing the amount of stuff we need to do before we can compile our Twig templates.

For example, rather than erroring out if a file requested by our Twig `getData` function, I recently added a less heavy handed [“fileExists” check](https://github.com/bolt-design-system/bolt/blob/master/apps/bolt-site/templates/_site-head.twig#L3) to grab a file if it’s there — and not worry about it if it’s not —> allows for more parallel Webpack and Twig compiles

sghoweri:
^ a relatively easy solution could be to just have Webpack compile a watered down version of the .boltrc config — say with just the Bolt core package included to generate Sass exported data

packages/build-tools/utils/config.schema.yml
```diff
@@ -188,6 +191,9 @@
         - none
         - console
         - cli
+    configFileUsed:
```
sghoweri: Hate the name, love the idea 😉. Maybe something like “configPath” or “pathToConfig”? #namingThingsIsHard

packages/build-tools/utils/config.schema.yml
```diff
@@ -76,6 +76,9 @@
       type: number
       title: Desired port for PHP Rendering Service
       default: 8087
+    alterTwigEnv:
+      type: array
+      description: Passed directly to the @basalt/twig-renderer config option of same name
```
sghoweri: Sure, but if you or me weren’t the ones looking at this, would we know what this actually does?

“Extends and alters the PHP-based Twig environment used when compiling Twig templates. Used for adding custom Twig extensions + [configuring environment options](https://twig.symfony.com/doc/2.x/api.html#environment-options) like `autoescaping` or `debug` mode.”

.boltrc.js
```diff
@@ -13,7 +13,28 @@
  *   npx webpack-cli --config="create-webpack-config.js" --progress
  * ```
  */
+const path = require('path');
+const globby = require('globby');
```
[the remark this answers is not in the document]
EvanLovely: No, there should be no work needed for extra setup. If the `@bolt/twig-renderer` was used, it'll just look for the nearest config file.

packages/build-tools/utils/config.schema.yml
EvanLovely: It's not a path to a config, because then the file would referencing itself - this turns into a path to the `.boltrc` file used. This is programmatically added, but needs to be in schema b/c it's before validation. Renaming in future would be an easy search and replace as it's not a config value anyone would place in a `.boltrc` file.

EvanLovely:
> Out of curiosity, what specifically do our Jest tests actually need in order to run? The Twig namespace JSON data (plus maybe some might need the global JSON data coming from our export data Sass function)?

Our Jest tests need to render a Twig template, and in order to do that the Bolt Core Twig Extension needs to be able to initialize. So here's the minimum:

- A config file needs to be read so it can find out the list of components, and the path to the data directory can be found
- The manifest needs to be created (grabbed via function but is written to `full-manifest.bolt.json`)
- The twig namespaces config needs to be created (grabbed via function but is written to `twig-namespaces.bolt.json`)
- Twig Env gets bootstrapped by declaring namespaces, then registering all Twig Extensions
- When the BoltCore Twig Extension is initiated [here](https://github.com/bolt-design-system/bolt/blob/97ddf1b9095d6f7cc30909c3487621ab41bdb347/packages/core-php/src/TwigExtensions/BoltCore.php#L18-L76), it looks in the data directory and reads any `*.bolt.json` and places them in the global `bolt.data` - this is how the components can get access to their own schema via the full manifest

All the scss created files (like `breakpoints.bolt.json`) could be absent, but if there were any Twig files that tried to access `bolt.data.breakpoints`, then it wouldn't find it. This should be good as those tend to only be used on the demo patterns for breakpoints, colors, etc. However, I'd suggest keeping the Twig Renderer using the environment that is close as possible to what gets rendered in that environment (static, pl, drupal, etc). So basically we could skip the webpack compile phase for a quicker start.

EvanLovely:
@sghoweri ok, I believe I've addressed all your questions; take a peak!

sghoweri:
> @sghoweri ok, I believe I've addressed all your questions; take a peak!

@EvanLovely I think you’re right. Thanks!

Any ideas on why the docs site can’t compile (on Travis and locally on this branch) — the problem I mentioned above? I think that’s the only thing currently holding this back!

EvanLovely:
Strange, if this line is commented out, then it works:

https://github.com/bolt-design-system/bolt/blob/97ddf1b9095d6f7cc30909c3487621ab41bdb347/packages/build-tools/tasks/task-collections.js#L153

PR #899, opened 2018-10-06
ensure build task does not exit during start task

EvanLovely:


EvanLovely:
Also fixes build task params. They were two booleans, but start task called them with a single object with two booleans.

PR #907, opened 2018-10-10
restores ability to point to custom config file with --config-file

EvanLovely:


remydenton:
Looks sensible to me.  I changed the base branch since it's a hotfix, will merge and release as soon as automated checks pass so we can test in out on the Drupal end of things asap.

PR #930, opened 2018-10-25
WWWD-2734 theming card content

remydenton:
## Jira

http://vjira2:8080/browse/WWWD-2734

## Summary

Fixes card theming

## Details

See the original ticket for a description of the problem.  I don't really understand the fix other than that removing the lines in this PR doesn't seem to affect the end result as far as I can tell.

**This definitely needs a review from someone who knows the intention behind those lines (i.e. are they just cruft or actually important?)**

### Confirm the problem
- Go to https://boltdesignsystem.com/pattern-lab/?p=components-card-band-variation.
- In IE 11, use dev tools to add plain text inside the body of the first card inside the dark theme band
- Confirm that the text appears white, which is incorrectly matching the theme of the band rather than the card.

screen shot 2018-10-25 at 2 54 19 pm

### Confirm the fix
- In IE 11, go to /pattern-lab/?p=components-card-band-variation on this branch
- Confirm that the text "Plain text" appears in the correct text color for the theme of the card (dark indigo).


remydenton:
Is that one-off/temporary thing that would get removed once we merge in the card updates?  Or should we be specifying `color: bolt-theme(text)` on effectively every possible element...

mikemai2awesome:
@remydenton 

Not one off, that needs to stay. If you look at the floating label and the inputs, they all have color defined, the inline label was an oversight. We should get in the habit of defining colors inside components.

remydenton:
Okay, updated pushed.  I also updated the PL demo, which now looks like this in IE (plain text in card is still wrong, but form label is working):

screen shot 2018-10-29 at 6 56 07 am

Additionally:
- IMO, It does seem to me like color should cascade-- i.e. if you don't specify color for an element like form label, it should inherit the color of the surrounding text.  I'm trusting @mikemai2awesome on this one, but curious if you agree @sghoweri.
- I know we've gone back and fourth on where to put demos like this that are less about showing usage than testing it.  Should these go in `card/_tests_` instead?

mikemai2awesome:
@remydenton 

Can you do a form card demo instead? I don't think we need to showcase the plain text bug in PL.

I also don't like the organization of tests, they shouldn't be mixed together with demos.

remydenton:
Okay, updates do the following:

1. Clean up form demos (in particular, add number prefixes for better sorting)
2. Add a demo of a themed band with themed cards to the forms section of pattern lab
3. Remove the demo of the plain text bug from the card section of pattern lab

I think this should be ready to merge, but I want to add these addenda:
- I still think form label (or any other component) should be able to inherit color from the surrounding text if it makes sense to do so.  Mike and I disagree on this and I defer to him, but still curious if you have an opinion @sghoweri.
- Not being able to put plain text in a card is a still-unresolved bug.  Ideally, I'd like to see a demo of this alongside the fixes in Mike's card refactor.

mikemai2awesome:
@remydenton Looks good. As I mentioned before, the plain text bug is fixed on my refactor branch, when theming is used correctly, this is not an issue. And yes, I think we need a good way to handle inherit colors.

sghoweri:
> I still think form label (or any other component) should be able to inherit color from the surrounding text if it makes sense to do so. Mike and I disagree on this and I defer to him, but still curious if you have an opinion @sghoweri.

@remydenton @mikemai2awesome inheritance is very tricky in this particular situation. 

For browsers with native CSS Custom Properties support it’s not an issue but for those that don’t,it starts to become a terrible game of whack a mole if some other unwanted component’s style’s trump what would otherwise get inherited... (it’s super brittle).

Happy to dive deeper into this with you guys!

remydenton:
Thanks guys!  I'm merging this in since it gets the job done and improves forms.  Looking forward to seeing the full card updates.

PR #940, opened 2018-10-30
fix: Add simple button for forms

remydenton:
## Jira

http://vjira2:8080/browse/WWWD-2739

See also
http://vjira2:8080/browse/WWWD-2724
Possibly http://vjira2:8080/browse/WWWD-2711

## Summary

Adds a variant of button for use in forms that does not invoke any Bolt javascript

## Details

The Preact-based `<bolt-button>` component has a history of being obstinate when it comes to Drupal forms.  When we were initially building out forms in Bolt, there were several obstacles and revisions, but we were ultimately able to get them resolved such that Drupal forms worked as expected when using `<bolt-button>`.

With Bolt 2.0, however, there were severe regressions to forms (see linked Jira issues).  These were:
- painful for stakeholders (Ross and Liz), who got chewed out because forms weren't working as expected.
- painful for myself, @charginghawk, @krlucas, and others who spent significant development and/or communication time last week trying to resolve these elusive and high priority issues.
- never fully resolved.  For example, http://vjira2:8080/browse/WWWD-2739 was ultimately resolved by disabling AJAX form submission. 

Is Drupal doing something "wrong" here when it comes to attaching javascript behaviors to buttons?  Maybe.  Maybe not.  We've sunk enough time troubleshooting though, both last week and previously, that this is no longer worth fixing.  To that end, this PR adds a simple, javascript-free/CSS only button that can be used by forms.  The primary downside is some (hopefully minimal) increased maintenance overhead on the Bolt side when making updates that affect `<bolt-button>` markup.  This will be accompanied by an update to the `pega_bolt` module to use this new twig template.

@krlucas suggested in our meeting about this last Thursday that we could potentially bring back `<bolt-button>` in forms once we had automated testing coverage.  I'm frankly skeptical, partly because even if we know _that_ form submission is broken, we don't have a clear path to fixing it (which is our current scenario).  I'd love to be proven wrong though.

## How to test

View the Bolt form demo on this feature branch at `pattern-lab/patterns/02-components-form-form-full-campaign-landing/02-components-form-form-full-campaign-landing.html` and confirm parity with the existing form button.


krlucas:
@remydenton Thanks for the great write up. You really synthesized the motivation and business case for the "simple button" (form button) with stunning clarity.

In addition, the code change is elegant. While you're adding a few non-comment lines, they're specifically about reducing abstraction.

As for when we re-introduce `<bolt-button>` to Drupal forms and deprecate this simpler button, automated testing is only a pre-requisite. I think what you might be suggesting is that the simpler "form-button" is here to stay because it will never be cost effective to make `<bolt-button>` pass tests in a Drupal Webform use case. I think I agree.

sghoweri:
Out of curiosity @remydenton and @krlucas, has anyone looked into the duplicate data attributes + placement of the Drupal-added classes on the HTML generated by the button component's Twig template (ie. I'm just looking at the vanilla HTML that's on the page before any Bolt or Drupal-specific JavaScript kicks in):

### Bolt v1.x
```html
<div data-drupal-selector="edit-actions" class="form-actions webform-actions js-form-wrapper form-wrapper" id="edit-actions">
  <bolt-button align="center" color="primary" size="medium" width="full" transform="none">
    <button class="webform-button--submit button button--primary js-form-submit form-submit c-bolt-button__root"
      data-drupal-selector="edit-actions-submit" data-disable-refocus="true" type="submit" id="edit-actions-submit"
      name="op" value="Contact me">
      <replace-with-children class="c-bolt-button  c-bolt-button--medium c-bolt-button--full  c-bolt-button--primary c-bolt-button--none c-bolt-button--center">
        <span class="c-bolt-button__item ">Contact me</span>
      </replace-with-children>
    </button>
  </bolt-button>
</div>

### Bolt v2.x
```html
<div data-drupal-selector="edit-actions" class="form-actions webform-actions js-form-wrapper form-wrapper" id="edit-actions">
  <bolt-button align="center" color="primary" size="medium" width="full" transform="none" data-drupal-selector="edit-actions-submit" data-disable-refocus="true" name="op" value="Contact me" class="webform-button--submit button button--primary js-form-submit form-submit">
    <button data-drupal-selector="edit-actions-submit" data-disable-refocus="true" type="submit" id="edit-actions-submit"
      name="op" value="Contact me" class="c-bolt-button  c-bolt-button--medium c-bolt-button--full  c-bolt-button--primary  c-bolt-button--center " is="shadow-root">
      <replace-with-children class="c-bolt-button__icon is-empty"></replace-with-children>
      <replace-with-children class="c-bolt-button__item ">Contact me</replace-with-children>
      <replace-with-children class="c-bolt-button__icon is-empty"></replace-with-children>
    </button>
  </bolt-button>
</div>
```

krlucas:
@sghoweri In trying to troubleshoot the button and Drupal AJAX, etc. I played around with hacking the Bolt twig template locally to put the attributes back in the same places they were with Bolt 1.x. It didn't seem to help with the binding of the Drupal events. I also didn't spend THAT much time on it.

sghoweri:
@remydenton two things I wanted to bring up in this PR's description:

1. Bolt buttons haven't been using Preact for over 9 months now -- they are using HyperHTML specifically so the virtual DOM wouldn't be something that would get fingers pointed out. 

Even Bolt v0.x's button uses [HyperHTML](https://github.com/bolt-design-system/bolt/blob/2650355b701bc838cdfa1ea49b249137d9b2d06a/src/_patterns/02-components/bolt-button/src/button.standalone.js#L35).

2. If the `<bolt-button>` component's _JavaScript_ was somehow causing, http://vjira2:8080/browse/WWWD-2711, why did [this hotfix ](https://github.com/bolt-design-system/bolt/pull/926) update the Band component's Twig file? Does this mean we can revert the change made in #926?

charginghawk:
>If the <bolt-button> component's JavaScript was somehow causing, http://vjira2:8080/browse/WWWD-2711, why did this hotfix update the Band component's Twig file? Does this mean we can revert the change made in #926?

release/phase-10 broke webform functionality in at least 5 ways:

1. Duplicate submissions (resolved by #926)
2. AJAX (resolved by http://vstash:7990/projects/PM/repos/pega_bolt/pull-requests/159/overview)
3. Open download in new tab (unresolved, http://vjira2:8080/browse/WWWD-2739)
4. Moving submit button to mobile popup (unresolved, http://vjira2:8080/browse/WWWD-2709)
5. Submit intermittently not disabled after first click (unresolved, http://vjira2:8080/browse/WWWD-2738)

So, #926 is a necessary fix, but there's a lot of stuff broken. Unfortunately, it seems like each issue has required individual solutions thus far - no two-birds-with-one-stone fixes.

remydenton:
Thanks for the in-depth review @sghoweri.  Some responses:

1. I initially tried using the button mixins, but found they ultimately weren't adequate for recreating a basic button (the `bolt-button()` mixin, for instance, doesn't include the default values for `.c-bolt-button--medium` and `.c-bolt-button--center`).  Since the main goal is to remove javascript from the equation, I don't think extending and limiting config options would help.  Long story short, I grant that the solution in this PR does have the shortcomings you mention, but they seemed the lesser of many different evils I considered.
2. Completely agree this would benefit from a schema.  This might be a good opportunity to revive #695 to get multiple schemas working.
3. Good point, I will add bolt-button as a dependency of bolt-form.
4. Those limitations were actually intentional based on these two ideas:
  - The fewer options the form button accepts, the simpler it is.
  - Adding additional options later is much easier than removing them (and/or maintaining options that might or might not be used)
  - I have not yet seen a need for all but the simplest options for form buttons.  Note, I don't see the secondary colored button in your screenshot on pega.com homepage-- how do I find that?
5. Not sure I totally follow.  Are you saying all or most of the button options come directly from stakeholder requests (and therefore there's reason to think we should support most of them for form buttons)?
6. This is the single most important point.  I completely agree that this sets a precedent for fragmentation, and I'd like to avoid that too.  But the flip side of that is that we've spent a lot of time discovering and troubleshooting these errors.  That was one thing the first time we build forms (a necessary evil), but when they show up as regressions, it's a whole new level of pain and frustration.   If I'm honest, the long-term idea behind this PR is that we need a javascript-free version of some components.  That doesn't mean we can't also have a web component version of the button, especially for use in things like Bound, but that at some point the cost-benefit analysis of having only a web component version stops making sense.

Before we merge this, let me spend a little time testing the changes in Drupal to confirm that it addresses the issues in @charginghawk's comment (and investigating your first suggestion a bit more as well).

Re: hyperhtml vs preact, my apologies.  I clearly get my terminology mixed up and mentally lump together things that are javascript-rendered.

Re: #926, I'm only suggesting that if the button wasn't rendered with javascript, we might have avoided the duplicate submissions despite the problems with the band logic.  At the very least, I think we would have gotten to the solution we did faster if there were fewer potential points of failure.

remydenton:
I've re-tested the Drupal-side bugs @charginghawk mentioned with and without the changes in this PR.  Here are the results:
1. :heavy_multiplication_x: Not fixed.  If I undo the fix in #926 and disable AJAX, I get duplicate submissions whether the button is a web component or not.  Fortunately, it's a non-issue since #926 has already fixed this, but it's significant in that it means this PR is not a silver bullet that will just fix everything.  Also, @sghoweri, you were right :).
2. :heavy_minus_sign: N/A. Since we're not using `<bolt-button>` anymore, there's no need to re-attach js to it.  AJAX does seem to work as expected though with the simple button.
3. :heavy_check_mark: Fixed. The new tab behavior and AJAX work together with the simple button.
4. :heavy_check_mark: Fixed. The simple button never disappears when switching between mobile and desktop.
5. :heavy_minus_sign: Unclear.  I wasn't able to reproduce this issue.  It might just be that I didn't try enough times-- @charginghawk reckoned about a dozen, but I got bored before I ever saw it happen.

So overall, it does solve many of the outstanding problems.

Importantly though, Drupal isn't blameless here.  While the webform module javascript (AJAX) is non-negotiable, some of the other things we're doing on the Drupal side are custom and questionable-- namely, the open in a new tab functionality and the move-the-button-around-when-switching-to-mobile are both unorthodox.  I don't want this workaround on the Bolt side to just enable more code like that.  To that end, I'd like to see:
1) An attempt to clean those up (specifically, how much effort would it be to nix the move-mobile-button code in favor of a simple separate mobile form @charginghawk?)
2) Automated regression tests (hi @krlucas :grin:).  This is a virtual necessity before we can consider going back to `<bolt-button>`.  It also reduces monotonous testing (like I just did here), and most importantly _defines how it all should work_.  It's really not fair for the Drupal side to blame Bolt for breaking functionality that isn't clearly spelled out anywhere (e.g. "submitting a form with AJAX enabled needs to open a new tab without triggering the browser's popup blocker").  Once we do have these tests, there's at least the possibility that the Bolt team could set up scenarios in Drupal lab to reproduce, fix, and prevent these kinds of errors.

krlucas:
Thanks Remy. It's not worth debating too much but whatever is "unorthodox"
about the new tab functionality had nothing to do with what broke (click
event targeting, binding and catching).

Andy and I worked through a similar  click binding conundrum with CDH
analytics. I got it working but it might be good to run through what I had
to do with you and Salem.

remydenton:
The only "unorthodox" part of that is that it's independent of form submission (i.e. instead of being a callback to the AJAX submission, it adds its own click handler and checks independently for form validation before opening a new tab).  @charginghawk proposed that could set us up for some kind of race condition between AJAX form submission and the new window javascript.  Not confirmed and it's not really unorthodox (two click handlers should be supported), but a test would still help in case it's something not as simple as "the button needs to support two click handlers".

charginghawk:
The move-the-button-around-when-switching-to-mobile is very jank. At the time, it was implemented because the form had to be absolutely positioned, and bands still had transform: none on them that made absolutely positioning it while it was in a band impossible. The band issue been resolved, so we could now implement a much cleaner "on mobile, absolutely position the webform" solution that wouldn't even require javascript (and yes, I would like to).

I think it just goes to show that the webforms were really complex and they only worked because of a confluence of bugs and jank requiring/enabling/covering each other. As we gradually fixed bugs, bugs hidden underneath surfaced, and technical debt from the crunch to pegaworld brought low our jenga cards house tower.

In terms of making it easier for ourselves moving forward, one lucky break was that on the Drupal side, we were able to bisect the band updates from the rest of the Bolt 2.x beta releases. If we're planning on making major updates to bands or buttons or other critical/ubiquitous components, we could space them out across different Drupal-side releases, so they get QA'd and stress tested (and hotfixed) individually. Really, we shouldn't have updated bands at all - it was only to accommodate the share tool, and it was a costly/hazardous overhaul in response to adhoc designs pieced together without designer or developer input.

remydenton:
Thanks to everyone for the feedback.  I'm going to go ahead and merge this, recognizing that it's not the end of the discussion.

Heads up @sghoweri, I've created a ticket for multiple schema per component: http://vjira2:8080/browse/BDS-746

PR #947, opened 2018-11-08
Stylelint CSS props order

mikemai2awesome:
## Jira

http://vjira2:8080/browse/BDS-356

## Summary

Turns on stylelint's CSS Properties Order based on Bolt's convention.

## Details

Added and re-organized CSS props in the stylelint config, and updated all SCSS files based on that.

## How to test

1. Run the branch locally.
2. Click on every single page for any visual regression.
3. Inspect the CSS changes and make sure they are only re-ordering the props, make sure nothing was removed by mistake.


mikemai2awesome:
@remydenton 

It does take a little while to get used to, but I don't see that being a problem.

I don't believe we need to introduce flexible order, that defeats the purpose. With the directional props, I made sure everything is going clockwise, that should be enforced to keep the consistency. 

At the end of the day, `npm run lint:style:fix` re-orders them automatically, even if you can't get used to the order for whatever reason, you can just run that command and be done with it.

remydenton:
Thanks @mikemai2awesome, adding the clockwise convention helps.  Let's merge this and see how it goes-- worst case scenario, we can just relax the criteria later.

@sghoweri, @danielamorse, @adam2661, give a shout if this turns out to be frustrating so we can revisit it.

PR #958, opened 2018-11-16
Fix: re-organize mockup pages

mikemai2awesome:
## Jira

http://vjira2:8080/browse/BDS-767

## Summary

Cleaned up outdated mockup pages in Pattern Lab.

## Details

Removed and updated a lot of mockup pages in PL to reduce confusion for QA. Certain pages were renamed and re-ordered, therefore those urls are changed, will need to update the links for QA once this is merged.

## How to test

Run the branch locally and check all the pages under Pages in the navigation.


apps/pattern-lab/src/_patterns/04-pages/20-d8-press-and-media/-15-d8-media-browse.twig
```diff
@@ -0,0 +1,3 @@
+{% include "@pl/_d8-news-browse.twig" with {
+  "page_title": "Media Coverage"
+} %}
```
remydenton: One problem with prefixing the main pattern with an underscore (`_d8-news-browse.twig`) is that there is now no way to see the source info for the press and media pages in pattern lab.  

Before:
screen shot 2018-11-20 at 7 12 54 am

After:
screen shot 2018-11-20 at 7 13 11 am

I think we need to have the source viewable one way or another.  The Agenda Manager pages use a similar strategy-- consider matching that one or updating to use whatever strategy you settle on.

mikemai2awesome: @remydenton 

If you look at how the Media Browse and Press Browse was set up previously (https://boltdesignsystem.com/pattern-lab/?p=pages-d8-press-browse--list-view), it's showing exactly as such: 

```twig
{% include "@pl/-d8-media-browse--list-view.twig" with {
  "page_title": "Press Releases"
} %}
```

It has nothing to do with the underscore, just how that twig was set up. I really don't like this method anyway, I would rather not include any internal partial, it does not benefit the viewer.

mikemai2awesome:
@remydenton 

I removed the include completely.

remydenton:
I hate to delay this any further, but I feel like now it's got a bunch of duplicate code.

I _think_ I prefer the approach in Agenda Manager where you have [a default page where you can inspect the code](https://boltdesignsystem.com/pattern-lab/?p=pages-agenda-manager-detail) and then [other](https://boltdesignsystem.com/pattern-lab/?p=pages-agenda-manager-detail--alan-keynote) [pages](https://boltdesignsystem.com/pattern-lab/?p=pages-agenda-manager-detail--aflac) that override small pieces for demo purposes.  Anyone building from it can just look at this first page.

@sghoweri, any preference?

mikemai2awesome:
@remydenton 

My problem with that method (https://boltdesignsystem.com/pattern-lab/?p=pages-agenda-manager-detail--aflac), this is still confusing:

```twig
{% include "@pl/-agenda-manager-detail.twig" with {
  pageTitle: "What Aflac is Learning About Consumer Expectations Through Monitoring Tech Disruptions",
  thumbnailPath: "logos/logo-aflac.png"
} only %}
```

The expectation for the code snippet in that panel is that it should be showing how you would use the component includes or embeds, internal includes like `@pl/-agenda-manager-detail.twig` is not something we ship, it's not usable. 

I like to reduce duplicate code and use includes to make it easier for us, but not at the expense of the consumer. You are seeing it from a POV where you know exactly what's going on, someone less savvy would not understand that. @remydenton 

Unless we can fix Pattern Lab, I'd rather write out the actual includes and embeds, instead of creating internal partials.

mikemai2awesome:
@sghoweri @remydenton 

Take a look at any design system site out there, the code panel from a component is supposed to showcase working code sample, there is a copy button next to the snippet so you can copy and paste, and expect that to work.

- https://polaris.shopify.com/components/forms/autocomplete#navigation
- https://www.carbondesignsystem.com/components/button/code

Including things like `@pl/whatever.twig` is not a piece of working code sample.

remydenton:
I see your point for sure @mikemai2awesome.  I just don't think opting for code duplication is the obvious choice, mostly because it has a big downside: it makes it harder to maintain the examples and increases the opportunity for mistakes.  Documentation with mistakes is the worst!  Worse than documentation that requires you to figure out what `@pl` means and find a different page, IMO.

Marking this as approved, but this is an outstanding problem we should address.  @sghoweri @danielamorse @margoromo

remydenton:
Wait, hold on-- I think we should actually be using json files for this use case (i.e. multiple instances of the same twig structure, just with different data).  I'm going to merge this PR and create a separate one to demonstrate that.

remydenton:
Follow-up PR is at #969.

PR #969, opened 2018-11-21
Use pseudo patterns to reduce redundant twig markup

remydenton:
## Jira

http://vjira2:8080/browse/BDS-767 (sort of).  See discussion on #958

## Summary

Removes redundant markup in press and media (WIP)

## Details

See discussion on #958 for background

This PR uses [pseudo patterns](https://patternlab.io/docs/pattern-pseudo-patterns.html) to accomplish the following two things at once:
- Avoid redundant twig templates
- Show the correct source twig in Pattern Lab's "pattern info" feature

This technique could also be used for other sections including:
- Agenda manager detail page examples
- Blog posts page examples

One drawback to the way I'm doing it in this PR is that it creates an additional default page at `/pattern-lab/?p=pages-d8-browse`.  One way around that would be to turn the default page into one of the two examples using a [markdown file](https://patternlab.io/docs/pattern-documenting.html) to properly set the pattern name.

## How to test

- Go to `/pattern-lab/?p=pages-d8-browse-press` and `/pattern-lab/?p=pages-d8-browse-media`
- Click 'Show Pattern Info' in the PL UI and confirm that both pages show the full source twig


mikemai2awesome:
This is great!

Just one thing, showing the Browse template page is confusing, there's no need for that. Can we hide that with an underscore?

sghoweri:
> @remydenton
> 
> This is great!
> 
> Just one thing, showing the Browse template page is confusing, there's no need for that. Can we hide that with an underscore?

@mikemai2awesome off the top of my head I don’t think that would work (although I could totally be wrong on this) — if I remember correctly, out of the box, templates with underscores would be excluded from compiling (so the pseudo patterns here wouldn’t work)...

Lemme do some digging though since hacking away at Pattern Lab PHP to improve this sort of thing is right up my alley 😛

sghoweri:
Hmmm 🤔 
https://github.com/pattern-lab/patternlab-php-core/blob/develop/src/PatternLab/PatternData/Rules/PseudoPatternRule.php#L50

remydenton:
I agree the extra page is undesirable.  The workaround I proposed in the PR description is basically this:
- Instead of having two json files (e.g. `-05-d8-browse~press.json` `-15-d8-browse~media.json`), we make one of them the default.  So for example, remove `-05-d8-browse~press.json` and change [`page_title: page_title|default('Page Title')`](https://github.com/bolt-design-system/bolt/pull/969/commits/51b41560ebe04241a3b1d5f56faa47f7db53d5c7#diff-5ff1546a4415727d5debcd8028c2baefR46) to `page_title: page_title|default('Press Releases')`
- Use a [markdown file](https://patternlab.io/docs/pattern-documenting.html) to change the title of this pattern so it's less confusing.

Thoughts?

mikemai2awesome:
@remydenton 

With the markdown solution, will there be 2 files `-15-d8-browse-press.md` and `-15-d8-browse-media.md`?

sghoweri:
+1 on making one the default

remydenton:
No, just one .md file @mikemai2awesome.  Its only goal is to change the title of the pattern that shows up in Pattern Lab from `d8 browse` to `d8 browse press`.

screen shot 2018-11-26 at 9 29 41 am

I'll give that a shot and update the PR

remydenton:
I've pushed some commits that re-purpose the default browse page as the `D8 Browse Press` page and remove one of the json files.

One weird thing I discovered is that markdown files don't seem to work if the twig file is prefixed with a `-`.  It's a bug in Pattern Lab as far as I can tell, albeit a minor one.

remydenton:
Agree with @mikemai2awesome that it's preferable to not have leading dashes on all files if we can help it.  I've pushed an update.

The one problem is that I now see these pages showing up at `/pattern-lab/?p=all`.  @sghoweri, you said that is not supposed to happen, right?

remydenton:
Okay, I think this is ready!  I updated the PL configuration to exclude all page patterns, removed now-unnecessary leading dashes from all page patterns, and refactored some of the other pages using the new json data strategy to simultaneously reduce redundant twig and show source code in the PL UI.

PR #970, opened 2018-11-21
Change "url" to "href" in link schema

danielamorse:
## Jira

 - http://vjira2:8080/browse/BDS-786

## Summary

Find instances of link includes and change "url" to "href".

## Details

Update instances of link includes (components and patterns) to reflect the new schema. Note: for now we're only making this update to *links*. Other components will be updated later.

## How to test

 - Review the files changed and confirm only link components have been updated from "url" to "href"
 - Go to https://boltdesignsystem-iynnpluhjz.now.sh/pattern-lab/index.html and confirm that the link patterns still work, i.e. their href's are populated.
 - Review a few other components (for example: action blocks, nav, breadcrumb) and confirm the same.

### DIFF
--- a/apps/pattern-lab/src/_patterns/02-components/block-list/00-block-list-docs.twig
+++ b/apps/pattern-lab/src/_patterns/02-components/block-list/00-block-list-docs.twig
@@ -3,7 +3,7 @@
   items: [
     include("@bolt-components-link/link.twig", {
       text: "Link 1",
-      url: "#!",
+      href: "#!",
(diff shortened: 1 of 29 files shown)


danielamorse:
@sghoweri @mikemai2awesome after some discussion, deciding to leave this one as is. We'll aim for one-to-one mapping of standard HTML attributes but in this case, where "url" is already widely in use, let's allow both, and not recommend switching to "href".

PR #985, opened 2018-12-10
Prevent escaping in all languages

adam2661:
## Jira

http://vjira2:8080/browse/BDS-543

## Summary

Apply escaping to all languages in code snippet component.

## Details

Changes are only in a js file. It's simple remove of variable and escape without checking language.

## How to test

Run this code locally. Go to `localhost:3000/pattern-lab`. Navigate: `Styleguide > Sassdocs > View All `then on right navigation Tools: `typhography > @mixin bolt-font-family` and/or `Tools: typhography > @mixin bolt-font-size` and check if `&, <, >` are showed unescaped. You can create new pattern-lab page and apply to it: `{% include "@bolt-components-code-snippet/code-snippet.twig" with {
  display: "block",
  lang: "html",
  content: "&lt;bolt-list&gt;
  &lt;bolt-list-item&gt;&lt;/bolt-list-item&gt;
  &lt;bolt-list-item&gt;&lt;/bolt-list-item&gt;
  &lt;bolt-list-item&gt;&lt;/bolt-list-item&gt;
&lt;/bolt-list&gt;"
} only %}` to check if all languages are escaped please chenge `lang` and `content` variable.

### DIFF
--- a/apps/pattern-lab/src/_patterns/02-components/code-snippet/10-code-snippet-display-variation.twig
+++ b/apps/pattern-lab/src/_patterns/02-components/code-snippet/10-code-snippet-display-variation.twig
@@ -1,10 +1,15 @@
 {% set schema = bolt.data.components['@bolt-components-code-snippet'].schema %}
+{% set code %}{%- verbatim -%}
+<header>
+  <h1>Headline</h1>
+</header>
+{%- endverbatim -%}{% endset %}
 
 {% for display in schema.properties.display.enum %}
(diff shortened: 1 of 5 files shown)


mikemai2awesome:
@adam2661 

I did some testing and it is still not working well. Take a look at the following:

### :white_check_mark: SCSS/CSS is working

scss

### :x: HTML is not working as expected

html

There is a discrepancy between using the `twig` vs the web component, can this be fixed?

cc @danielamorse

PR #986, opened 2018-12-12
Add display flex to bolt list component

adam2661:
## Jira

http://vjira2:8080/browse/BDS-707

## Summary

Adding possibility to set display prop to flex that gives an opportunity to set dividers horizontal and vertical on items.

## Details

I'm not 100% sure that @mikemai2awesome want to do this in this way but i don't see any other way to do that.

## How to test

Run this code locally. Go to: `Components -> List -> View All` and check if all variations on flex list are working correctly.


adam2661:
Hi @mikemai2awesome you have right. Now spacing and inset spacing is handled by padding. I will update this.

adam2661:
@mikemai2awesome i remove the border on entire list element but i see a problem with spacing using margin. I try to manage it by adding only dividers between elements but without good effect. In this case, i leave a border around the element like on the images below.
image
image
Please give me information what you think about it or/and any suggestions.

adam2661:
The code is updated as we discussed on Friday.

PR #987, opened 2018-12-12
Replace rounded property with border radius in buttons

remydenton:
## Jira

http://vjira2:8080/browse/BDS-660

## Summary

Replaces button's `rounded` property that should have been boolean with a `border_radius` property that accepts multiple values

## Details

See the jira ticket for the rationale behind replacing rounded="rounded" (a string that should have been a boolean) with a `border_radius` (`border-radius` for web components) prop that accepts a number of strings.

The possible values for border-radius are the following.  The names of these properties are a weak point--I'm open to suggestions:
`none`
~`small`~
~`medium`~
`full`

*Update:* changed to the following based on PR feedback:
~`none`~
`regular`
`full`

*Update 2:* changed to the following based on additional feedback:
`regular`
`full`

Other notes:
- I discovered that the schema format for deprecating multiple components is different than I had used previously.  We should update the wiki as well as other components with multiple deprecations. 
_Edit: I updated [the deprecation wiki page](https://github.com/bolt-design-system/bolt/wiki/Deprecating-component-properties)_
_Edit 2: I created a PR to fix other components with this issue: #988_
- Should we consider a global border radius setting for each of the possible values?  We already have a single `$bolt-border-radius` global variable which corresponds to the `small` value for border-radius introduced in this PR.
_Edit: I created a follow up issue to address this: http://vjira2:8080/browse/BDS-835_

## How to test

- Review the demos of the new property at `/pattern-lab/patterns/02-components-button-05-button-border-radii/02-components-button-05-button-border-radii.html`
- Confirm that all usages of "rounded" have been removed
- Ensure that the deprecated property still works for now, both for twig and web components
```
This button should still work for now even though it's deprecated
<bolt-button rounded="rounded">Rounded button</bolt-button>

This button is the new way of making the pill style buttons
<bolt-button border-radius="full">Full Radius</bolt-button>

The border-radius property should override rounded if both are specified
<bolt-button border-radius="small" rounded="rounded">Small radius</bolt-button>
```

## Deprecations

The `rounded` property for buttons is deprecated and will be removed in a future version of Bolt.  Use the new `border_radius` property with a value of `full` instead.


packages/components/bolt-button/src/_button.settings.scss
```diff
@@ -6,7 +6,12 @@ $bolt-button-shadow: $bolt-shadow--small;
 $bolt-button-shadow--hover: $bolt-shadow--large;
 $bolt-button-border-width: $bolt-border-width;
 $bolt-button-border-style: $bolt-border-style;
-$bolt-button-border-radius: $bolt-border-radius;
+$bolt-button-border-radii: (
```
mikemai2awesome: How about a separate ticket to deal with a border system? Border styles and border radius are repeated in quite a few components, might be a good idea to make this reusable.

remydenton: Makes sense, I'll stub one out

remydenton: Issue created at http://vjira2:8080/browse/BDS-835

remydenton:
Removed "none" option as discussed during scrum today.  CC @sghoweri @mikemai2awesome

remydenton:
Wait, did we decide against "small" and "medium" in favor of "regular"?

PR #1009, opened 2018-12-28
WIP BDS-789 card adaptor

remydenton:
## Jira

http://vjira2:8080/browse/BDS-789

## Summary

Work-in-progress adaptor for card.  CC @adam2661


packages/components/bolt-card/src/card.twig
```diff
@@ -1,77 +1,206 @@
-{% set schema = bolt.data.components["@bolt-components-card"].schema %}
+{# If the props being passed match the deprecated schema, use the adaptor for the deprecated card.  #}
 
-{% if enable_json_schema_validation %}
-  {{ validate_data_schema(schema, _self) | raw }}
-{% endif %}
+{% if contentItems %}
```
remydenton: Right now, the only indication that the old card should be used is if `contentItems` is present.  That should at least also include a check for the twig blocks that exist in old cards.  Anything else?

Another idea I considered was to add an extra flag that would allow the consumer to explicitly indicate whether it was an old or new card they were using.  e.g. if they set a `is_new_card` param to `true`, it would skip the adapator layer for the old card and they could be sure it would work when the adaptor was eventually removed.

packages/components/bolt-card/src/card.twig
```diff
@@ -1,77 +1,206 @@
...
+          {% include pattern_template(item.pattern) with item only %}
+        {% elseif key != "content" %}
+          {% include pattern_template(key) with item only %}
+        {% else %}
+          {{ item }}
+        {% endif %}
+      {% endfor %}
+    {% endblock card_body %}
+    {% if block('card_footer') %}
+      <hr style="margin-left: -2rem; margin-right: -2rem; width: calc(100% + 4rem)">
```
remydenton: I think we decided we could just leave off the negative margin and accept that the separator won't go edge-to-edge.  Is that correct @mikemai2awesome?

packages/components/bolt-card/src/card.twig
```diff
@@ -1,77 +1,206 @@
...
+      content: cardMedia,
+    },
+    body: {
+      content: cardBody,
+    },
+    actions: cardActions
+  } only %}
+{% else %}
+  {# @TODO In the next major version of Bolt, remove the adaptors and replace the contents of this file (card.twig) with the contents of _adaptor-for-new-card.twig #}
+  {% include "@bolt-components-card/_adaptor-for-new-card.twig" %}
```
remydenton: Should we rename `_adaptor-for-new-card.twig`?  It is not really an adaptor-- maybe just `_new-card.twig`?

As @adam2661 and I discussed, it would be ideal to have all the code for the old card to be in a separate file as well, but we weren't able to allow the twig blocks to overridable when doing that (open to suggestions if anyone has them).

apps/pattern-lab/src/_patterns/02-components/card/90-deprecated-cards.twig
```diff
@@ -0,0 +1,725 @@
+<div style="max-width: 500px; padding: 2rem; margin: 1rem auto">
```
remydenton: Once this work is all done, I think we should actually remove this file of deprecated examples.  We certainly don't want people referring to it for new work, and ideally we should update the card schema to error out when the deprecated format is used in pattern lab.

It's too bad that means we can't keep it for testing... maybe just prefix it with an underscore so it can easily be restored when doing any additional work on this task?

packages/components/bolt-card/src/card.twig
mikemai2awesome: Remove this. I've taken care of all borders in my latest updates.

apps/pattern-lab/src/_patterns/02-components/card/90-deprecated-cards.twig
mikemai2awesome: Agree. I will take a look at all the docs again once this is merged into my branch.

packages/components/bolt-card/src/card.twig
mikemai2awesome: Hold on, not sure if I understand this part. This is changing the old card?

remydenton: Right-- this code takes any existing usages of card that override the `card_footer` block and adapts them so that they will work with the new card (all of the code for which has been moved to `_adaptor-for-new-card.twig`)

adam2661: Instead add inline style we can add some class for that or we can pass remove `hr` from our adapter and if someone override block footer and will what have divider bettwen content then can pass `hr` or something else in block footer. Please write what you think about it.

packages/components/bolt-card/src/card.twig
adam2661: It's done.

apps/pattern-lab/src/_patterns/02-components/card/90-deprecated-cards.twig
adam2661: I agree with your thoughts @remydenton  i can underscore this but if @mikemai2awesome want to look and update docs maybe will be better to leave this right now like it is. Then Mike updates this in his PR.

packages/components/bolt-card/src/card.twig
adam2661: @remydenton right now i'm checking `contentItems`, blocks `body` and `footer`

packages/components/bolt-card/src/card.twig
adam2661: @remydenton After a lot of tries, I was able to move all code for old card to separate file. I achieved that by creating two variables in the main file and then check them in old card adapter file.

apps/pattern-lab/src/_patterns/02-components/card/90-deprecated-cards.twig
mikemai2awesome: Yup. Leave it to me.

remydenton:
Looking really good @adam2661!  Nice work getting the old card code moved to a separate file with the blocks still working.

I just spent some time digging through the pega.com Drupal codebase for card usages.  I added a few of them to our deprecated examples to demonstrate some things the adaptor is still missing.  Specifically:
- We need to support overriding of the `card_media` block in addition to `card_body` and `card_footer`.
- There are some other props we need to whitelist so that they get passed through-- for example, `attributes`.  Again, this is extra tricky because the old schema wasn't complete (it didn't mention an attributes prop...).  I think the best way to make sure we've got them all is to look at the old card source code for any variables that were in use.
- We should use @mikemai2awesome's new solution for freeform footer content.  That will let us get rid of the `<hr>` workaround, and might help solve some other possible edge cases (one of the examples I pulled in from Drupal looks like it has too much padding).

In theory, we should be able to account for all possible existing usages just by looking at the old card.twig and making sure anything it used to support still works.   That said, it may also be helpful to search through the pega.com codebase to see if you can find other examples of edge cases.  This is definitely turning out to be hard and it's possible there will be usages of card that we don't account for that break when the new card gets pulled in 😰.

packages/components/bolt-card/src/card.twig
adam2661: I use class from @mikemai2awesome example

adam2661:
@remydenton i cover all new examples like:
- overrides all blocks `footer, media, body`
- add attributes handling
- remove `hr` workaround

packages/components/bolt-card/src/card.scss
```diff
@@ -90,7 +90,6 @@ bolt-card {
 // Old card footer
 
 .c-bolt-card__footer {
-  @include bolt-padding(small #{$bolt-card-body-spacing});
```
remydenton: @mikemai2awesome, heads up that we removed the padding from the footer class.  Apparently, there was no padding on it previously, which allowed the footer to be used as a block in cases like the cards at the top of this page:

https://www.pega.com/about/news/press-releases

As you can see, the Drupal team added their own padding when using freeform text in the footer...
hhttps://www.pega.com/services/partnerships/isv-program

mikemai2awesome: Cool, then that's on them. They will need to keep track of manual work like this and expect to QA it whenever there are component updates.

remydenton:
@adam2661, hopefully this is the last thing: I found that the pega.com site is using [@bolt/card-w-teaser-list.twig](http://vstash:7990/projects/PEGAWWW/repos/pegawwwd8/browse/docroot/themes/custom/pegawww_theme/templates/node/node--resource.html.twig#164-166).  That was a twig file that used to exist but was deleted in the refactor, along with `card-w-teaser.twig`.

It's kind of a pain but I think the right thing to do is add those two twig files back as adaptors, including
a) a comment at the top of the twig file that they are deprecated and will soon be removed
b) make sure they work (I don't think there are any existing examples of their usage in PL, so we'll probably have to reverse engineer and add a couple examples to our deprecated example).

@mikemai2awesome, does that sound like what you were thinking too?

charginghawk:
@remydenton I didn't even know we were using card-w-teaser-list! But I can confirm that we aren't using card-w-teaser - it stopped working correctly so we just switched to using card.twig:

http://vstash:7990/projects/PM/repos/pega_bolt/pull-requests/157/diff

mikemai2awesome:
@remydenton @adam2661 @charginghawk 

If we are retaining the `card-w-teaser-list.twig`, please stop using `ui-list`, use the new `bolt-list` set to block display with separators.

charginghawk:
@mikemai2awesome teaser-w-card mostly worked. There was a weird issue where CTA chevrons weren't appearing, but it was otherwise mostly functional.

danielamorse:
@remydenton, @adam2661, I finally reviewed this PR now that the `bolt-card-link` updates have been pulled in. I tested in Chrome and IE, especially the deprecated card examples, and reviewed the adapter code itself. Everything looks good!

adamszalapski:
@remydenton `card-w-teaser.twig` and `card-w-teaser-list.twig` right now is habdled by adapter.
@mikemai2awesome for `card-w-teaser-list.twig` i use `bolt-list` component.

mikemai2awesome:
@remydenton @adamszalapski 

Am I good to merge this into the card refactor branch?

PR #1080, opened 2019-02-14
Bolt chip list documentation and testing

adamszalapski:
## Jira

http://vjira2:8080/browse/BDS-955

## Summary

Add testing documentation and update component documentation

## Details

I update this component tobe using `list` component instead of `inline-list` for generation list of chips. But i want to point that this component it's pointless from my perspective we have `list` component and we have even examples how to use it with chips. 
https://boltdesignsystem.com/pattern-lab/patterns/02-components-list-10-list-item-variations/02-components-list-10-list-item-variations.html
I think that we should deprecate this component and then remove from components list. Please give your opinion on this subject @mikemai2awesome, @danielamorse, @remydenton, @sghoweri.
When i'm thinking more about it we can remove `bolt-list` component to because it's duplicated with `list` component.

## How to test

Run this code locally and check if this component renders correctly. Then run test locally by typing `npm run test:js` in terminal.


mikemai2awesome:
Per our discussion, I am in favor of deprecating `inline-list`, `bare-list`, `block-list`, `ui-list`, `chip-list`, `button-group`, they should be examples but not components or objects.

PR #1102, opened 2019-03-04
Bolt icons testing

adamszalapski:
## Jira

http://vjira2:8080/browse/BDS-952

## Summary

Write tests and updating configuration of jest tests

## Details

## How to test

Run this code locally and check if all jest test are passing: `npm run test:js`


adamszalapski:
> @adamszalapski all tests passed, but I saw this error:
> 
> ```
> Error: ENOENT: no such file or directory, open '/Users/morsd/sites/bolt/packages/components/bolt-icons/src/icons/yeti.js'
> ```
> It's looking for the icon in "src" not "__tests__". Is this expected?
> 
> image

Thanks, @danielamorse is expected behavior. After compiling icons we remove the icon that was created only for these tests. Please check if you have this three task before all tests.  
image I will investigate what is going on. The biggest problem is when is running `npm run test:js` the error is occurring but if use ` npx jest packages/components/bolt-icons/__tests__/icons.js` to run only this test, the error is gone. Maybe @sghoweri have some quick fix for that. If not i will try to fix this by myself.

danielamorse:
@adamszalapski yep, those three tasks run.

It would be better not to get this error, as it makes me think that something is broken when I run the tests. If it's not too much work, can we try to fix?

I also noticed:
- When I run `npm run test:js`, `packages/components/bolt-icons/src/index.js` gets modified with the line:
```
+ export * from './icons/yeti';
```
- When I run `npm start` these two files are also modified/added:
```
m packages/components/bolt-icon/icon.schema.yml
+ packages/components/bolt-icons/src/icons/yeti.js
```

Ideally, git status would be clean after running `start` or `test`. Is that doable?

adamszalapski:
@danielamorse and @sghoweri all issues that you guys raised should be fixed. Please try to check this PR one more time.

adamszalapski:
@danielamorse, in my opinion, is not related to `chmod` but with time when files are created. As far as i know, JS doesn't know when a file is created fully he knows only that command to create a file is finish. And `index.js` want to access files that not yet created. I will try to fix this but really i was thinking that i fix this because on my end this behavior is not occurring.

danielamorse:
@adamszalapski ah, you are correct. I ran it again and the issue re-appeared.

adamszalapski:
> @adamszalapski ah, you are correct. I ran it again and the issue re-appeared.

Hi @danielamorse my assumptions were wrong. The problem is not that js don't know when files are fully saved but `button` tests use icons to and i recreate them after `icons` tests. Right now i move cleaning function to be run after all tests and i think that fix this issue. Please check on your end.

jest-global-teardown.js
```diff
@@ -3,12 +3,21 @@ const rimraf = require('rimraf');
...
 
 const DIR = path.join(os.tmpdir(), 'jest_puppeteer_global_setup');
+
 module.exports = async function() {
   // close the browser instance
   await global.__BROWSER_GLOBAL__.close();
+  const config = await getConfig();
+
+  config.iconDir = config.iconDir.filter(
+    item => !item.includes('test/jest-test-svgs'),
```
sghoweri: Ideally these assets would live with the icon task or test for the icons themselves BUT I'm ok with this as is (we can always iterate) 👍

PR #1201, opened 2019-06-11
Interactive Pathways | Component | Tooltip for the Interactive Pathways

joekarasek:
Having a little trouble with this one around the web component behavior. Looking forward to chatting with Salem when he gets back. 
1) The content of the tooltip (button text and bubble dialogue) is currently being set through attributes. If this is going to work with the Drupal Lingotek translation module we will need to refactor it so that it is passed in through slots (I believe)
2) On very narrow screens (under 400px), the tooltip is not respecting the edge of the screen.

### DIFF
--- a/packages/components/bolt-interactive-pathway/src/interactive-pathway.js
+++ b/packages/components/bolt-interactive-pathway/src/interactive-pathway.js
@@ -42,7 +42,7 @@ class BoltInteractivePathway extends withLitHtml() {
    * @todo this could use a refactor for performance
    * @param node
    */
-   _getParentBoltStepNode(node) {
+  _getParentBoltStepNode(node) {
     if (node.tagName === 'BOLT-INTERACTIVE-STEP') {
       return node;
     } else {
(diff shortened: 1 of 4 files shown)


joekarasek:
Also, not sure why `tooltip.scss` line 420 `width: 300px` is set. Seems to be part of the overflow issue on smaller screens.

margoromo:
@danielamorse - Can you review this to see if you can help out Basalt in Salem's absence?

mikemai2awesome:
@joekarasek In general, the tooltip component was not built up to our standards, it's been a pain to work with. I am all for a refactor.

Regarding the defined width, it's there to make sure long text would would as well, try to pass a few lines of text into the content and you'll see why.

You can also see in the CSS that the bubble can appear from all directions, not just up. This was never implemented, therefore it does not respect the edge of the screen.

joekarasek:
This has been addressed on the main epic branch for the micro journeys (formerly interactive pathways).

PR #1244, opened 2019-07-02
Hotfix: bolt button's incorrect display in IE11

mikemai2awesome:
## Jira

http://vjira2:8080/browse/BDS-1525
http://vjira2:8080/browse/BDS-1529

## Summary

Fixes an issue with the `bolt-button`'s display renders incorrectly in IE11.

## Details

There's an issue with the `bolt-button`'s CSS where the `:host` selector was used, and the CSS was written incorrect using scss extend. This removes that extend and rewrite the CSS manually. When writing `:host` selector CSS, it cannot be combined with other selectors. 

## How to test

Run the branch locally and check the following:

- [x] 1. Given I am viewing the [product landing page](http://localhost:3000/pattern-lab/patterns/04-pages-10-d8-product-pages-product-landing/04-pages-10-d8-product-pages-product-landing.html) in IE11.
- [x] 2. When I navigate near the bottom of the page, I should see a section with a blockquote, that comes right after the section with cards.
- [x] 3. Then I look at the `call to action` button below the blockquote. This button should have auto width and is centered aligned.


mikemai2awesome:
@remydenton @danielamorse I added a test for aligning inline buttons.

PR #1248, opened 2019-07-03
Video | QA | Fix background video bug in IE

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1528

## Summary

Fix a bug in IE which caused the edges of background videos to be hidden.

## Details

The background video bug appears to be related to Band rework. When the Band markup changed, a tiny bit of CSS no longer applied to the correct element. The `is-expanded` class used to be on the same element as `c-bolt-band` (the element `bolt-band`), but the new component pattern puts `c-bolt-*` classes on an inner element, not the outer custom element. So, the CSS rule needed to be updated.

## How to test
Run locally and verify the following:
- [x] Given I am on the [Background Video page](https://boltdesignsystem.com/pattern-lab/patterns/02-components-video-45-video-with-inline-script-and-external-controls-as-background/02-components-video-45-video-with-inline-script-and-external-controls-as-background.html) *in IE11* (use local URL)
- [x] When I click on the "Watch Video" button
- [x] Then I see the "Background Video" play and it is fully visible, i.e. the top and bottom are not cut-off.


packages/components/bolt-band/src/band.scss
```diff
@@ -29,6 +25,10 @@ bolt-band {
   width: 100%;
   min-height: 0;
   transition: min-height $bolt-transition; // Do not transition band widths
+
+  .is-expanded & {
```
mikemai2awesome: I strongly recommend when we are dealing with nested styles like this, use `@at-root` to lessen the chance for errors. Read about it here: http://sassbreak.com/getting-back-to-our-roots/

mikemai2awesome: Sorry, I meant to request changes, but hit approved instead. :P @danielamorse

PR #1295, opened 2019-07-27
WIP: Epics/with without

glassdimly:
## Jira

(Enter a link to the Jira issue this PR applies to, if any)

## Summary

Adds a with/without Pega page.

## Details

Adds a with-without component included from the with-without page... and lots of other things.


margoromo:
Added http://vjira2:8080/browse/BDS-1712 and http://vjira2:8080/browse/BDS-1713 and assigned to @sghoweri and @mikemai2awesome

margoromo:
@mikemai2awesome - this would not have corresponding overarching ticket, but perhaps the Asana epic could be referenced? They are managing this work on their end until it gets to the point of integration. I do have the code review tickets referenced here too which could also be used in that field.

mikemai2awesome:
@margoromo Gotcha, I was mainly looking for more information so I can get a good sense of context here. If this isn't ready to review yet, I will hold off on my deep dive into the code and UX concerns. I can see now there's WIP in the title.

glassdimly:
@mikemai2awesome I'm circling back here to your comments. Do I need to remove this as a component given its snowflake architecture? What's your suggestion here? Should this live directly under `packages`?

charginghawk:
Correct me if I'm wrong, but I believe the closest existing place for this kind of work would be the pages section:

https://github.com/bolt-design-system/bolt/tree/master/docs-site/src/pages/pattern-lab/_patterns/04-pages

That said, this isn't actually supposed to ever be merged in, so where it lives in the code base probably doesn't make a big difference. It can be a component in the Drupal theme, in any case.

mikemai2awesome:
@glassdimly 

I believe @sghoweri and @EvanLovely talked about creating an Experiments section. And as Thomas said, this is not getting merged in, it can be a component in the Drupal theme if need be.

PR #1306, opened 2019-07-31
Fix: correct close button color in IE11

mikemai2awesome:
## Jira

http://vjira2:8080/browse/BDS-1698

## Summary

Fixes the color of the modal's close button in IE11.

## Details

Updated the theme logic in JS to apply to close button and used that to determine whether the button should be light or dark.

## How to test

Run this fix in Drupal and test the video modal on this page. Use IE11 and make sure the close button is displaying the correct color. https://staging-www.pega.com/products/crm-applications


remydenton:
@mikemai2awesome, looks like you need to update the jest snapshots to get the tests to pass.

PR #1329, opened 2019-08-08
Bugfix/wwwd 1347 duplicate id attributes

remydenton:
## Jira

http://vjira2:8080/browse/BDS-1347

## Summary

Prevents duplicate ID on link element when passed in through attributes

## Details

- The test is somewhat elaborate-- please confirm that this makes sense.
- The fix simply removes an id attribute, if present, from being printed on the internal attributes. I have questions about this though:
  - Is this design pattern in bolt-link-- printing the same attributes on the custom element and within it-- the right way to do this in the first place?
  - The button component handles this problem very differently.  Specifically, it [removes id from custom element](https://github.com/bolt-design-system/bolt/blob/master/packages/components/bolt-button/src/button.twig#L158), but doesn't add it to the button internals.  The result is that you can't add an ID to a button element.  IMO, that's a bug as well.

### Reproduce the bug
- Create a link using the following twig:
```
  text: "This is a link",
  url: "https://pega.com",
  attributes: {
    id: "foo"
  }
} only %}
```
- Confirm that the `id="foo"` appears on both `<bolt-link>` element and the `<a>` element within it.

### Confirm the fix
- Check out this branch
- Using the same example, confirm that `id="foo"` only appears on the `<bolt-link>` element


packages/components/bolt-link/__tests__/link.js
```diff
@@ -220,4 +220,48 @@ describe('link', () => {
...
+      link.useShadow = false;
+      link.updated();
+      return link.outerHTML;
+    }, template.html);
+
+    const renderedHTML = await html('<div></div>');
+    renderedHTML.innerHTML = renderedLinkHTML;
+
+    // Loop though all elements to see if any two have the same ID.
+    const allElements = renderedHTML.getElementsByTagName('*');
```
danielamorse: @remydenton can you try `renderedHTML.querySelectorAll('#my-link')` here, and then check that the length is not more than `1`?

danielamorse: On second thought, your test is more through than mine, but if `querySelectorAll` works for you, you could still scope your search to `querySelectorAll('[id]')`.

remydenton: Great point.  I think I was under the impression that `querySelectorAll()` wouldn't keep looking for a second node after it found a first if the selector used an ID (`cause you're not supposed to have duplicates...).  But maybe that's just a jQuery thing-- I'll try it out.

danielamorse: I had the same thought and had to test it to find out. Worked for me 👍

remydenton: @danielamorse, if I check out code before the fix, set up a twig link with an ID (and add the `no-shadow` attribute in `link.twig`), open my browser console and use either `querySelectorAll('[id]')` or `querySelectorAll('#my-link')`, I get the elements I would expect based on your comments (i.e. one `<bolt-link>` and one `<a>`, each with the id "my-link")

However, if I do the following within my test, it gives different answers
```
const elemsWithAnyId = renderedHTML.querySelectorAll('[id]');
console.log(elemsWithAnyId.length);
// logs 2

const elemsWithMyId = renderedHTML.querySelectorAll('#my-link');
console.log(elemsWithAnyId.length);
// logs 1
```
(and no, it's not because there's some other element that has some _other_ id, though that is a concern with writing the test that way.  The second one isn't seeing the `<a>` element).

I'm totally stumped.  I suspect it has something to do with the `await html('<div></div>')` workaround, or something else about the process of going from twig to rendered HTML to js.

Thoughts?  I've already spent _way_ too much time trying to figure it out 😁 .

danielamorse: @remydenton, it looks like `querySelectorAll` is behaving differently in our test environment than it is in the browser. When I roll back to before the fix, I can log the `<bolt-link>` and `<a>` elements' attributes from within our Jest test and verify that both have the `my-link` ID. It's just that `querySelectorAll` is returning only the first.

Spoke with @sghoweri and learned that we have two different JS environments in our tests:
 - Inside `await page.evaluate(html => {})`, which uses Puppeteer and is full-featured.
 - Outside `await page.evaluate(html => {})` (but still inside `test()`), which uses a library called [basichtml](https://www.npmjs.com/package/basichtml) to implement HTML in Node and comes with [some caveats](https://www.npmjs.com/package/basichtml#current-caveats--exceptions).

So, it's not totally unexpected that `querySelectorAll` would not work perfectly. Upgrading to the latest version of `basichtml` might actually get us some new features. I commented the code so we can can revisit later.

I think your test is great for finding duplicates. I just scoped it to loop over elements that have any `[id]` vs all elements, since that query works reliably ([see commit](https://github.com/bolt-design-system/bolt/pull/1329/commits/992c6dccaac7dd9002eb7f2aee85e942b2fd14cf)).

remydenton: By the way, I created http://vjira2:8080/browse/BDS-1762 for the bug in button.

PR #1226, opened 2019-06-24
Add dependency mapper and quicker Jest tests

EvanLovely:
WIP for testing utilities to help with running subsets of tests. More info to be added when PR is ready for review.

## Jira

(Enter a link to the Jira issue this PR applies to, if any)

## Summary

(Enter a one-line summary of the changes in this PR.  This will be included in the release notes)

### Monorepo tests (`yarn test:monorepo`) 

Now will check all Twig files to see if they `include`, `embed`, or `extend` other Twig files and if so, ensure that their dependencies are correctly declared in their respective `package.json`. This also includes changes to get all of those correct and the test passing. This sets the foundation for other tools to accurately use dependency mapping.

### New command `yarn test:js:quick`

This is a beta command that will eventually replace `yarn test:js`. The difference is the new command looks at what's different between your PR and base branch (usually `master`), finds the component packages that changed, then finds all the recursive dependencies that are also Bolt packages, and then only runs tests in those packages. This is why the above test is important - we're sure that dependencies are correctly declared. 

#### How this works

The new `@bolt/testing-utils` package also has a CLI: `bolt-list-pkg-paths-changed` that can be ran from the root of the repo and will show the packages changed:

```bash
$ ./node_modules/.bin/bolt-list-pkg-paths-changed
Comparing this commit "HEAD" to base of "master":
These packages have changed, filtering tests to just these directories:
- @bolt/components-action-blocks : packages/components/bolt-action-blocks
- @bolt/components-device-viewer : packages/components/bolt-device-viewer
- @bolt/components-form : packages/components/bolt-form
- @bolt/components-icon : packages/components/bolt-icon
- @bolt/components-icons : packages/components/bolt-icons
- @bolt/components-smooth-scroll : packages/components/bolt-smooth-scroll
- @bolt/testing-helpers : packages/testing/testing-helpers
- @bolt/testing-utils : packages/testing/testing-utils
- @bolt/twig-renderer : packages/twig-renderer
- @bolt/scripts : scripts
packages/components/bolt-action-blocks|packages/components/bolt-device-viewer|packages/components/bolt-form|packages/components/bolt-icon|packages/components/bolt-icons|packages/components/bolt-smooth-scroll|packages/testing/testing-helpers|packages/testing/testing-utils|packages/twig-renderer|scripts
```

The last line is sent to STDOUT and all the other lines are sent to STDERR so when this command is ran:

```bash
jest "$(bolt-list-pkg-paths-changed)"
```

[...]

PR #1334, opened 2019-08-09
ci: setup 2 different unit tests jobs if on master

EvanLovely:
## Jira

BDS-1607

## Summary

Enables `yarn test:js:quick` (just tests for packages that have changed) to run in CI when not on `master` and sets up `yarn test:js` to run on `master`.

## Details

See #1226 

## How to test

The branch `feature/example-change` has this code and also a simple change in the button; see [it's Travis build](https://travis-ci.com/bolt-design-system/bolt/builds/122663328) and note that only the button test runs.


colbytcook:
Closing because PR is outdated

PR #1338, opened 2019-08-15
List | FE | Spacing consistency and code clean up

mikemai2awesome:
## Jira

http://vjira2:8080/browse/BDS-1610
http://vjira2:8080/browse/BDS-1671

## Summary

Fixes the incorrect spacing for items in an inline or flex list with separators.

## Details

1. Rewritten some of the CSS to make sure the spacing logics are correct.
2. Said logics are now commented and explained.
3. The spacing in a flex list with separators is now consistent with block and inline lists.
4. The spacing in the last item of an inline list with separators is removed.
5. Used the `bolt-repeat-rule` mixin wherever applicable, this fixes a few IE11 and no-js issues.

## How to test

Pull down the branch locally and perform the following.

### Check tests

- [ ] 1. Given I have pulled down the branch locally
- [ ] 2. And I am in the root folder
- [ ] 3. When I run the command `npx jest packages/components/bolt-list/__tests__/list.js`
- [ ] 4. Then I should see all tests indicated as "Pass"

### Check inline list's last item spacing

- [ ] 1. Given I am viewing the separator variations page
- [ ] 2. When I am viewing each of the inline list sections
- [ ] 3. Then I should see the list's last item does not have margin-right or padding-right

### Check flex list's spacing

- [ ] 1. Given I am viewing the separator variations page
- [ ] 2. When I am comparing the flex list sections and the inline list sections
- [ ] 3. Then I should see the same spacing in between items, the only difference would be flex list items' widths are evenly distributed while inline list items' widths fit to their content

### Check flex lists in IE11

- [ ] 1. Given I am viewing the display variations page
- [ ] 2. And I am using IE11
- [ ] 3. When I am viewing the Flex section
- [ ] 4. Then I should see a horizontal list of items evenly distributed across the width of the browser.

### Check testing instructions

- [ ] 1. Given I am in the `bolt-list` component folder
- [ ] 2. And I open the `TESTING.md` file
- [ ] 3. When I go through the instructions text
- [ ] 4. Then the text should be well formatted with clear information architecture
- [ ] 5. And the instructions are intuitive and easy to follow without any typos


packages/components/bolt-list/src/list.scss
```diff
@@ -69,21 +96,42 @@
...
+  overflow: hidden; // [1]
 }
 
-[class*='c-bolt-list--separator'] {
-  &.c-bolt-list--display-flex {
-    overflow: hidden;
+@each $spacing-value in $bolt-spacing-values {
+  $spacing-value-name: nth($spacing-value, 1);
+
+  @include bolt-repeat-rule((
```
sghoweri: Code smell. We should dig into this a bit more.

mikemai2awesome:
## The Flex List with Separator Problem

We've had this problem since the component was released, and it wasn't discovered until I try to use the component extensively. Here's screenshot of the problem shown on `master`

### Master

master

This breaks whenever the items do not have the exact same width and height. And the biggest problem is using `overflow: hidden`, this is a big no-no for any layout container, because it can cut off things that bleed, such as tooltips, dropdown, button shadows, card shadows, etc.

### Expected Outcome

expected

The expected outcome is to have it work similar to the action blocks component where the separators would all connect to create a grid.

### Attempted Fix

fix

Despite my best efforts and also ignoring the `overflow: hidden` problem, I still cannot get the separators to work as intended. And this is resulting in A LOT of messy code.

## Compromise

This component is supposed to handle simple layouts and alignments, and I am very against using `overflow: hidden`. So I've come to this conclusion: keep the flex list simple. When changing the list's display to `flex`, it will only do exactly that. It will not do anything else that seems like magic. As displayed below, the ONLY difference between an inline list and a flex width would be that the flex list items would try to fill up the available space of its container.

This component is not widely used or implemented in Drupal, so it's safe to do this now.

final1
final2

## Feedback

What do you think of the compromise? Is there a solution to this without using `overflow: hidden` that I am not aware of? Discuss. I am refreshing all the reviews for now.

@adamszalapski @danielamorse @sghoweri

mikemai2awesome:
@danielamorse No use case right now, it was originally intended to replicate the action blocks behavior. Also, I updated the text.

sghoweri:

sghoweri:
PR was released with v2.9.0

PR #1342, opened 2019-08-21
Feature/svg animation multi

WessWillis:
## Jira

https://app.asana.com/0/1134135746083363/1132926339924692

## Summary

New SVG Animations component with Radar, Orbit, and Automation.

## Details

This the initial version of  bolt-svg-animations,  which allows a user to choose between multiple animated SVGs and control parameters such as speed or theme. 

The individual SVGs are first created in SVGator.  I've refined and trimmed down the generated CSS quite a bit, could possibly do more if necessary. 

Radar uses a .png background, provided by Yao, as the gradient in the comp is not possible with raw SVG alone.   Radar also supports a light/dark theme prop, while the others do not.   

## How to test

Visit   /pattern-lab/?p=viewall-components-svg-animations


glassdimly:
Hi @WessWillis,

This PR is not something we can review with all these with/without changes and all of Salem's work in it.

I know I told you to merge this stuff in for the self-drawing circle stuff when it was targeting with-without, but now can you check out `epic/interactive-pathways`, roll your changes, and only your changes, against that? I don't mind if it's a single commit caused by copying your files over.

The general principle here is that when a branch is targeting another branch, it should be cleanly merge-able and the diff should reflect only the changes necessary for that branch's functionality.

Thanks,
Jeremy

glassdimly:
I just noticed that the reason for all these changes is that @EvanLovely changed the merge target of this branch! Anyhow, can you rebase as I suggested above, @WessWillis ?

PR #1355, opened 2019-08-23
fix(build): error on yarn build:noisy; fix integer cast for verbosity…

glassdimly:
`yarn build:noisy` sets the verbosity at 5 in the PL build. I was needing this to test a specific image that was causing the build to fail. But what I found is that `yarn build:noisy` was broken.

Error:
```
$ yarn build:noisy
yarn run v1.16.0
$ bolt build --verbosity 5


Starting build (serial)
Config Schema Errors:
1) instance.verbosity is not of a type(s) integer
2) instance.verbosity is not one of enum values: 0,1,2,3,4,5
Error: Please fix the config being used in Bolt CLI.
```

The problem is that `parseInt` must be called with a radix.

## How to test

run `yarn build:noisy` from `docs-site/`


sghoweri:

sghoweri:
PR was released with v2.9.0

PR #1391, opened 2019-09-05
Fix/horizontal scroll from character

glassdimly:
## Jira

https://app.asana.com/0/1126340469288208/1137915602201453

## Summary

Fix content overflow; responsiveness for bolt-connection and bolt-character

## How to test

Check out micro-journey in band. Check different widths. Make sure to refresh after scaling viewport upward.


packages/micro-journeys/src/character.scss
```diff
@@ -29,10 +29,17 @@
...
     align-items: center;
+    max-width: 75px;
+    max-height: 75px;
     overflow: hidden;
     border-radius: 50%;
     background-color: white;
     @include bolt-shadow('level-20');
+
+    @include bolt-mq(xsmall) {
+      max-width: unset;
```
joekarasek: Pretty sure 'unset' does not work in IE11. Lets cross that bridge on the IE11 styling cleanup though

glassdimly:
@joekarasek Can you please give me steps to reproduce this? I created this PR to address that very problem.

I'm in Chrome and FF looking at Micro Journey in Band, testing at all widths from 320px all the way up.

True I'm pushing this out the door quickly and may have missed something, but please do give me a bit more feedback here.

PR #1404, opened 2019-09-10
feat(bolt-animate): add debug optional debug flag

glassdimly:
## Jira

https://app.asana.com/0/1126340469288208/1139247289936295/f

## Summary

Add debug code to bolt-animate

## Details

In with/without, I needed to see what animations were failing and why.

Having permanent debug code in place made it much easier to see.

I think we should make this permanent because animation sequencing is often tricky to debug.

## How to test
```
  await triggerAnims({
    animEls,
    stage: 'IN',
    debug: true, // <- here is the flag you can now use
  });
```

### DIFF
--- a/packages/components/bolt-animate/utils.js
+++ b/packages/components/bolt-animate/utils.js
@@ -77,7 +77,7 @@ export async function triggerAnims({ animEls, stage = 'IN', debug = false }) {
   let eventName;
   switch (stage) {
     case 'IN':
-      eventName = 'bolt-animate:end:out';
+      eventName = 'bolt-animate:end:in';
       orderProp = 'inOrder';
       hasAnimProp = 'hasAnimIn';
       break;
--- a/packages/components/bolt-animate/utils.js
+++ b/packages/components/bolt-animate/utils.js
@@ -2,9 +2,11 @@
  * @param {Object} opt
  * @param {BoltAnimate[]} opt.animEls
  * @param {string} [opt.stage='IN']
+ * @param {boolean} [opt.debug=false]
+ *
  * @return {Promise<{ success: boolean, animEl: BoltAnimate }[]>}
  */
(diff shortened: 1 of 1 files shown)


packages/components/bolt-animate/utils.js
```diff
@@ -89,11 +98,15 @@ export async function triggerAnims({ animEls, stage = 'IN' }) {
...
+    if (debug) {
+      console.debug(`${eventName}: order:${order}`, animElsToTrigger);
+    }
     // eslint-disable-next-line no-await-in-loop
     await triggerAnimOnEls({
-      animEls: animEls
-        .filter(a => a[hasAnimProp])
-        .filter(a => a[orderProp] === order),
+      animEls: animElsToTrigger,
       stage,
```
EvanLovely: Wouldn't we want to pass `debug` that we got from `triggerAnims` into `triggerAnimOnEls` too? Or is that intentional as one would only want one or the other since having both debugs would be noisy? Also, I love that you used `console.debug` instead of `console.log` :+1:

glassdimly: Yep, I was just thinking that too as I was reading through my PR

packages/components/bolt-animate/utils.js
```diff
@@ -64,16 +69,20 @@ async function triggerAnimOnEls({ animEls, stage }) {
...
+ * @param {boolean=} [opt.debug=false]
  */
-export async function triggerAnims({ animEls, stage = 'IN' }) {
+export async function triggerAnims({ animEls, stage = 'IN', debug = false }) {
   let orderProp;
   let hasAnimProp;
+  let eventName;
   switch (stage) {
     case 'IN':
+      eventName = 'bolt-animate:end:out';
```
EvanLovely: wrong event name; should be `in` not `out`

PR #1415, opened 2019-09-12
chore(button): add on_click and on_click_target to button schema

glassdimly:
## Summary

Adds `on_click` and `on_click_target` to Button schema.

## Details

While integrating the Button component into the Editor, we realized that Button extend BoltAction, which has `on_click` and `on_click_target` attributes. We need these attributes in the Editor, but we pull from Button schema, and `bolt-action` has no schema (so it wouldn't be possible to merge them).

Therefore, adding this to `bolt-button` the same way it's added to `bolt-trigger`'s schema.

Would be great to get this included in the upcoming release.

### DIFF
--- a/packages/components/bolt-button/button.schema.yml
+++ b/packages/components/bolt-button/button.schema.yml
@@ -69,6 +69,12 @@ properties:
     enum:
       - regular
       - full
+  on_click_target:
+    description: '`className` (e.g. "js-click-me") used in `querySelector` to reference a web component on the page. `onClick`, the `on_click` method name will be called on this element.'
+    type: string
+  on_click:
+    description: 'The name of a method on the `on_click_target`.'
+    type: string
   align:
     description: 'Horizontal alignment of items (text and icon) inside the button. Note: the values left and right are deprecated, use start and end instead.'
     type: string


glassdimly:
@sghoweri Yes, I thought about that and it was my first choice. 

However, I saw that this workaround was already implemented in `bolt-trigger` so I had assumed this architectural decision had already been made this way for a reason.

Want me to move `bolt-trigger` and these others into `bolt-action`?

Let me know how to proceed.

glassdimly:
In general I'm not trying to be presumptuous in re-factoring!

glassdimly:
Hi @sghoweri,

Sorry for the back-and-forth here!

Just so we're on the same page, `bolt-trigger`, `bolt-button` and `bolt-link`extend `boltAction`. `bolt-button` and `bolt-link` do not extend `boltTrigger`.

I looked at `bolt-action` and the reason why I didn't add `on_click` and `on_click_target` to the schema is because `bolt-action` doesn't have a schema, and it doesn't have a `package.json`, because it is an `element` not a `component`.

I could probably add a schema that was just a `yml` file not referenced from a `package.json` and infer documentation of the props from `boltTrigger` but I didn't see that pattern in any other `element`... cool, LMK and I will do it.

One option here is actually extending `boltTrigger` from `bolt-button` and `bolt-link`. But I'm not familiar enough with these components to actually advocate for this.

I'm happy to take any of these options, just let me know how to proceed.

glassdimly:
So the real fix here would be to output a merged schema that looks up to parents and merges in those schemas. However, in order to do this, we'd need two schema files per component. For instance we could have:
`animate.schema.yml` and `animate.merged.schema.yml`, with `animate.merged.schema.yml`, being the one listed in `package.json` and looking up to class parents to find schemata.

glassdimly:
This is old and has been dealt with elsewhere. Closing.

PR #1424, opened 2019-09-16
fix(micro-journeys): remove transform from connection band to fix ie 11

glassdimly:
## Jira

https://app.asana.com/0/1126340469288208/1139858504448256/f

## Summary

Remove transform from Connection Band to fix ie 11

## Details
The connection band in IE uses a transform:
```
#connectionGradientBG {
    transform: translate(-28px, 0px);
}
```

This causes it to break in IE 11. Here's a stack overflow issue on this subject: https://stackoverflow.com/questions/23047098/css-translate-not-working-in-ie11-on-svg-g/28726517

I just used the SVG coordinates instead of translate to put the band in the right place in the first place, translate was not needed.

Before:
IMG_8573

Make sure to clear cache!
IMG_8574

After:
IMG_8575

## How to test

Visit Micro Journeys in IE 11, confirm fix shown in screenshots


joekarasek:
Looks like the animated svgs are now static and present in IE11. No changes in other browsers.

PR #1450, opened 2019-09-24
Fix/text links

WessWillis:
## Jira

https://app.asana.com/0/1126340469288208/1140960381571000/f

## Summary

There was bug filed for a link having incorrect styling.   The root cause was the `<bolt-link>` not being used, so the copy in question was not a link.  This updates the html file to properly create the link, albeit linking to "#" as these are merely examples.  

## How to test
Note the text styling of  "It's not mind reading, its signals" on this page: 
 
`/pattern-lab/?p=experiments-micro-journey-in-themes`


WessWillis:
@glassdimly  apologies, I'd set the base to  _master_ rather than _develop_.   This has been updated and the PR is correct now.

PR #1457, opened 2019-09-26
Core | JS | Fix tabs, accordion, carousel nesting bugs

danielamorse:
## Jira

- http://vjira2:8080/browse/BDS-1856
- http://vjira2:8080/browse/BDS-1861

## Summary

Fix rendering bugs that happen when Tabs, Accordion, or Carousel components are nested inside one another.

## Details
- Reworked ssr-hydration to handle nested components
- Dispatch custom event when layout changes
- Add `update()` method to video to fix control bar bug when inside tabs or accordion
- Fix hidden pagination dots in IE11

## How to test
- Run locally and test in Chrome and IE:
  - Accordion demos, especially "Content" page
  - Tabs demos, especially "Content" page
  - Carousel demos


docs-site/src/pages/pattern-lab/_patterns/02-components/accordion/999-accordion-web-component.twig
```diff
@@ -12,15 +12,15 @@
 {% set basic_usage %}
 <bolt-accordion>
   <bolt-accordion-item>
-    <bolt-text slot="trigger">Accordion item 1</bolt-text>
```
sghoweri: @danielamorse could you see if we can patch the slotted CSS selector for these to remove the margin-bottom styles you mentioned so we know `<bolt-text>` still works with trigger the way that end users would expect?

danielamorse: @sghoweri reverted this and actually no longer see the original issue (extra space). If it reappears I'll fix via CSS.

packages/core/elements/ssr-keep/README.md
```diff
@@ -0,0 +1,33 @@
+## `<ssr-keep>`
```
sghoweri: +1 on the docs for this!! 💯

packages/components/bolt-accordion/__tests__/__snapshots__/accordion.js.snap
```diff
@@ -1,615 +1,564 @@
...
+    >
+      Lorem ipsum dolor sit, amet consectetur adipisicing elit.
+    </bolt-text>
     
-  <!---->
-  <!---->
-</div>
-`;
-
-exports[`<bolt-accordion> Component Default <bolt-accordion> with Shadow DOM renders 1`] = `
```
sghoweri: @danielamorse why did these specific Jest tests / snapshots get nuked? Didn't I just add these?

packages/components/bolt-accordion/__tests__/accordion.js
```diff
@@ -144,53 +137,21 @@ describe('<bolt-accordion> Component', () => {
   });
 
   test('Default <bolt-accordion> with Shadow DOM renders', async function() {
```
sghoweri: @danielamorse compare the Snapshot diff from Jest with the updates I just put in vs after — notice how a ton of the rendered markup isn't quite all there + you don't actually get snapshot HTML of the inner rendered structure. Was this intended?

^ I also added the more reliable (and newer) approach to making sure the UI renders before grabbing the markup instead of using the original `.updated()` approach which I've found can periodically throw errors if a component isn't yet ready when this method tries to get run.

packages/components/bolt-tabs/src/tabs-dropdown.scss
```diff
@@ -40,11 +40,6 @@ $bolt-tabs-xsmall-bp: 420px; // Slightly increase the default 400px xsmall break
 
   @include bolt-mq($until: $bolt-tabs-xsmall-bp) {
     margin-left: auto;
-    @include bolt-margin-right(xsmall);
```
sghoweri: @danielamorse why was this removed?

danielamorse:
@sghoweri per our discussion on testing I've updated the Accordion tests to include the Shadow DOM HTML that was originally output by your test. [See commit](https://github.com/boltdesignsystem/bolt/pull/1457/commits/32d3a7de19ba1fefd9a4c7b431f123c2c571f96d#diff-b5a67f98e74bf289f24c57642eecc23d).

I'm keen on further automating the basic three tests most of our components have (twig, shadow, noShadow), but I'm going to tackle that after 2.9.

Also, definitely interested in extending `ssr-keep` to include the features you outline above after this release.

P.S. Look at that build go! What speed!

sghoweri:
PR was released with v2.9.0

PR #1471, opened 2019-10-03
Hotfix/with without fixes

glassdimly:
## Jira

https://app.asana.com/0/1132619174015272/1140685646931269/f
https://app.asana.com/0/1132619174015272/1142071935279842/f
https://app.asana.com/0/1132619174015272/1142061852792839/f

## Summary

Fix self-drawing circle rotation in edge; fix mobile padding overlap; add IntersectionObserver; code cleanup

## Details

This branch contains a number of fixes, many of which I implemented  as I discovered them when working to fix IE 11.
* Add `InteresectionObserver` to animation on when  initially scrolled into view
* Fix mobile padding overlap
* Force full bleed as per stakeholder discussions
* Circle CSS cleanup and refactor
* Fix self-drawing circle erratic rotation in Edge
* Short-circuit if no toggle found to fix errors on non-with/without PL pages
* Fixing fallout from forcing full bleed, correcting general spacing issues especially on tablet
* Some fixes for the lowest blocks to prevent overflowing the container on tablet

This does not include IE 11 fixes, but Edge is solid, minus the selection shown around the toggler, which is already tracked.

## How to test

Examine /pattern-lab/?p=experiments-with-without-intersection-test and /pattern-lab/?p=viewall-experiments-with-without


packages/with-without/wwo-self-drawing-circle/self-drawing-circle.js
```diff
@@ -56,17 +56,20 @@ class SelfDrawingCircle extends withLitHtml() {
...
   render() {
     return html`
       ${this.addStyles([styles])}
-      <div id="wwo-self-drawing-circle" class="wwo-self-drawing-circle">
-        <div class="wwo-self-drawing-circle-rotation">
+      <div
+        id="wwo-self-drawing-circle"
+        class="wwo-self-drawing-circle wwo-self-drawing-circle__rotater"
+      >
+        <div class="wwo-self-drawing-circle__spinner">
```
glassdimly: In Edge, if a transformed element is animated, it becomes erratic. So we move our transforms outside the spin animation.

packages/with-without/js/handleResize.js
```diff
@@ -3,36 +3,41 @@ import { triggerAnimateInOnInOnlyContent } from './handleActiveRegionChange';
...
-
-  // @TODO replace with theme token.
-  const isMobile = window.matchMedia('(max-width: 800px)').matches;
-  const isDesktop = window.matchMedia('(min-width: 1201px)').matches;
-  const isTablet = !isDesktop && !isMobile;
-
-  const container = document.querySelector('.c-pega-wwo__wrapper');
-
-  if (container) {
+const handleResize = (shouldForce = false, shouldTriggerAnims = true) => {
```
glassdimly: Overloaded to be called to set height after load without triggering any animations.

packages/with-without/js/with-without.js
```diff
@@ -6,86 +6,124 @@ import {
...
-}
+};
 
-// Wire up the toggler to the event region switcher.
-Array.from(document.querySelectorAll(toggleInputClass)).forEach(el => {
-  el.addEventListener('change', e => {
-    handleActiveRegionChangeRequest(e.target.id === 'w', wwoSwiper);
-  });
-});
+(() => {
```
glassdimly: This diff is so gnarly because I made it an IIFE so I could use a guard clause to check for the presence of the toggler ID to short-circuit load. 🤷‍♂

charginghawk:
@glassdimly I'm pulling this into Drupal and testing it there. Off the bat, It looks like the "don't tread on me" is still getting covered below 1250px:

https://hotfix-with-without-fixes.boltdesignsystem.com/pattern-lab/?p=experiments-with-without

glassdimly:
Alright, I believe I've addressed all of these issues, or we've agreed they're not necessary for this release:

1. Self drawing circle now rotates in the proper direction and resets and stops its spin before re-animation
2. We talked about the full-bleed content on the call: stakeholders requested full bleed early and we planned for this. If for some reason this changes, I have moved the full-bleed behavior into its own class, so it can be easily removed. I cannot guarantee there will be no fallout from this, but I added a couple fixes so that this should work fine.
3. **1239/1251**: added padding right to with pega results block
image
**1231** is now fixed
image

4. **1467** overlap is now fixed.
image

5 and 6. We agreed that supporting scaling content up and down without page refresh is not launch critical

7. I have not been able to replicate this between-breakpoints problem on Chrome/FF. I tried starting big, moving down to **1230**. I tried starting small, moving up to **1230**. I tried refreshing at **1230**. Perhaps this is fixed? Can you check once build finishes on this branch?

Misc.: Yeah, I had trouble with adding that icon to icons, the size wasn't supported. I was actively avoiding bolt components as per my instructions to avoid dependencies on bolt. I will do it differently next time!

Other: 
* Fixed excessive black space at the bottom of tablet container problem
* Fixed height on mobile/tablet after accordion expand
* Edge fixes
* Hide with/without entirely on IE 11 as per @charginghawk's discussion with stakeholders

margoromo:
@sghoweri @glassdimly This hotfix was checked and approved by stakeholders.

PR #1475, opened 2019-10-04
fix: more robust animation resets

joekarasek:
fix: more robust animation resets to allow retrigger when no idle or build out is defined

## Asana

https://app.asana.com/0/1126340469288208/1140132371899568/f

## Summary

More robust reset of animations when idle or build out are not defined.

## Details

There was an issue with the radar svg on the first step stage of the demo not retriggering the second time the step became active. We discovered that bolt-animate components that do not have a defined idle or build-out animations were not properly resetting to their initial state. 

## How to test

http://localhost:3000/pattern-lab/patterns/06-experiments-micro-journeys-combos--92-micro-journey-in-themes/06-experiments-micro-journeys-combos--92-micro-journey-in-themes.html

1) Go to the above URL
2) On the first step, the radar svg behind the character on the left should fade in then out
3) Go to another step, then return to the first step. The radar svg should once again fade in then out.


packages/components/bolt-animate/utils.js
```diff
@@ -63,11 +61,12 @@ async function triggerAnimOnEls({ animEls, stage, debug = false }) {
                 `animation taking too long for stage "${stage}", cancelling and moving on to next one.`,
                 animEl,
               );
+              animEl.removeEventListener(eventName, handleEvent);
               resolve({
                 success: false,
                 animEl,
               });
-            }, duration + 100);
+            }, duration + 250);
```
glassdimly: This is great, thanks for changing this.

packages/components/bolt-animate/utils.js
```diff
@@ -129,9 +126,7 @@ export async function triggerAnims({ animEls, stage = 'IN', debug = false }) {
   const animOrders = [...orders].sort((a, b) => a - b);
 
   for (const order of animOrders) {
-    const animElsToTrigger = animEls
-      .filter(a => a[hasAnimProp])
-      .filter(a => a[orderProp] === order);
+    const animElsToTrigger = animEls.filter(a => a[orderProp] === order);
```
glassdimly: Let's say I passed in an array of els for `OUT` which included an el that didn't have an `OUT` animation. Wouldn't this now attempt to trigger a non-existent `OUT` animation on that el now? More broadly, I don't fully understand why we're removing this code. Happy to jump on the phone if that's easier.

packages/components/bolt-animate/src/animate.js
```diff
@@ -166,6 +160,23 @@ class BoltAnimate extends withLitHtml() {
...
+      }
+    } else {
+      if (this.hasAnimIdle) {
+        this._triggerAnimIdle();
+      }
+    }
+
+    this.triggerUpdate();
+  }
+
```
glassdimly: Yes. This is the fix! Great.

PR #1476, opened 2019-10-07
feat(with-without): toggle outline removal on hover-capable devices

glassdimly:
## Jira

https://app.asana.com/0/1132619174015272/1142061852792836/f
and 
http://vjira2:8080/browse/WWWD-4121

## Summary

with/without toggle outline removal on hover-capable devices

## Details
I fixed this in an ugly way. I spent about an hour or so researching and trying various combinations of things, but I could not find a way to affect the outline on checkboxes that was afflicting our toggle. The internet's ideas don't work, nor did mine. So I simply added a function to the toggle event, checked for Edge in the `UserAgent` also check for ` @media (any-hover: hover)`, and removed focus if both are found.

This is certainly a hack, but it works and its what I can offer on this ticket within our contract. If someone who is an Edge expert wants to take a crack: go for it.

In general I don't think this is too bad. I'm going to guess that most people tabbing around browsers for focus will be using a modern browser rather than Edge, and it's only being remove in case where hovering is possible, i.e. there is a pointing device. Finally, a lost focus isn't a dealbreaker.

## How to test

Replicate bug on Edge in develop with/without, verify removal of bug on this branch.


margoromo:
@charginghawk @sghoweri @glassdimly This hotfix was reviewed and approved by stakeholders.

charginghawk:
Looks good to me, merging into the main wwo hotfix branch.

charginghawk:
@glassdimly Please resolve the merge conflict.

PR #1486, opened 2019-10-10
(WIP/Debug) Debug/with without chrome fail

glassdimly:
## Summary

With/Without fails on Chrome in some contexts Mac and Windows, but specifically reproducible on Chrome 77 Windows 8.1 in Browserstack.

## Details

This is not a difference between triggerAnims and triggerAnimsOut. See link: https://develop.boltdesignsystem.com/pattern-lab/?p=experiments-with-without-overlay-animation-test

It is not a DomReady problem because IntersectionObserver also doesn't work: https://develop.boltdesignsystem.com/pattern-lab/?p=experiments-with-without-intersection-test

It _is_ a shadow dom problem, because w/wo works without the shadow dom enabled: https://develop.boltdesignsystem.com/pattern-lab/?p=experiments-with-without

Micro Journeys work fine.

## How to test

Load with/without on Chrome in Windows 8.1 in Browserstack.

### DIFF
--- a/packages/with-without/index.js
+++ b/packages/with-without/index.js
@@ -1,6 +1,7 @@
 import { polyfillLoader } from '@bolt/core/polyfills';
 
 polyfillLoader.then(res => {
+  console.error('Polyfills just finished loading, kicking off w/wo.');
   import(/*
     webpackMode: 'eager',
   */ './js/accordion');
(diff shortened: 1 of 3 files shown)


glassdimly:
Fixed by @sghoweri modifying `IntersectionObserver`.

PR #1491, opened 2019-10-10
Fix/ie edge svg animations

WessWillis:
## Jira

https://app.asana.com/0/1126340469288208/1143553749250524/f
https://app.asana.com/0/1126340469288208/1143212411844535/f

## Summary

IE/Edge fixes for microjourneys

## Details

Fixed: 

- Radar has a strange background appearing in various browsers/states
- Interactive Pathway nav active link has wrong color in dark theme 
- Animations have wrong color in light theme IE11  (css collision)
- Automation has poor start position in IE11 (these are unanimated, static version needs to look good)
- Triple Band connection gradient is off in edge/ie11  (don't use gradientTransform in an SVG)
- Removed unused svg-animations twig file to avoid future confusion 

## How to test
Visit  svg-animations and micro-journeys in pattern lab in IE and Edge


packages/build-tools/tasks/webpack-tasks.js
```diff
@@ -89,6 +89,7 @@ async function server(customWebpackConfig) {
           ui: false,
           notify: false,
           open: false,
+          tunnel: false, //TODO leaving for easy toggling to test IE locally.  remove before merge to master.
```
glassdimly: @WessWillis I'm going to remove this, but in the future please do this cleanup before submitting a PR for review.

glassdimly: Also, there's a ton of linter errors, which was blocking the build from completing and therefore blocking merge.

Ways to fix this:
* Do not use `--no-verify` when pushing.
* `yarn lint:scss --fix && yarn lint:js --fix`

image

WessWillis: I actually didn't use `--no-verify`.    I did runn `lint:scss/js --fix` at the end but apparently forgot to commit.  I'm wondering why the tests didn't catch all that on push?    

re: `tunnel: false`, that doesn't do anything.   While we're still working on develop it seemed acceptable to leave the toggle there so its simpler to switch to `true` when doing IE testing. Otherwise on each  branch i have to dig around and find that file/line again.    That said, is the rule that we don't push anything to develop that we wouldn't push to master?

sghoweri:

sghoweri:
PR was released with v2.9.0

PR #1501, opened 2019-10-17
Tabs | Bug | Fix tabs inside band IE/Edge bug

danielamorse:
## Jira

- http://vjira2:8080/browse/WWWD-4173

## Summary
Fix IE/Edge bug that breaks dropdown menu when Tabs are nested inside a Band.

## Details
In IE/Edge Tabs are disconnected and reconnected when placed inside a band. After disconnecting, certain event handlers were not getting re-added and breaking the dropdown menu.

To fix, I made sure the `ready` flag was unset on disconnect so that the component will properly setup the next time it connects.

## How to test

- View Tab Content page in IE and Edge. There is a new example of a Tab inside a Band. Make sure the dropdown works properly. Resize the window and test.


sghoweri:
PR went out with version: v2.8.3

PR #1509, opened 2019-10-22
Image | JS | Fix images in tabs appearing fuzzy

danielamorse:
## Jira

- http://vjira2:8080/browse/WWWD-4245

## Summary

Fix images in tabs appearing fuzzy.

## Details

Images that are in an inactive tab were not getting properly lazyloaded because they had no width.
We must check that an image has width before setting `sizes`.

## How to test

Run locally and view the Tabs Content page in an incognito window. Click on the "Image" tab and verify the image is not fuzzy.


sghoweri:
PR was released with v2.9.1

PR #1510, opened 2019-10-23
BDS-1761 inline form demos

remydenton:
## Jira

http://vjira2:8080/browse/BDS-1761

## Summary
~The goal here is to create an inline form, ideally without requiring code changes.  I've created four different possible ways of handling it in the following demo and am looking for feedback on which is best.  Note that the second and third examples are the only ones that don't shift vertically when the error message is displayed.~

Adds demos of how to create inline forms using existing components

https://feature-bds-1761-inline-form-demos.boltdesignsystem.com/pattern-lab/?p=components-form-inline-inputs


sghoweri:
One thing I’d recommend thinking about here is to keep in mind that whatever approach works best here should really be  _internally_ tucked away. 

Component encapsulation (even if it means duplicating styles) > internal component reuse in this particular use case!

For example:
```
<Form>
  <FormRow>
    <Input />
    <Submit />
  </FormRow>
</Form>
```

remydenton:
@sghoweri, I'm not sure I agree with (or maybe just don't understand) the benefits of a separate web component for form layout.  Aren't we currently trying to work towards fewer different layout components?

docs-site/src/pages/pattern-lab/_patterns/02-components/form/65-form-full-inline.twig
```diff
@@ -0,0 +1,55 @@
...
+
+  {% set full_width_submit_button %}
+    {% include "@bolt-components-form/form-button.twig" with {
+      text: "Sign up now",
+      tag: "submit",
+      width: "full",
+    } only %}
+  {% endset %}
+
+  <h3>Inline form: standard width</h3>
```
mikemai2awesome: I would name this "Inline inputs: auto width". And add a paragraph:

Use a `bolt-list` with `display` set to `inline`, and pass each input as an item. This will create an inline layout with multiple inputs. The width of each input would be adjusted to fit its content.

docs-site/src/pages/pattern-lab/_patterns/02-components/form/65-form-full-inline.twig
```diff
@@ -0,0 +1,55 @@
...
+    spacing: "xsmall",
+    align: "start",
+    valign: "start",
+    items: [
+      form_field,
+      submit_button
+    ]
+  } only %}
+
+  <h3>Inline form: full width</h3>
```
mikemai2awesome: I would name this "Inline inputs: flexible width". And add a paragraph:

Use a `bolt-list` with `display` set to `flex`, and pass each input as an item. This will create a flexible layout with multiple inputs. The width of each input would be adjusted to the available space to fill up the full width of the layout.

docs-site/src/pages/pattern-lab/_patterns/02-components/form/65-form-full-inline.twig
```diff
@@ -0,0 +1,55 @@
+{% extends "@bolt/form.twig" %}
+
```
mikemai2awesome: I would name this file form-inline-inputs.twig instead.

remydenton:
I agree with all those changes, thanks @mikemai2awesome.

sghoweri:
> @sghoweri, I'm not sure I agree with (or maybe just don't understand) the benefits of a separate web component for form layout. Aren't we currently trying to work towards fewer different layout components?

@remydenton did you want to dive into this a bit more during our next Dev huddle?

At a high level it comes down to a couple things:
1. Single responsibility principle / separation of concerns (having some things do one thing and _only_ one thing really well)
2. Differentiating general purpose layout tools from focused, use case-specific layouts / layout tools
3. Reducing our cyclomatic complexity (ex. modifying the grid system shouldn’t require knowing about or updating the FormGroup component)
4. Simplicity from an end-user POV. Ex. Form UI that needs to get grouped together do so using the FormGroup parent wrapper and FormGroup Item around each item. That’s it. No ambiguity, no complexity reusing this pattern — just use FormGroup and that’s it.

^ what FormGroup does under the hood is totally up for discussion (reusing low level layout tools vs custom component-specific styles) however the implementation-specific details aren’t something a develop or content author should know or care about.

Too much flexibility = ambiguous, hard to replicate, hard to support, inconsistent UI


Also FWIW, I very recently had some hands on experience dealing with this exact use case in React which further reinforced a lot of my gut feelings around these types of problems being solved.

CC @danielamorse @mikemai2awesome @margoromo @adamszalapski

remydenton:
Thanks for elaborating @sghoweri!  I'd love to discuss further in a dev huddle.

sghoweri:
PR was released with v2.11.0

PR #1513, opened 2019-10-24
Accordion | JS | Fix nested accordion bug

danielamorse:
## Jira

- http://vjira2:8080/browse/BDS-1870

## Summary

Fix bug where an accordion inside an accordion does not render properly.

## Details

It appears that changing when `ssrHydrationPrep` gets called resolves this issue on its own ([see commit](https://github.com/boltdesignsystem/bolt/commit/b7f14efa74e778f888e9c122882929869fc03105#diff-d45fd62db4da766d397ba2b3eaae5990)). Still, the changes here make the component more resilient. Before, accordion items were initialized on 'rendered', which could be unpredictable as components may render multiple times, and on certain renders we got different results when we queried the 'renderRoot`. Now, we're initializing on 'ready', inside a `Promise`, which waits for all accordion items to finish before setting up the Accordion.

## How to test

- Check out `fix/nested-accordion-bug` branch locally.
- Go to the [Accordion content page](http://localhost:3000/pattern-lab/patterns/02-components-accordion-40-accordion-content-variations/02-components-accordion-40-accordion-content-variations.html
) locally.
- Verify the nested accordion in the second example properly renders and functions as expected.
- For good measure, also check the Accordion tab on the [Tabs Content page](http://localhost:3000/pattern-lab/patterns/02-components-tabs-30-tabs-content/02-components-tabs-30-tabs-content.html) locally.
- Test in Chrome, IE11, and Edge

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/02-components/accordion/40-accordion-content-variations.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/02-components/accordion/40-accordion-content-variations.twig
@@ -265,6 +265,25 @@
   } only %}
 {% endset %}
 
+{% set accordion_content %}
+  {% include "@bolt-components-accordion/accordion.twig" with {
+    items: [
+      {
+        trigger: "Accordion item 1",
+        content: "This is the accordion content.",
+      },
+      {
+        trigger: "Accordion item 2",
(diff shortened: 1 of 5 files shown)


sghoweri:
PR was released with v2.11.0

PR #1515, opened 2019-10-24
Update items documentation.

charginghawk:
## Summary

Clarifying what turned out to be a gotcha for me.

## Details

https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-list/src/_list-item.twig#L22-L24


mikemai2awesome:
@charginghawk I've updated the schema along with this PR (https://github.com/boltdesignsystem/bolt/pull/1518). Let's approve that one instead.

PR #1518, opened 2019-10-25
Fix: List component's inline@breakpoint display options not aligning correctly in Firefox

mikemai2awesome:
## Jira

Issue first raised in: http://vjira2:8080/browse/WWWD-4211?focusedCommentId=68872&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-68872

## Summary

Fixes an issue where the List component does not align correctly in Firefox when `display` is set to `inline@breakpoint`.

## Details

1. Added CSS to make sure `.c-bolt-list` container takes up full width of its parent.
2. Updated docs to demo the alignment when List is set to display inline@breakpoint.

## How to test

Run the branch locally and check the following:

- [ ] 1. Given I am viewing the align variations page (http://localhost:3000/pattern-lab/?p=components-list-align-variations)
- [ ] 2. Then I should see inline lists and inline@breakpoint lists
- [ ] 3. And each list's items are aligning correctly according to the align option specified
- [ ] 4. When I view the page in Chrome, FIrefox, Safari, and IE11
- [ ] 5. Then I should see that the alignments are identical across different browsers

### DIFF
--- a/packages/components/bolt-list/src/list.scss
+++ b/packages/components/bolt-list/src/list.scss
@@ -80,7 +80,7 @@
     // The inline here is talking about the items inside, the List component itself is still a block level element that would fill up the space of any container.
     &.c-bolt-list--display-inline,
     &.c-bolt-list--display-flex {
-      width: calc(100% + #{bolt-spacing($spacing-value-name)} + 1px); // Width must be defined in order for the list to dislay correctly in Firefox.
+      width: calc(100% + #{bolt-spacing($spacing-value-name)}); // Width must be defined in order for the list to dislay correctly in Firefox.
     }
 
     @each $breakpoint in $bolt-breakpoints {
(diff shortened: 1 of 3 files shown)


packages/components/bolt-list/src/list.scss
```diff
@@ -77,9 +77,21 @@
     margin-bottom: bolt-v-spacing(#{$spacing-value-name}) * -1;
     margin-left: bolt-spacing(#{$spacing-value-name}) * -1;
 
+    // The inline here is talking about the items inside, the List component itself is still a block level element that would fill up the space of any container.
     &.c-bolt-list--display-inline,
     &.c-bolt-list--display-flex {
-      width: calc(100% + #{bolt-spacing($spacing-value-name)} + 1px); // The inline here is talking about the items inside, the List component itself is still a block level element that would fill up the space of any container.
+      width: calc(100% + #{bolt-spacing($spacing-value-name)} + 1px); // Width must be defined in order for the list to dislay correctly in Firefox.
```
danielamorse: @mikemai2awesome `+ 1px` is causing the list to overflow the container by 1px and causes X scrollbar to appear. I'm guessing 1px is to round up any sub-pixel rendering. This would be an issue only if the list is used directly inside a container with overflow scroll/auto. Any concern here?

mikemai2awesome:
@danielamorse Good catch! That code is obsolete, I've removed the 1px.

packages/components/bolt-list/src/list.scss
mikemai2awesome: Removed the 1px.

danielamorse:
> @danielamorse Good catch! That code is obsolete, I've removed the 1px.

Awesome, thanks @mikemai2awesome!

sghoweri:
PR was released with v2.10.0

PR #1519, opened 2019-10-25
feat(micro-journeys): starter templates for one and two character

glassdimly:
## Jira

https://app.asana.com/0/1126340469288208/1146085721698826/f

## Summary

Add Single and Multiple Pathways starters, add specific One and Two Character layouts from starter docs from Ross.

## Details

The PDF describing these specific layouts can be found on the asana task.

## How to test

This button is added, with two pathways representing one and two character layouts respectively from the PDF:

image

One char: 
image

Two char:
image


margoromo:
In review with @charginghawk and Ross.

charginghawk:
This looks good to me (and it's good to know how to add starters for the future). Once Ross approves, I think we're set.

sghoweri:
PR was released with v2.11.0

PR #1520, opened 2019-10-25
debug(micro-journeys): DO NOT MERGE debug code

glassdimly:
## Jira

Debug PR for drupal, do not merge


remydenton:
Per Slack discussion, the reason for this not working is that the micro journeys package wasn't even getting pulled into Drupal ;).  Closing since it won't be needed.

PR #1528, opened 2019-11-02
feat(micro-journeys): refactor character, dialog, and 2-char layout t…

glassdimly:
…o cooperate with document flow

## Note:

Ready for review

## Tickets:
 https://app.asana.com/0/1126340469288208/1147196921996420/f
https://app.asana.com/0/1126340469288208/1147196921996417/f
https://app.asana.com/0/1126340469288208/1146297619076801/f
https://app.asana.com/0/1126340469288208/1146633993326305/f
https://app.asana.com/0/1126340469288208/1146633993326310/f

## Summary

This is a major refactor of `bolt-character`, `bolt-dialog-bar`, and `bolt-two-character-layout` so that they no longer use absolute positioning and transforms to achieve layout.

Instead, they respect document flow, can be resized by their containers, and tolerate as much text as is needed. Tested on Chrome, FF, and Edge. They look much better on mobile and have incorporate the fixes from Yao.

It is not working in IE 11. I have broken off a ticket for IE 11: https://app.asana.com/0/1126340469288208/1147761775577724/f


packages/core/utils/equalize-relative-heights.js
```diff
@@ -0,0 +1,121 @@
...
+  } catch (e) {
+    tryCount++;
+    if (debug) {
+      console.debug(
+        `debug:persistentlyAttemptToEqualizeRelativeHeights:tryCount: ${tryCount}`,
+        items,
+      );
+    }
+    if (tryCount <= tryLimit) {
+      window.requestAnimationFrame(() => {
```
glassdimly: This is a naive approach that fails on complex pages.

packages/micro-journeys/src/character.js
```diff
@@ -42,16 +46,30 @@ class BoltCharacter extends withLitHtml() {
...
     self = super(self);
     self.useShadow = hasNativeShadowDomSupport;
     return self;
   }
 
+  connectedCallback() {
+    super.connectedCallback();
+    setTimeout(() => {
+      this.dispatchEvent(
+        new CustomEvent(`${BoltCharacter.is}:connected`, {
```
glassdimly: Here you can see I'm triggering an event on `connect`, but this isn't when content is actually loaded and ready for traversal. Many some children still have 0 width.

sghoweri:
> This PR is on hold until we work out the component mount and render issues.

@glassdimly is this still the case or is this good to review?

packages/core/utils/equalize-relative-heights.js
```diff
@@ -0,0 +1,96 @@
+/**
```
remydenton: @sghoweri, did we end up deciding there wasn't an obvious way to handle the use case without needing this util?

sghoweri: @remydenton not without significant rework... 

Side note related to this: before I hit the merge button this I'm gonna move this equalize relative heights util into the Micro Journey package itself give the incredibly specific use case this util is trying to solve.

sghoweri:
PR was released with v2.11.0

PR #1530, opened 2019-11-04
feat(micro-journeys): prevent body content render on inactive steps and prevent inactive pathway render

glassdimly:
## Jira
https://app.asana.com/0/1126340469288208/1144425881076982/f

## Summary
This avoids render on inactive pathways and steps to improve performance.

## Detail
This turned out to take more time than I had expected (naively, I thought simply forgoing render on inactive would work). The challenge here was that when switching to a new Step and Pathway, the content was empty. Then, it would animate in, and then out, the in again. The solution was to render the component to the DOM before switching to it so`bolt-animate` could animate out before animating in. I had to add a new flag to `bolt-animate` which forces a fast animate out.

FYI: I need to merge this into https://github.com/boltdesignsystem/bolt/pull/1528 because it fixes a couple of issues on that PR. So let's review this PR first.

## Testing
Go to pattern-lab/?p=experiments-micro-journeys

Note: to test shadowDomless content, you must manually disable shadowDom in `utils/environment.js` with 
```
export const hasNativeShadowDomSupport = false;
```

This is because the mount/unmount confuses the `form` tag hack.
Screenshot:
image

I have added a note to the Micro Journey component demo.


docs-site/src/pages/pattern-lab/_patterns/06-experiments/micro-journeys/_shadow-dom-demo.twig
```diff
@@ -24,6 +24,7 @@
 <div style="overflow: hidden; max-width: 100vw;">
   <hr>
   <h2>Shadow DOM – yes:☝️ no:👇</h2>
+  Note: Shadow DOM-less content is not working properly, the form tag hack is broken and some children are adding a broken Shadow DOM. To test Shadow DOM-less content you must manually set `export const hasNativeShadowDomSupport = false;` in `utils/environment.js` until this is fixed.
```
sghoweri: Just keep in mind that I do not believe that this "hack"  is "broken". Rather, we're hitting the upper limit of what's reasonably possible with this technique.

If an element being rendered doesn't have a parentNode and/or loses track of where in the DOM tree it's located, then it wouldn't have any way of knowing whether or not it's being rendered within a particular HTML element.

We'll likely be retiring this technique in the near future when we switch to LitElement and come up with a new pattern for more easily testing Shadow DOM-less UI we're building.

sghoweri:
@glassdimly I'm taking another look at this now — should have my review for this done shortly!

sghoweri:
PR was released with v2.11.0

PR #1531, opened 2019-11-05
Core | JS | Move 'optimizedResize' into Core

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1882

## Summary

Move `optimizedResize` event into Bolt Core and rename it `throttledResize`. Add `debouncedResize` event.

## Details
`optimizedResize` was duplicated in a several components. I've refactored into a single file in Core. I've also included a new `debouncedResize` Custom Event, which is currently only used by image but can be used by any other component.

### Fixes/Improvements
- `throttledResize` is now only included once on the page, even if multiple components use it.
- `debouncedResize` event listeners are now properly removed when Images are disconnected. Previously, `this.onResize` was wrapped in `debounce()` and not removed on `disconnecting()` ([see code](https://github.com/boltdesignsystem/bolt/pull/1531/files#diff-9d632ea9e7c654cc2b08adafc542ecf3L99-L149)).

## How to test

- Review files changed
- Checkout `feature/move-optimized-resize-to-core` locally
- Test the following components cross-browser, especially what happens on resize, and verify there are no regressions:
  - Carousel
  - Image
  - Nav Priority
  - Tabs


packages/core/utils/debounce.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Debounce limits the rate at which a function can fire by waiting until that function has ceased to be invoked for N milliseconds.
```
sghoweri: +1 on the docs! Nice!!

I’d have to double check but does the extra line break in between to doc lock and the code matter?

packages/core/utils/optimized-resize.js
```diff
@@ -0,0 +1,42 @@
...
+  function addThrottledEvent(type, name, obj) {
+    obj = obj || window;
+    let running = false;
+
+    function func() {
+      if (running) {
+        return;
+      }
+      running = true;
+      requestAnimationFrame(function() {
```
sghoweri: Can we add a feature detection check here to verify the browser supports raf?

#cantBeTooCareful

sghoweri: ^ not a blocker though! I want to get this merged down ASAP — seeing some promising perf results when testing this out (especially in IE 11)

packages/core/utils/debounce.js
danielamorse: I think you're right. All the examples I see have no line-break. Fixed. ([commit](https://github.com/boltdesignsystem/bolt/pull/1531/commits/2d3239ccc6f1ca7609be5b86a99cc138f6264b03))

packages/core/utils/optimized-resize.js
danielamorse: I hear ya, but Isn't `requestAnimationFrame()` safe to use these days unless you support Opera Mini or IE 9?

https://caniuse.com/#feat=requestanimationframe

Easy enough to add a `setInterval` fallback if you think that's best. Let me know and I'll do it as a follow up.

sghoweri:
PR was released with v2.11.0

PR #1535, opened 2019-11-05
Bugfix/bds 1879 pass renderable content as list items

remydenton:
## Jira

http://vjira2:8080/browse/BDS-1879

## Summary

Allows render arrays to be used as items in list component

## Details

I think I figured out the reason for this "is iterable" pattern.  Specifically, for each cell in table, you might optionally want to specify some attributes for the cell itself, in which case `item` should be an array with both an `attributes` and a `content` key.  However, because that gets tedious if you _don't_ want to specify cell attributes, you can also just pass a string for each cell item as a shortcut.  As @charginghawk discovered though, this doesn't work if your item is a render array.

So, what's the upshot?
- For `bolt-list`, this pattern doesn't make sense because there's no support for separate `attributes` for each item.  Maybe there was and it got removed?  Maybe there's reason to believe we'll want it in the future?  In any case, I've updated the check to be `if item.content` instead of `if item is iterable` so that you can pass a render array for item (as long as it doesn't have a `content`) key.
- For `bolt-table`, I made the same change.  This allows you to use render arrays when using the shorthand and not specifying cell attributes.
- For `bolt-ul` and `bolt-ol`, it's basically impossible to pass a render array for an item because they interpret any array as a nested list.  I don't think there's really any good way out of this.

## How to test

- Go to `/pattern-lab/patterns/02-components-list-__tests__--list-with-object-in-item/02-components-list-__tests__--list-with-object-in-item.html` on this branch and confirm that you see the text `test-attr="test-value"`.  This indicates that something other than a string (in this case, an attributes object) can be passed as a renderable list item.
- You can also run automated nightwatch tests to confirm the same thing (I used `yarn test:e2e:quick-local`).  It should fail without the other commits on this branch and pass with them.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/02-components/list/__tests__/-list-with-object-in-item.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/02-components/list/__tests__/-list-with-object-in-item.twig
@@ -1,0 +1,7 @@
+{% set item_1 = create_attribute({'test-attr': 'test-value'}) %}
+
+{% include "@bolt-components-list/list.twig" with {
+  items: [
+    item_1
+  ]
+} only %}
--- a/packages/components/bolt-list/__tests__/bolt-list.e2e.js
+++ b/packages/components/bolt-list/__tests__/bolt-list.e2e.js
@@ -1,0 +1,22 @@
+let currentBrowser;
+
+module.exports = {
+  tags: ['component', 'list'],
+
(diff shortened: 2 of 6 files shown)


remydenton:
CC @sghoweri and @danielamorse (feel free to review if you want, I just didn't want to add the whole world as reviewers).

sghoweri:
+1 on adding a new e2e test with this!

sghoweri:
PR was released with v2.11.0

PR #1537, opened 2019-11-07
fix: Remove redundant button tag logic and dependency on deprecated schema

remydenton:
## Jira

N/A (ticket based on Slack convo)

## Summary

Remove redundant button tag logic and dependency on deprecated schema

## Details

This removes some redundant/confusing twig logic related to processing the deprecated tag prop.  In summary:

* The button will no longer break if the `tag` section of the schema is deleted.
* The only place that a user-supplied `tag` prop has any effect is in the DEPRECATED block near the top of the twig file, where it's converted to the appropriate type (although a `tag` variable is used later in the file, it's essentially an internal variable and will never inherit from a passed twig prop)
* The logic for determining the tag has been simplified to the following:
    * If a `url` param is present, the tag will always be `<a>`
    * Otherwise, the tag will always be `<button>`
* Logic for determining the type attribute has been simplified to the following:
    * If a `url` param is passed, there will be no type prop
    * Otherwise, `type` prop will be a valid type ("reset" or "submit") or "button" as a default. 

Note: there is a minor change of logic here: if you pass both a `url` and a `type` prop, url will always win and there will be no type.  Previously, if you passed both a `url` and `type=submit` or `tag=submit`, the type/tag would win.  I don't consider this a breaking change because there is no valid reason to do that and it's effectively unsupported.

## How to test

Confirm no regressions


remydenton:
This is a surprising test result.  https://travis-ci.com/boltdesignsystem/bolt/jobs/254233892#L1483-L1484

It looks like the results on this PR are actually the ones I would expect, which would mean it fixed a previous bug.  I wasn't expecting it to though...
https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-typeahead/typeahead.twig#L108
https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-typeahead/typeahead.twig#L122

remydenton:
Here's another test fail that I would actually expect: https://travis-ci.com/boltdesignsystem/bolt/jobs/254233892#L1781

Previously, if you passed the (deprecated) `tag` prop but no `url` prop, you would get a non-sensical `<a>` tag without an href.
With the updates, if you do the same thing, you instead get a `<button>` tag.

This is similar to my note in the PR description.  Trying to specify an `<a>` tag without passing a `url` is contradictory, so we choose to honor the url (or lack thereof) instead.  It is technically a backwards-breaking change, but for an unsupported/non-sensical combination of props.

remydenton:
> It looks like the results on this PR are actually the ones I would expect, which would mean it fixed a previous bug.

Aaaand just kidding.  Looks like I introduced a bug. 😅 

Testing coverage: 1
Remy: 0

remydenton:
Thanks for the post-merge test updates @sghoweri, all the snapshot changes look to be as-expected.

remydenton:
Should be all set @danielamorse, ready for another review.

sghoweri:
PR was released with v2.11.0

PR #1546, opened 2019-11-12
Fix/bolt character ie 11

glassdimly:
## Jira

https://app.asana.com/0/1126340469288208/1147761775577724/f

## Summary

Make refactored `bolt-character` and `bolt-two-character-layout` work on IE 11.

## Details

Everything looks good on IE 11. I had to set some fixed widths, as you can see, because IE 11 really didn't want to use `calcs` and allow things to have their own widths no matter what I tried, but I made some classes that give more space to `bolt-character` given whether or not one or both sides have content, so it's looking pretty good.

## How to test

On IE 11 and a non-IE 11 machine, see:
pattern-lab/patterns/06-experiments-micro-journeys-15-character-kitchen-sink/06-experiments-micro-journeys-15-character-kitchen-sink.html
and 
pattern-lab/patterns/06-experiments-micro-journeys-90-micro-journeys/06-experiments-micro-journeys-90-micro-journeys.html


packages/global/styles/07-utilities/_utilities-ie-flex-aspect.scss
```diff
@@ -0,0 +1,10 @@
+.u-bolt__flex-aspect {
+  max-width: 100%;
+  display: block;
+}
+
+.u-bolt__flex-aspect-full {
+  width: 100%;
+  //min-width: 100%;
+  display: block;
+}
```
remydenton: These don't follow the naming conventions of global utility classes (hyphens only, no underscores).

But beyond that, I vote for not having these be global utility classes.  We generally try to be quite conservative with adding or modifying those since once they're in the wild they can have very wide reaching ramifications.  Also, none of our other utility classes are browser-specific, and as far as I can tell this is a pretty limited use case.  Can we refactor this as classes in the `c-bolt-character-*` namespace instead?

sghoweri: Yes please — 100% agree with @remydenton's comment here on our coding standards

glassdimly:
I'm integrating these changes now, thanks all.

sghoweri:
PR was released with v2.11.0

PR #1547, opened 2019-11-12
BDS 1909 JS schema parent child example

remydenton:
## Jira

http://vjira2:8080/browse/BDS-1909

## Summary

Add example of parent-child js schema using accordion component

## Details

A few minor questions:
* Should all parent and child schemas live at the root of the component directory or in appropriate subdirectories?
* Did I get the naming convention right for the child component schema file (kebab-case, not camelCase)?

## How to test

Confirm no regression in the accordion component documentation or functionality.
https://boltdesignsystem.com/pattern-lab/?p=viewall-components-accordion


sghoweri:
@remydenton 

> A few minor questions:

> Should all parent and child schemas live at the root of the component directory or in appropriate subdirectories?

If the component is flat (no sub-folders / doesn't need them), then we should keep it flat. 

If the component has separate folders for sub-components, then I'd recommend keeping the sub-component schema with the related sub-component pieces.

> Did I get the naming convention right for the child component schema file (kebab-case, not camelCase)?

I THINK so... I'm still not 100% sure about the whole Capitalized Camel Case naming convention I've seen with components... kebab-case (atm) is still the gold standard we're running with.

remydenton:
> If the component has separate folders for sub-components, then I'd recommend keeping the sub-component schema with the related sub-component pieces.

You mean like accordion item @sghoweri (i.e. `accordion-item.schema.js` should be moved to packages/components/bolt-accordion/src/AccordionItem/)?  I can certainly move it.

sghoweri:
PR was released with v2.13.0

PR #1552, opened 2019-11-13
Feature/editor save events

WessWillis:
## Jira

https://app.asana.com/0/1126340469288208/1145996329931297/f

## Summary

Editor close now awaits response from drupal.  

## Details

Previously, editor save events utilized a Custom Event, and the editor closes with no knowledge or feedback on it's success.   This approach awaits a function attached to the window object, allowing the editor to reliably save knowing that the drupal update was successful, or to display an error and remain open if it fails.  

There is an example function provided which includes asyc simulation.  It should be overwritten by Drupal.

## How to test

Open the micro journeys page in pattern lab, edit a page and then save.  To see a failure, update  `const shouldThisFailToDemo = false;` on line 68 in `/packages/editor/src index.js` and repeat the process.


remydenton:
Also, one last thing, ideally this wouldn't include changes to the `.incache` file.  @sghoweri, is there a reason that's not excluded in git?

https://github.com/boltdesignsystem/bolt/pull/1552/files#diff-645222a46eb9ed6e19866a193c62e0f2

sghoweri:
> Also, one last thing, ideally this wouldn't include changes to the `.incache` file. @sghoweri, is there a reason that's not excluded in git?
> 
> https://github.com/boltdesignsystem/bolt/pull/1552/files#diff-645222a46eb9ed6e19866a193c62e0f2

@remydenton the short answer: we're including that in Git to speed up local development + avoid hitting API rate limits with Github.

I've considered adding it to our `.gitignore` however we'd need to test to make sure non-authenticated Github API requests aren't going to come back and bite us :)

sghoweri:
We regenerate with every new release anyways so I'm not worried

packages/editor/src/index.js
```diff
@@ -55,6 +56,41 @@ function addGrapesCssToPage() {
   });
 }
 
+/**
+ * Example function for how to save editor results
+ * Assign a function like this to `window.__handleEditorSave` to override this example one.
+ * @param {Object} opt
+ * @param {string} opt.html - The HTML from the saved editor
+ * @param {string} opt.id
```
remydenton: This docblock is also outdated.  The function doesn't support an `opt` param.

sghoweri:
PR was released with v2.11.0

packages/editor/src/index.js
EvanLovely: That's actually how you document a destructured object - look close, the function takes only 1 param: an object with the properties of `html` and `id`. Those are documented as `opt.html` and `opt.id`

packages/editor/src/index.js
```diff
@@ -1,6 +1,7 @@
 import { query } from '@bolt/core/utils';
 import { version } from 'grapesjs/package.json';
 import { detect } from 'detect-browser';
+import { golang } from 'locutus';
```
EvanLovely: This should NOT be there! I'm almost positive this is the code editor automatically adding an import. Pull this out immediately @sghoweri

EvanLovely: @WessWillis was this intentional?

packages/editor/src/index.js
remydenton: Yep, good call!  Thanks for clarifying.

EvanLovely:
@remydenton don't hesitate to reach out to me with any implementation questions

packages/editor/src/index.js
sghoweri: Removing right now

packages/editor/src/index.js
```diff
@@ -223,30 +260,24 @@ function init() {
             );
           }
           const container = editor.getContainer();
-          cleanup();
-          container.innerHTML = html;
```
remydenton: This line, `container.innerHTML = html;`, has been moved lower down, after `handleEditorSave()`.  However, `container.innerHTML` gets passed as an argument to `handleEditorSave()`.  The upshot is that handlerEditorSave() no longer receives the correct value for `html`.

packages/editor/src/index.js
```diff
@@ -223,30 +260,24 @@ function init() {
...
-            /** @type {HTMLElement} */
-            animEl => {
-              if (animEl.innerHTML === '') {
-                animEl.remove();
-              }
-            },
-          );
 
-          trigger.innerText = 'Edit';
-          editorState = EDITOR_STATES.CLOSED;
```
remydenton: This line, `editorState = EDITOR_STATES.CLOSED;`, was removed in this PR and not added back.  I think updating the editorState is necessary to prevent some bugs I'm seeing on the Drupal end.

remydenton:
Heads up, I'm going to try testing out some fixes to my last two comments directly on the Drupal end in hopes of avoiding more than one more Bolt release.  I'll create a PR against this repo if my changes work.

PR #1556, opened 2019-11-15
Feature/bds 1911 icon cleanup

christophersmith262:
## Jira

http://vjira2:8080/browse/BDS-1911

## Summary

Currently, the way icon builds / custom icons work creates a tight coupling between the components-icon package, components-icons package, and the build process itself. The goal of this PR is to remove the circular dependency between components-icon and components-icons in the build process, allowing packages other than components-icons to provide icons, without directly altering components-icon(s).

## Details

There are no user facing changes here. It just consolidates logic for icon generation into a package-independent library (utility/icon-generator.js) that can be called for any package in the system to package icons in that package.

Most of the logic from the icon task has been moved there. There is one holdout "generateSchemaFile", which can't be made package independent yet, so it seems, but we should address that in a follow up.

## How to test

Test:
1. Building bolt and viewing icons in pattern lab.
2. Doing a build using the documented custom icon procedure, and testing the result.

### DIFF
--- a/packages/build-tools/tasks/icon-tasks.js
+++ b/packages/build-tools/tasks/icon-tasks.js
@@ -23,9 +23,7 @@ const failedRebuildingIconsMsg = 'Failed to rebuild Bolt SVG Icons!';
 
 async function build() {
   const iconSpinner = new Ora(
-    chalk.blue(
-      initialBuild ? startBuildingIconsMsg : startRebuildingIconsMsg,
-    ),
+    chalk.blue(initialBuild ? startBuildingIconsMsg : startRebuildingIconsMsg),
   ).start();
 
   try {
@@ -37,7 +35,7 @@ async function build() {
       extendedIconDirs,
     );
 
-    await function generateSchemaFile(icons);
+    await generateSchemaFile(icons);
 
     iconSpinner.succeed(
       chalk.green(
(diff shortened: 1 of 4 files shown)


sghoweri:
For what it's worth, I am very much in favor of further decoupling our build tasks + how we can make the build tools (ie. task collections) more customizable.

sghoweri:
One other thing I'm looking into with this: how does this work tie in with the larger Icon component refactor work (complete overhaul) of how we bundle together and use our SVG icons.

christophersmith262:
> where / what / when exactly would we be calling this outside of the build tools?

We wouldn't ever call this outside of build tools. See http://vjira2:8080/browse/BDS-1911.  The idea is that the build tools can discover components from .boltrc.js that have an "iconDir" property in their package.json and apply `generateIcons` to the package.

> One other thing I'm looking into with this: how does this work tie in with the larger Icon component refactor work (complete overhaul) of how we bundle together and use our SVG icons.

That's a good question.  I'm not sure what the larger changes with icons are.  Is that on a branch somewhere?

sghoweri:
> That's a good question. I'm not sure what the larger changes with icons are. Is that on a branch somewhere?

Working on it! 🙂

sghoweri:
> We wouldn't ever call this outside of build tools. See http://vjira2:8080/browse/BDS-1911. The idea is that the build tools can discover components from .boltrc.js that have an "iconDir" property in their package.json and apply generateIcons to the package.

@christophersmith262 just saw that! Now THAT is a neat idea! Auto-add custom SVG assets that are called out in the component's `package.json` file... HMMMM 🤔

sghoweri:
PR was released with v2.13.0

PR #1565, opened 2019-11-21
feat: add `children` to figure schema, setup basic demo

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1781

## Summary

To bring Twig schema more in line with WC schema, I'm demoing a potential new pattern: passing inner renderable items (`children`) as an array of objects, with an optional slot.

Relates to #1548.

## Details

## How to test

(List clear steps to test that the changes are working)

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/02-components/figure/05-figure.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/02-components/figure/05-figure.twig
@@ -1,9 +1,17 @@
+{% set image %}
+  {% include "@bolt-components-image/image.twig" with {
+    src: "/images/placeholders/500x500.jpg"
+  } only %}
+{% endset %}
+
 {% include "@bolt-components-figure/figure.twig" with {
-  media: {
-    image: {
-      src: "/images/placeholders/500x500.jpg",
-      lazyload: false
+  children: [
+    {
+      component: image,
+      slot: 'media'
     },
-  },
-  caption: "Fig. 1: This is Bill. He is awesome."
+    {
+      component: "Fig. 1: This is Bill. He is awesome."
+    }
+  ],
 } only %}
--- a/packages/components/bolt-figure/figure.schema.yml
+++ b/packages/components/bolt-figure/figure.schema.yml
@@ -4,6 +4,18 @@ properties:
   attributes:
     type: object
     description: A Drupal-style attributes object with extra attributes to append to this component.
+  children:
+    type: array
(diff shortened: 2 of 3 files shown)


danielamorse:
@colbytcook @remydenton closing this PR. I don't see us going in this direction, using a generic "children" prop to pass an array of items, which are then sorted into slots inside the Twig template.

I think it's simpler and more consistent to add flat, component-specific props with type "array" when we need to do this sort of thing, and handle it on a case-by-case basis, rather than introducing a whole new "children" pattern.

Keeping branch open if you want to take a look.

remydenton:
Thanks @danielamorse, I think that's the right decision.  

As a general rule, I'd like to keep components as decoupled as possible.  So whenever possible (and I think 95% of the time it is), instead of this kind of thing:

```
{% if image %}
    {% include "@bolt-components-image/image.twig" with image only %}
{% elseif icon %}
    {% include "@bolt-components-icon/icon.twig" with icon only %}
{% elseif video %}
    {% include "@bolt-components-video/video.twig" with video only %}
{% elseif table %}
    {% include "@bolt-components-table/table.twig" with table only %}
{% endif %}
```

just do this, leaving the inclusion of the child pattern to the consumer:

```
{{ content }}
```

Is there a reason we couldn't do that in this case?  Not that we need to spend time cleaning it up now, just curious if this is a use case where my proposed pattern wouldn't work.

danielamorse:
@remydenton agreed we shouldn't do this:
```
{% if image %}
    {% include "@bolt-components-image/image.twig" with image only %}
{% elseif icon %}
    {% include "@bolt-components-icon/icon.twig" with icon only %}
...
```
`{{ content }}` suffices. It leaves the door open to "unapproved" content, but that's something we can address with documentation rather than strictly preventing it via schema.

The "children" pattern gives us a little more control over how each item is output. You might put one item in a slot or wrap another in a `<div>`. Still, these are things we can do already by other means.

remydenton:
Cool.  I think a children pattern is fine if you're trying to add wrappers.  I would still keep it agnostic though, i.e.:
``` 
{% for children as child %}
  <div>
    {{ child }}
  </div>
{% endfor %}
```

The ability to pass unapproved content is one of the few downsides to that approach.  IMO, other approaches that force certain content have far more downsides, and I have yet to see a case where someone actually tries to pass something unapproved.  At the end of the day, it's the same as passing an invalid value for any of the other props (there's nothing stopping you, but it likely won't work).

PR #1573, opened 2019-11-26
Bolt band in modal fix

adamszalapski:
## Jira

http://vjira2:8080/browse/WWWD-4249

## Summary

When we use a bolt band with modal and a lot of content was added to this modal in mobile Safari top of the content was cut.

## Details

I update styles in bolt band and compare is are some differences before and after the change.
compare-band-versions
Nothing changes after the update but i must update band snapshots to pass new styling.
Right now in Safari and other browsers bands in modal looks good.

## How to test

Check if all test is passed and if this page https://boltdesignsystem-rjqreb3m5.now.sh/pattern-lab/patterns/02-components-modal-50-modal-usage-content/02-components-modal-50-modal-usage-content.html is displayed correctly in Safari and other browsers in mobile view.


danielamorse:
@adamszalapski I pushed a potential fix ([see changes](https://github.com/boltdesignsystem/bolt/pull/1573/files#diff-f55ebcb122224a28748df4f3ab670403)).

To fix the Safari bug I had to change `flex: 1` to `flex: auto` on `.c-bolt-modal__container-body` and remove `height: 100%`. Setting `height: 100%` or `flex: 1` alone would trigger the bug in Safari, so both had to change. 

The `height` property should be safe to remove because it's a flex-item which will grow to fill the available space anyway. I'm not sure why `flex: auto` was required, but `auto` sets `flex-basis` to `auto` rather than `0`, and that must matter to Safari when it aligns the flex items in this particular case 🤷‍♂ 

Then, IE11 was still having the same issue, so there I hooked into the `.is-first-child` class to set it to `align-items: flex-start`. This hack is less than ideal, but I figure better that mobile modal content is aligned top than running off the page.

I tested the band demo in Chrome, Safari, IE11, Edge. @adamszalapski, would you please test the band demo and some of the others for regressions? I'd feel a lot better about the fix with some more testing 😄 

cc: @sghoweri @mikemai2awesome

adamszalapski:
Thanks, @danielamorse for your help. I do some testing and create JEST test for modal with the band. It looks like everything working as expected. Please check it on your end: @sghoweri @mikemai2awesome

adamszalapski:
@sghoweri any idea why in long text examples on local tests are passing but are failing on Travis. I noticed that on Travis modal is bigger but I don't know why.

packages/components/bolt-button/__tests__/button.js
```diff
@@ -4,12 +4,20 @@ import {
   renderString,
   stopServer,
   html,
+  vrtDefaultConfig as vrtConfig,
```
mikemai2awesome: Hi @adamszalapski, is this related to band or modal? Why did we need to change the button test?

adamszalapski: I change the button test because of one a while (not always) button test ware failing and right now i adjust the configuration of button test.

mikemai2awesome: is it needed in this PR or should it be in its own PR? feels like a separate issue from band and modal.

adamszalapski: i can remove it but i don't know if when i push the code test not fail on the button. I don't want to leave this on luck but if we decide to remove these changes i can remove them.

adamszalapski: I remove all code related to button

packages/components/bolt-modal/src/modal.scss
```diff
@@ -110,10 +111,16 @@ bolt-modal:not([ready]) {
       @include bolt-margin-top(auto);
       @include bolt-margin-bottom(auto);
 
-      flex: 1;
-      height: 100%;
+      flex: auto;
```
danielamorse: @sghoweri you mentioned that we shouldn't use shorthand for `flex` because it behaves differently across browsers, but I can't actually find any docs confirming that, at least for the browsers we care about.

Can you confirm this is still an issue?

sghoweri: Most of the worst reported flexbox bugs (at least those involving the shorthand syntax) are in IE 10 however [this one](https://github.com/philipwalton/flexbugs#flexbug-4) might still be applicable.

In any case, I’m not going to raise a stink about changing this particular instance using the longhand syntax given the testing you guys have done.

_However_, as a general rule of thumb (as a soft rule), I still encourage us to try to stick to using the longhand syntax for most things... 

https://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/

sghoweri:
PR was released with v2.13.0

PR #1577, opened 2019-11-26
Fix Micro Journey-related Issues with the Two-Character Layout

glassdimly:
This fixes several `bolt-two-character-layout` issues, including:
* https://app.asana.com/0/1126340469288208/1130525745807684/f
* https://app.asana.com/0/1126340469288208/1151276850580125/f
* Connection band disappears in editor
* Connection band background is sometimes invisible in Chrome
* `bolt-two-character-layout` layout problems in IE

Detail:
In general I simplified the code here a bunch and made it readable and maintainable. I also added a persistent `setTimeout` promise to ensure that child components have `offsetHeight`. This is because bolt `render` event was reporting that components were rendered when they had not yet been painted on the dom. I also tried mutation observers on both shadow and non-shadow components, as well as `requestAnimationFrame` and `setTimeout(0)`

How to test:
Visit http://localhost:3000/pattern-lab/?p=experiments-micro-journeys and click around
Visit http://localhost:3000/pattern-lab/?p=experiments-editor-simple, drag in the 2-character starter, and hit Edit->Save->Edit->Save->Edit->Save. Character layout should be good
Check out http://localhost:3000/pattern-lab/?p=experiments-micro-journeys in IE 11 and Edge
In Chrome, click http://localhost:3000/pattern-lab/?p=experiments-micro-journeys and click the "Empowered Agents" step. Connection should retain background.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/06-experiments/editor/25-editor-with-two-character-layout.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/06-experiments/editor/25-editor-with-two-character-layout.twig
@@ -1,0 +1,14 @@
+{% set content %}
+  <bolt-interactive-pathways theme="{{ theme ? theme }}">
+    <bolt-text subheadline font-size="xxlarge" slot="interactive-pathways-lead-text">How Pega technology resolves</bolt-text>
+    <bolt-interactive-pathway pathway-title="Billing Inquires">
+      <bolt-interactive-step tab-title="Proactive service that meets or exeeds customer needs and offers a high return on investment to customers">
+        {% include "./billing-micro-journey-fragments/_step-one.twig" %}
+      </bolt-interactive-step>
+    </bolt-interactive-pathway>
+  </bolt-interactive-pathways>
+{% endset %}
+
+{% include '@bolt-components-editor/editor-in-pl.twig' with {
+  content: content,
+} only %}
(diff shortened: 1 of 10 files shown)


glassdimly:
@remydenton What's the status of this review?

remydenton:
Thanks for checking @glassdimly.  I'm still working through this one, want to make sure it solves all the bug reports that we suspect are related.

glassdimly:
@remydenton  Darn. My only thought here is to increase `attemptTimeout` to `1000`. If that fixes it, then it's just a reeally slow render. We would have to think about that. Otherwise, I have no idea why the characters are not getting rendered to the DOM and would have to look at it.

glassdimly:
Darn. This doesn't make a whole lot of sense to me, because it means that _characters aren't rendering to the dom_. 

I am setting up Drupal. Will you be able to make some time today to walk me through how to actually add the JS to Drupal?

glassdimly:
@remydenton We figured out that the Drupal problems here have to do with importing changes into the Editor in order to incorporate the new templates; so this is ready for review!

remydenton:
I've confirmed that commenting out [this line that adds js to the editor](https://github.com/boltdesignsystem/bolt/blob/master/packages/editor/editor-in-pl.twig#L3) in pattern lab changes nothing in editing mode.

It doesn't affect this PR, but it does make the whole editor component harder to understand, integrate, and troubleshoot and should be removed IMO.

sghoweri:
PR was released with v2.13.1

glassdimly:
@remydenton,

Great catch. You have proved that the `window` `bolt-global.js` already applies in the editor and is not needed by commenting it out, and that's what I would do if I were going to confirm.

PR #1597, opened 2019-12-06
fix(editor): remove false error on editor navigation away from saved …

glassdimly:
…content

## Jira

https://app.asana.com/0/1126340469288208/1150897080194573/f

## Summary

Remove false error on editor navigation away from saved content.

## Details

Previously, the editor setting `noticeOnUnload` was set to `PROD` which meant it warned when navigating away from content on prod only. However, we're handling the save event and throwing an `alert` when the drupal save promise fails. So this is no longer needed.

## How to test

-     Go to /pattern-lab/patterns/06-experiments-editor-10-editor-simple/06-experiments-editor-10-editor-simple.html
-     Click edit
-     Make a change
-     Click save (IMPORTANT!)
-     Navigate to another page

### DIFF
--- a/packages/editor/src/index.js
+++ b/packages/editor/src/index.js
@@ -252,12 +252,6 @@ function init() {
             // @ts-ignore
             window.__handleEditorSave || exampleHandleEditorSave;
           const html = editor.getHtml();
-          const unsavedChanges = editor.getDirtyCount();
-          if (unsavedChanges) {
-            console.warn(
-              'There were unsaved changes we will lose on next page reload...',
-            );
-          }
           const container = editor.getContainer();
 
           trigger.innerText = 'Saving...';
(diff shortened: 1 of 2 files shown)


glassdimly:
Thanks for your research, @remydenton!

I was able to easily ensure that the notice happens when changes are not saved. I also remove `IS_PROD` on `noticeOnUnload` because I think it's better if the editor behaves in a consistent way across platforms and I don't see why it shouldn't give notice on unload on dev.

packages/editor/src/index.js
```diff
@@ -262,6 +262,7 @@ function init() {
           });
 
           if (ok) {
+            window.onbeforeunload = null;
```
remydenton: This works, though I think it deserves an inline comment since it's not at all obvious what it's doing if you don't know the back story.

Also, any particular reason you decided it doesn't belong inside `cleanup()`?

glassdimly: Good ideas; done!

sghoweri:
PR was released with v2.13.0

PR #1599, opened 2019-12-06
Fix/wwwd 4357 pathway ready event only when ready

remydenton:
## Jira

http://vjira2:8080/browse/WWWD-4357

## Summary

Fixes support for analytics click tracking on `bolt-interactive-pathway` elements

## Details

Previously, the `bolt-interactive-pathway` component would emit a `ready` event before there was any content in the ShadowDOM.  That was caused by the following line which would result in an empty render under certain conditions https://github.com/boltdesignsystem/bolt/blob/master/packages/micro-journeys/src/interactive-pathway.js#L170

This PR updates the `rendered()` function so that it will only run (and, in turn, trigger the "ready" event) if those same conditions are met.

## How to test

- Run the jest test.  It should fail without the fix commit on the branch, and pass after it's applied.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -31,7 +31,7 @@ module.exports = {
   },
   transformIgnorePatterns: [
     'node_modules/(?!(lit-html|@bolt|@open-wc)/)', // add any additional packages in node_modules that need to be transpiled for Jest
-    'packages/(?!(components|core|analytics|config|testing|generators)/)', // add any additional packages in node_modules that need to be transpiled for Jest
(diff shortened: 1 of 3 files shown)


glassdimly:
Also just want to say: I love the test on this branch. Great to have some tests in place. 🎈 🎂 🥇 🥓 !

remydenton:
Looks like the `customElements is not defined` errors isn't specific to this branch, nor even to microjourneys.  I see it on the current master branch, for example. when trying to view the blockquote component:

Screen Shot 2019-12-09 at 1 05 28 PM

TLDR, it shouldn't need to block review and merging of this PR.  We'll open a ticket for it on the Bolt JIRA board.

sghoweri:
PR was released with v2.13.0

PR #1632, opened 2019-12-20
Hotfix: Fix IE11 Typeahead bugs

danielamorse:
## Jira

- http://vjira2:8080/browse/BDS-1928
  - http://vjira2:8080/browse/BDS-1931
  - http://vjira2:8080/browse/BDS-1930

## Summary

Fixes IE11 Typeahead bug that prevents search terms from submitting, breaks autocomplete menu.

## Details

In IE11, when the Typeahead is inside a Band, the component connects and disconnects multiple times as the page renders. In our code, we are removing all external event listeners on disconnect to prevent duplicate event listeners from being added. This breaks IE11 because the event listeners are never re-added when the component re-connects. This PR adds a workaround so that every time Typeahead re-connects the "ready" event will fire, and any external event listeners added initially will be re-added.

**Key code changes:**
- Unset `_wasInitiallyRendered` on Typeahead and Autosuggest on disconnect so that the 'ready' event fires when they re-connect, and any external event listeners will be re-added. Fixes IE11 bug.
- Update the docs according to new suggested use.

While debugging this issue, I fixed a few other bugs that were making the job harder. I thought they were important enough to include them here:

**Additional bug fixes**
- Prevent multiple form submissions by setting `preventDefault` on submit button click.
- Fix IE11 JS error when you click submit button by removing `new KeyboardEvent`, which doesn't work in IE. Just call the `onSelected` method directly.
- Fix the order of arguments passed to the `onChange` event, was causing the event value (i.e. 'up', down') to be passed as the input value (i.e. the search term)

Along with this hotfix, there are minor changes required on the Drupal side so that the event listeners are added as shown in our updated Typeahead demos. I have made these changes locally and will share the code with the Drupal team.

## How to test
Check out branch locally. Open Typeahead demos one at a time in a new Tab. Test the following in Chrome, IE11 and Edge:
- Submit a search by pressing 'enter' key
- Submit a search by clicking submit button (search icon)
- Select an autosuggested result
> Open one demo at a time in a new tab, _not in Pattern Lab Iframe_. Close the Pattern Lab UI all together. Otherwise it may trigger multiple windows to open.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/02-components/typeahead/_typeahead-demo.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/02-components/typeahead/_typeahead-demo.twig
@@ -5,7 +5,15 @@
 
 {{ typeahead_markdown_docs | markdown }}
 
-{% include twig_path %}
+{% set demo_content %}
+  {% include twig_path %}
+{% endset %}
+
+{% include "@bolt-components-band/band.twig" with {
+  content: demo_content,
+  theme: "none",
+  size: "none"
+} only %}
 
 <bolt-code-snippet syntax="light" lang="twig" class="u-bolt-block u-bolt-margin-top-medium">
   {% spaceless %}
--- a/packages/components/bolt-typeahead/__demos__/dynamically-fetch-data/typeahead.dynamically-fetch-data.js
+++ b/packages/components/bolt-typeahead/__demos__/dynamically-fetch-data/typeahead.dynamically-fetch-data.js
@@ -3,48 +3,54 @@ const dynamicTypeaheadDemo = document.querySelector(
   '.js-typeahead-hook--dynamically-fetch-data',
 );
 
-if (dynamicTypeaheadDemo) {
-  dynamicTypeaheadDemo.addEventListener('ready', function(e) {
-    if (e.detail.name === 'bolt-typeahead') {
-      // note: make sure to let Typeahead know when the data fetched is ready
-      dynamicTypeaheadDemo.on('getSuggestions', async value => {
-        return await new Promise(async resolve => {
-          await fetch('/build/data/typeahead.data.json')
-            .then(function(response) {
-              return response.json();
-            })
-            .then(function(data) {
-              return resolve(data);
-            });
+const setupEventHandlers = () => {
+  // note: make sure to let Typeahead know when the data fetched is ready
+  dynamicTypeaheadDemo.on('getSuggestions', async value => {
+    return await new Promise(async resolve => {
+      await fetch('/build/data/typeahead.data.json')
+        .then(function(response) {
+          return response.json();
+        })
+        .then(function(data) {
+          return resolve(data);
         });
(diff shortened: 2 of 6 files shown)


sghoweri:
> @danielamorse code looks good but i find some inconsistent and strange behavior on Edge and IE.
> In Chrome when typing something and click on the icon, press the enter key or choose from autosuggestion the behavior is always the same, open in the same window but in Edge, IE, Firefox and Safari is different.
> 
> **Egde case:**
> 
> 1. Pressing enter - website open in the same window
> 2. Clicking the icon - open two windows one in the same window and second in the new card
> 3. Choosing from autosuggest - open website in the same window
> 
> **IE case:**
> 
> 1. Pressing enter - website open in the same window
> 2. Clicking the icon - opens the website in the new card
> 3. Choosing from autosuggest - open website in the same window
>    And in IE autosuggest dropdown looks little broken (don't have background).
>    image
>    Like the behavior, it's not a critical bug that i think we need to fix the visual issue on IE.
> 
> **Firefox case:**
> 
> 1. Pressing enter - open two windows one in the same window and second in the new card
> 2. Clicking the icon - open two windows one in the same window and second in the new card
> 3. Choosing from autosuggest - open website in the same window
> 
> **Safari case:**
> 
> 1. Pressing enter - opens the website in the new card
> 2. Clicking the icon - opens the website in the new card
> 3. Choosing from autosuggest - open website in the same window
>    We should have the same behavior in all browsers.

@adamszalapski is that screenshot from testing locally or from testing the compiled (prod) version on now.sh?

sghoweri:
^ regarding the duplicate behavior (opening in the same window + separate window), I'm wondering if we need to nuke the default event (e.preventDefault) before running the custom logic? 🤔

danielamorse:
@adamszalapski I can't reproduce that search behavior or the dropdown menu transparency.

I do see some strange behavior when I test inside the Pattern Lab iframe (i.e., not in a separate tab). In that case multiple tabs do open. This can be ignored. Please test in a separate tab and close all other Typeahead tabs, including in other browsers. Sorry, I should've called that out earlier. I've updated the testing steps accordingly.

The transparency issue may be due to some missing CSS. Start your local server, re-save [this file](https://github.com/boltdesignsystem/bolt/blob/master/docs-site/src/templates/_page-header.twig), and check again.

If you are still seeing these issue, let's chat on Monday. Thanks again for the review!

@sghoweri @remydenton do you see these issues when you follow the updated testing steps?

sghoweri:
> Actually - @danielamorse can we chat about this once you get in? Things still seem to behave a bit odd for me in IE 11 so I just want to make sure what I'm seeing is expected...

Confirmed that the fix is still working on Sauce Labs  + on our local testing machine so I'm assuming the issue I was seeing locally is limited to my version IE 11 from modern.ie. 

Marking as approved and moving forward with the v2.13.3 patch release.

sghoweri:
PR was released with v2.13.3

PR #1633, opened 2019-12-21
fix(micro-journeys): black connection band in Safari

glassdimly:
## Jira

https://app.asana.com/0/1126340469288208/1154819220959425/f

## Summary

Fix connection band turning black in Safari.

## Details

Previously, the connection band was going transparent in Chrome. I therefore added logic to trigger a repaint of this component in Chrome. However, it turns out that this very logic caused the connection band to turn black in Safari.

So now I am checking whether or not the browser is chrome before triggering `refreshLinearGradient`.

😞 

## How to test
/pattern-lab/?p=experiments-micro-journeys
* click back and forth between the first step and "Empowered Agents" several times.
* See that in Safari, connection band remains not-black
* See that in Chrome, connection band maintains its color.


sghoweri:
PR was released with v2.14.0

PR #1636, opened 2019-12-22
fix(micro-journeys): ie11 shorten connection length in two-character-…

glassdimly:
…layout

## Jira

https://app.asana.com/0/1126340469288208/1154819220959450/f

## Summary

- Shorten connection length in two-character-layout in IE 11
- Fix mistaken both-side-content class; add class for in-two-char-layout
- Fix IE 11 styling in two-char layout
- Add button to trigger animate in on PL demo
- Add example of 2-char layout with side content to PL

## Details
I fixed the elongation of the connection band in IE 11 relative to the comps and to other browsers
I added a `two-character-layout` with side content to the PL demo here: `pattern-lab/?p=experiments-character-kitchen-sink`
I fixed that the "both-side-content" class was getting added erroneously to all content with just a one side content.
I added a new class to `character` that checks to see if `character` is inside a `two-character-layout` and adds a class. Then, in IE 11, I opt out of the extra spacing afforded to IE 11 `character` if it has side content. This was intended only for stand-alone characters.

## Testing
/pattern-lab/?p=experiments-character-kitchen-sink in IE 11, see new Two-Character layout with side content, confirm that it looks good.


sghoweri:
PR was released with v2.15.0

sghoweri:
PR was released with v2.15.0

PR #1644, opened 2019-12-27
Popover | FE | New component to present small overlay

danielamorse:
## Jira

(Enter a link to the Jira issue this PR applies to, if any)

## Summary

Introduces a new component called popover: a small overlay that opens on demand.

## Details

1. Proper `aria-` attributes are in place to ensure accessibility.
2. `content` can accept any elements because popover should be used to present more structured information and calls to action.
3. Twig and JS are fully built out.

### Still to do

- [ ] Props demo
- [ ] Web component demo
- [ ] Tests
- [ ] Add escape key support to close? CC @danielamorse @mikemai2awesome 

## How to test

Pull down the branch and run locally. Review the code and the docs.


packages/components/bolt-popover/popover.schema.yml
```diff
@@ -0,0 +1,53 @@
+title: Popover
```
sghoweri: @danielamorse @mikemai2awesome why do we have two different schema files here -- one `.js` and the other `.yml`? 

1. What's the difference between the two?
2. If we only have open (the JS one ideally), can we get rid of the Yaml one + change the package.json schema key?

packages/components/bolt-popover/src/popover.js
```diff
@@ -0,0 +1,207 @@
...
+    super.connectedCallback && super.connectedCallback();
+    this.setAttribute('ready', '');
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback && super.disconnectedCallback();
+    this.removeAttribute('ready');
+  }
+
+  setOpen() {
```
sghoweri: I'm personally a fan of using some of the more common UI API methods (`open()`, `close()`, etc) when possible. That's not a deal breaker here -- just something to keep in mind so our component APIs (especially web component APIs) are pretty consistent across the board.

packages/components/bolt-popover/src/popover.scss
```diff
@@ -0,0 +1,236 @@
...
+  @include bolt-padding(0);
+  @include bolt-font-family(body);
+  @include bolt-font-size(xsmall);
+  @include bolt-font-weight(regular);
+  visibility: hidden; // display 'none' interferes with Popper.js calculating tooltip size
+  opacity: 0;
+  position: absolute;
+  transform: translate3d(0, #{bolt-spacing(xxsmall)}, 0);
+  z-index: bolt-z-index(popover);
+  min-width: 200px;
```
sghoweri: What's up with this magic number 😉 ? Toss in a comment!

packages/components/bolt-popover/src/popover.scss
```diff
@@ -0,0 +1,236 @@
...
+  overflow: hidden;
+  color: bolt-theme(text);
+  border-radius: bolt-border-radius(small);
+  border-width: $bolt-popover-border-width;
+  border-style: $bolt-popover-border-style;
+  border-color: $bolt-popover-border-color;
+  box-shadow: bolt-shadow('level-20');
+  background-color: bolt-theme(background);
+  transition: opacity $bolt-transition-timing * 2
+    cubic-bezier(0.25, 0.8, 0.25, 1);
```
sghoweri: todo: transition timing consistency / design tokens...

packages/components/bolt-popover/src/popover.scss
```diff
@@ -0,0 +1,236 @@
...
+    // right: 0;
+    // bottom: 100%;
+    // transform: translate3d(0, -#{bolt-spacing(xxsmall)}, 0);
+  }
+
+  @at-root .c-bolt-popover--top & {
+    @include bolt-popover-bubble-tip-point-to(null);
+    // bottom: 100%;
+    // left: 50%;
+    // transform: translate3d(-50%, -#{bolt-spacing(xxsmall)}, 0);
```
sghoweri: @danielamorse or @mikemai2awesome make sure to clean this up post `v2.15.0`

sghoweri: Same goes with all the other commented out code below 👇

packages/core-v3.x/utils/is-focusable.js
```diff
@@ -0,0 +1,63 @@
+/**
```
sghoweri: 💯

sghoweri:
PR was released with v2.15.0

sghoweri:
PR was released with v2.15.0

packages/components/bolt-popover/src/popover.js
danielamorse: Agreed, I can also add `open` and `close`. This method decides whether it should or should not open, so it's a little different. The naming could be clearer.

PR #1648, opened 2019-12-31
Feat/bds 1937 add compatibility mode

remydenton:
## Jira

http://vjira2:8080/browse/BDS-1937

## Summary

Adds a compatibility flag that determines whether a webpack build is necessary when esModules is enabled.

## Details

If both `prod` is false and `esmodules` is true, by default local builds will only get the "modern" webpack bundle (as defined in create-webpack-config.js).  This dramatically improves build time while allowing local development for most situations. 

If, however, you want to test or debug the legacy build while working locally (the most obvious use case is if you want to troubleshoot IE 11), this PR allows you to do so by running `yarn run start:compat` (under the hood, `yarn bolt start --compat`).

## How to test

On this feature branch:
- Run `yarn run start:compat`.  Confirm that you see two webpack build progress indicators in the browser while starting (and that the docs site works in IE 11)
- Run `yarn run start`.  Confirm that only one webpack build runs (and that the docs site does not works in IE 11)


remydenton:
Hmm, looks like part of the build not passing is that the changes in [this commit](https://github.com/boltdesignsystem/bolt/pull/1648/commits/db1d8b8) have unintentionally modified the logic.

I assumed that [these changes to cli.js](https://github.com/boltdesignsystem/bolt/pull/1648/commits/db1d8b8553f00fee1ea61c1af29d46656ac20633#diff-777dcf159c0d2151d846c1a1ef32c7c0) would mean that if the `prod` flag is set, the compat flag is also set.  However, when I run `yarn jest` locally, it build both new and legacy builds before this commit (meaning config.prod is true), but only the new build after this commit (meaning config.compat is false).

Needs work :/

remydenton:
@adamszalapski, good suggestion about adding `start:compat` to the root package.json, I've made that update.

I did try testing in IE 11 again with `yarn run start:compat` and it seemed to work fine.  I don't know why it would be working for you-- what do you see when you do that?  And you're saying you did check that you have both a `bolt-global.js` and a `bolt-global.modern.js` file in the `www/build` directory?

remydenton:
Re: the failing jest test, it turns out that the [changes I made to the cli in updateConfig()](https://github.com/boltdesignsystem/bolt/pull/1648/commits/db1d8b8553f00fee1ea61c1af29d46656ac20633#diff-777dcf159c0d2151d846c1a1ef32c7c0) don't apply when running jest commands.  That results in `config.compat` being `false` even though `config.prod` is `true` when running jest tests.

It looks like this same problem applies to other config that's conditionally set in `updateConfig()` (for example, `config.i18n`)-- none of it applies when running tests.  @sghoweri, any thoughts about the best way to handle this?

sghoweri:
PR was released with v2.16.0

PR #1674, opened 2020-01-07
Fix/refresh two char layout on child edit

glassdimly:
## Jira

https://app.asana.com/0/1126340469288208/1154819220959437/f

## Summary

Add "Layout Refresh" button to Editor; refresh slots after content is added to slots.

## Details

This PR has two changes. My first approach fixed a related problem, but did not address the slot refresh problem.
1. Adds a "Layout Refresh" button. This button currently only recalculates the layout for the Two-Character Layout. However, more could be added here if other components get out of whack.
2. Refreshes the `slot` content after content is added. The bug this fixes is that slot content wasn't being rendered even though HTML was present. This is because while `this.slot('new_slot')` was populated, `this.slots['new_slot']` was not. To fix this, I call `this.setupSlots()` and `this.triggerUpdate()`, but from the editor context.

## How to test

Two ways:
1. Modify the text on a Micro Journey, making it longer. See that Characters are not aligned. Now hit "Refresh Content" button and see that now they are aligned.
image
2. Attempt to add content to a new slot on Character that _does not_ have any content in that slot (and never did). This content will now appear. Previously, it did not work.


remydenton:
Closing in favor of rebased hotfix PR: #1677

PR #1675, opened 2020-01-07
Table of content | Academy | New component to navigate within an article

mikemai2awesome:
## Jira

http://vjira2:8080/browse/BDS-1971

## Summary

Introduces a new component called Table of Content: a navigation list, usually found on an article page, of its section titles.

## Details

1. Each item is set up as a link, and can be set as the current item by setting the `active` prop.
2. Fully built out the JS and Twig.
3. Documented all props and existing use cases (Academy Mission Details page).

### Still to do:
- [ ] Implement smooth scrolling with the `active` prop.

## How to test

Pull down the branch and run locally. Review the code and the docs.


packages/components/bolt-toc/src/_toc-item.js
```diff
@@ -0,0 +1,65 @@
...
+    return {
+      url: String,
+      active: {
+        type: Boolean,
+        reflect: true,
+      },
+      role: {
+        type: String,
+        reflect: true,
+      },
```
sghoweri: +1 nice use of the `reflect` config option!

packages/components/bolt-toc/src/toc.js
```diff
@@ -0,0 +1,93 @@
...
+
+    // Remove all children not in the "keep" array
+    Array.from(this.children)
+      .filter(item => !this.nodesToKeep.includes(item))
+      .forEach(node => {
+        node.parentElement.removeChild(node);
+      });
+
+    this.ssrPrepped = true;
+  }
```
sghoweri: I 2nd this todo. At a minimum, we should try to get this packaged up as a decorator within the `@bolt/element` package.

packages/components/bolt-toc/src/toc.js
```diff
@@ -0,0 +1,93 @@
...
+  static get properties() {
+    return {
+      header: String,
+      uuid: String,
+    };
+  }
+
+  constructor() {
+    super();
+    this.uuid = this.uuid || Math.floor(10000 + Math.random() * 90000);
```
sghoweri: Not a blocker (more just an aside for later) -- we should think about taking our different UUID use cases within our components (Tabs, Accordions, Tooptip / Popover, etc) and consolidating / switching them over to use a super tiny utility library.

mikemai2awesome: Tell me a new method and I'll use it. This is the current way I copied from other components.

mikemai2awesome:
@sghoweri 

1. Mistake. Updated to `text`.
2. Table of Contents is for this one specific thing. It shouldn't be used freely willy (we see this happening with the `navbar` already). Its primary function is to navigate within a particular page, it shouldn't be used to navigate in between pages, that function should be reserved for `nav`. Also, the semantic/accessible markup is different from `toc` to `nav`. BUT, we can build a more generic component that can be used to power `toc`, `nav`, and possibly `menu` in the future. Having these separate components wouldn't stop that.
3. Updated.
4. Yes. @danielamorse should add that to the JS task.
5. I disagree, Gary. I disagree. This is actually very close to the comps. If the indicator is full height, it would be bleeding into the header background which is no bueno. Overall, everything should be pretty darn close other than the colors.
6. D'oh! I will update the `menu`.

packages/components/bolt-toc/src/toc.js
danielamorse: @mikemai2awesome I'll take this as a follow-up task and cc you.

packages/components/bolt-toc/src/toc.js
danielamorse: @sghoweri I'll make a ticket 👍

sghoweri: 👍

sghoweri:
PR was released with v2.17.0

PR #1678, opened 2020-01-08
Chore/schema cleanup

glassdimly:
## Jira

Schema cleanup (no ticket)

## Summary

Fix typo or consistency errors I found when surveying schema files.

## Details

 -  Remove `additionalProperties: true` from `logo.schema.yml`. @remydenton  says this is the default.
 - `pagination.schema.yml` has `total` and `current` pages as `number`, but that would include floating point. Moved to `integer`. https://json-schema.org/understanding-json-schema/reference/numeric.html#number I can't find `number` in `YAML` spec: https://yaml.org/spec/1.2/spec.html
 -  Fix`share.schema.yml` properties `inline` and `copyToClipboard` titles set to `DPRECATED` rather than `DEPRECATED`.
 - Anal retentive fix.`tabs.schema.yml` does not quote `auto` while `on` and `off` are quoted. I unquoted `on` and `off`

### DIFF
--- a/packages/components/bolt-logo/logo.schema.yml
+++ b/packages/components/bolt-logo/logo.schema.yml
@@ -6,4 +6,3 @@ properties:
   invert:
     type: boolean
     description: Set to true to invert the logo colors.
-additionalProperties: true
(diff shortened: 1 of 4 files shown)


sghoweri:
PR was released with v2.16.0

PR #1681, opened 2020-01-13
feat: Add suggested bolt.rc file for external builds

remydenton:
## Jira

http://vjira2:8080/browse/BDS-1874

## Summary

Adds suggested .boltrc for use in external Bolt builds

## Details

Although http://vjira2:8080/browse/BDS-1874 mentions a starter package.json, since that task was created that part has become unnecessary (thanks @christophersmith262).  I believe that the automated test mentioned in that task also isn't really relevant anymore since the package.json integrity is the main thing it would test.  The docs site inherits from this rc file, so the main thing that would go wrong is if someone adds a component to the wrong file (i.e. the one for the docs site and not the starter).  Not sure how we'd write a test for that 🤔 .

I'm open to suggestions on file naming-- I ended up following the suggestions in the ticket since it seemed most simple, especially since we don't need a package.json file.

## How to test
- Confirm no regressions in the docs site
- Confirm that all components listed in `.boltrc.suggested.js` are relevant to external builds, and that all those components included in the rc file for recent bolt release builds are present.


remydenton:
Thanks @sghoweri, I'll start with #2 first.  We're actually moving toward doing the build for the Drupal end by first pulling down the Bolt repo itself, which means we don't need a separate package.json file to specify dependencies (because they're all guaranteed to be present already-- we're cutting out the npm middle man, so to speak).  All we need is an rc file to tell us which components we should actually include in the build.

Back to #1, since we don't need a package.json file for the dependencies, it seemed unnecessary to have one at all.  All that matters versioning-wise is that the starter rc file matches the other code in any given git commit.

Finally, I wanted to stay true to the specs in the JIRA ticket.

That said, I think it will still work for our purposes if this is in its own full package in a different directory (albeit it slightly clunkier).  @christophersmith262, shout if you disagree.  Otherwise, I'll update this PR.

christophersmith262:
As long as the file location stays consistent from release to release I don't care about where it is in the repository.

remydenton:
Ready for another pass @sghoweri.  Please confirm that I got the new packages from master into the right places (e.g. toolbar is internal only, right?).  I also decided against adding component dependencies to the new `package.json` since they're unnecessary for our purposes and strictly speaking there aren't dependencies.  We could potentially add them in the future.

docs-site/.boltrc.js
```diff
@@ -80,77 +82,27 @@ const config = {
...
-      '@bolt/components-link',
-      '@bolt/components-list',
-      '@bolt/components-menu',
-      '@bolt/components-modal',
-      '@bolt/components-nav-indicator',
-      '@bolt/components-nav-priority',
-      '@bolt/components-navbar',
-      '@bolt/components-navlink',
-      '@bolt/components-logo',
+      '@bolt/blueprints',
```
sghoweri: Not technically an internal package (but I could see why this should remain here for now)

docs-site/.boltrc.js
```diff
@@ -80,77 +82,27 @@ const config = {
...
-      '@bolt/components-tabs',
-      '@bolt/components-teaser',
-      '@bolt/components-text',
-      '@bolt/components-tooltip',
-      '@bolt/components-trigger',
-      '@bolt/components-ul',
-      '@bolt/components-ol',
-      '@bolt/components-video',
-      '@bolt/components-grid',
+      '@bolt/components-toolbar',
```
sghoweri: Same with this one although this component is much more like a normally published deliverable compared to `@bolt/blueprints`, albeit an early sneak.

docs-site/.boltrc.js
```diff
@@ -80,77 +82,27 @@ const config = {
...
+      '@bolt/docs-search',
+      // '@bolt/schema-form', // Component Explorer being temporarily disabled until we've migrated our Twig Rendering Service to Now.sh v2
+      '@bolt/shadow-toggle',
+      '@bolt/theme-switcher',
+
+      // Components that are excluded from the base release build.
       '@bolt/micro-journeys',
```
sghoweri: Why is editor included but not micro-journeys? Is this an extra package that's installed + added to the `.boltrc` config locally downstream?

docs-site/.boltrc.js
```diff
@@ -1,8 +1,10 @@
 const path = require('path');
 const resolve = require('resolve');
 const argv = require('yargs').argv;
+const deepmerge = require('deepmerge');
+const baseConfig = require('@bolt/foundation/.boltrc.js');
```
sghoweri: This is one of two things that I feel particularly strong about updating before we release.

Specifically, I feel like this should really get pulled in via adding a `main` field on the foundation package.json (`"main": ".boltrc.js",`) and then updating this require statement:
```
const baseConfig = require('@bolt/foundation');
```

docs-site/src/pages/docs/50-guides/15-building-a-new-component.md
```diff
@@ -85,7 +85,7 @@ polyfillLoader.then((res) => {
 ]
 ```
 
-**<u>Step 7</u>:** Add component to `package.json` AND `.boltrc.js` within the Pattern Lab dir `/docs-site/`
+**<u>Step 7</u>:** Add component to `package.json` within the Pattern Lab dir `/docs-site/` AND the `.boltrc.js` within the `@bolt/foundation` package.
```
sghoweri: @remydenton this out of date line in the docs reminded me about the 2nd thing I think we should update here before publishing: updating our Yeoman generator script to auto-update the foundation package's `.boltrc.js` config (vs the docs site config): https://github.com/boltdesignsystem/bolt/blob/master/packages/generators/yeoman-create-component/generators/component/update-boltrc.js#L7

docs-site/.boltrc.js
sghoweri: @remydenton Actually on second thought, I'm now a little iffy on this...

I want to make sure we have a path forward to deliver pre-compiled CSS and JS Bolt code so I'd actually recommend holding off changing this until we have a plan for that!

docs-site/src/pages/docs/50-guides/15-building-a-new-component.md
sghoweri: @remydenton I'm gonna go ahead and make this update before merging this down.

sghoweri:
PR was released with v2.17.0

PR #1685, opened 2020-01-15
feat(editor): remove click handlers on element select; add function d…

glassdimly:
…ocs; add visual test

## Jira

https://app.asana.com/0/1126340469288208/1152302051041273/f

## Summary

Remove click handlers on element select in Editor.

## Details

Previously, when clicking a `bolt-cta` element in the editor that had an `on-click-target`, that event would fire and prevent editing.

Now, the first time a `bolt-cta` element is clicked, it fires the modal. After the modal is closed, the element can be edited normally.

I achieved the removal by using the object `clickHandler` property which is set in `declarativeClickHandler` on Editor actions `select` and `deselect`.

## How to test

* pattern-lab/?p=experiments-editor-modal
* Edit `bolt-cta` text


glassdimly:
@remydenton,

Yes, that's correct. The video pops up one time You must click it a second time. From the above:

> Now, the first time a bolt-cta element is clicked, it fires the modal. After the modal is closed, the element can be edited normally.

Let me know if this is acceptable.

glassdimly:
So the problem is that there are two clickhandlers. One that selects the object, and the other that fires the modal event. I am listening for the selection click, then I am removing the modal event and adding it back after selection.

I could probably remove all clickhandlers on all elements on open and re-add them on save, but this feels better. It also demonstrates the click event while editing on the first click, at the same time allowing the element to be edited on second click. So even while it requires a second click, it feels usable.

packages/experimental/editor/src/editor.js
```diff
@@ -396,16 +396,57 @@ export function enableEditor({ space, uiWrapper, config }) {
...
+  });
+
+  /**
+   * Re-add removed clickHandlers to selected components
+   *
+   * @param {grapesjs.Component} model
+   * @return {void}
+   */
+  editor.on('component:deselected', (/** @type {grapesjs.Component} */ model) => {
+    console.log(model.view.el);
```
remydenton: Oops, we should have remove debugging code.

glassdimly: Shoot. Here ya go: https://github.com/boltdesignsystem/bolt/pull/1710

sghoweri:
PR was released with v2.16.0

PR #1691, opened 2020-01-16
feat(editor): add prop schema descriptions underneath traits

glassdimly:
## Jira

https://app.asana.com/0/1126340469288208/1157449394066073/f

## Summary

Add prop schema descriptions underneath traits.

## Details
I chose the approach I did (appending our own div to existing traits at runtime) because otherwise we would have to create custom traits from scratch, and traits are extendible, so we'd have had to create them all from scratch.

Before:
image

After:
image

## How to test

Edit components. See that the descriptions from their prop schema are added underneath the traits in editor. Make sure editor still works.

### DIFF
--- a/packages/experimental/editor/src/editor.js
+++ b/packages/experimental/editor/src/editor.js
@@ -426,6 +426,29 @@ export function enableEditor({ space, uiWrapper, config }) {
     render(content, editorSlots.slotControls);
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


sghoweri:
PR was released with v2.15.0

sghoweri:
PR was released with v2.15.0

PR #1701, opened 2020-01-22
Allow button group to render non-composed items.

charginghawk:
## Jira

http://vjira2:8080/browse/BDS-1978

## Summary

Updates button group to accept items other than composer button data arrays.

## Details

If the item array has "text" (the only required field), it uses the old method. Otherwise, it just renders the item.

## How to test

Make sure button groups continue to work.


margoromo:
Thank you @charginghawk !

margoromo:
@remydenton @relimck @sghoweri @mikemai2awesome - please hold on review, there needs to be some discussion internally before we move on this.

mikemai2awesome:
I don't think this is consistent with how we handle custom content in other components.

sghoweri:
PR was released with v2.18.0

PR #1712, opened 2020-01-27
TOC | JS | Add smooth scroll and scroll-spy behavior

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1975

## Summary

Add smooth scroll and scroll-spy behavior to Table of Contents component.

## Details

- Uses an adapted version of [React Waypoint](https://github.com/civiccc/react-waypoint) for scroll-spy behavior (i.e. update active item as you scroll). For now, Waypoint-specific code lives with the component in `./waypoint`. This could easily be refactored into Core and used with Nav Priority later on.
- Updates component to use `getUniqueId` helper
- Makes sure only one active item at a time and initial active item is taken into account.
- Fix SSR hydration prep issues in IE11 by switching from `<ssr-keep>` to a new method, avoids the race conditions caused by `<ssr-keep>` loading late in IE11.

## How to test

- Review code changes
- Test TOC demos, especially "Use Case Sticky" for scroll spy behavior

### DIFF
--- a/packages/components/bolt-toc/src/_toc-item.js
+++ b/packages/components/bolt-toc/src/_toc-item.js
@@ -66,8 +66,8 @@ class BoltTocItem extends withContext(BoltElement) {
   }
 
   contextChangedCallback(name, oldValue, value) {
-    if (name === 'activeItem') {
-      if (this === value) {
+    if (name === 'activeItem' && value) {
+      if (value === this) {
         this.active = true;
       } else {
         this.active = false;
--- a/packages/components/bolt-toc/src/_toc-item.js
+++ b/packages/components/bolt-toc/src/_toc-item.js
@@ -57,6 +57,7 @@ class BoltTocItem extends withContext(BoltElement) {
     super.connectedCallback && super.connectedCallback();
 
     if (this.url && this.url.indexOf('#') === 0) {
+      // todo: update `this.target` when url prop changes
       this.target = document.querySelector(this.url);
     }
   }
--- a/packages/components/bolt-toc/src/_toc-item.js
+++ b/packages/components/bolt-toc/src/_toc-item.js
@@ -2,12 +2,17 @@ import {
   customElement,
   BoltElement,
   html,
-  styleMap,
   ifDefined,
   unsafeCSS,
 } from '@bolt/element';
-import classNames from 'classnames/dedupe';
+import classNames from 'classnames/bind';
 import { withContext } from 'wc-context';
+import {
+  smoothScroll,
+  scrollOptions,
+  getScrollTarget,
+} from '@bolt/components-smooth-scroll';
+
 import tocItemStyles from './_toc-item.scss';
 import schema from '../toc.schema';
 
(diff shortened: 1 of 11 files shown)


packages/components/bolt-toc/src/_toc-item.js
```diff
@@ -44,6 +49,60 @@ class BoltTocItem extends withContext(BoltElement) {
...
+    super.connectedCallback && super.connectedCallback();
+
+    if (this.url && this.url.indexOf('#') === 0) {
+      this.target = document.querySelector(this.url);
+    }
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback && super.disconnectedCallback();
+  }
```
sghoweri: @danielamorse is this needed or just force of habit?

packages/components/bolt-toc/src/toc.js
```diff
@@ -29,43 +43,230 @@ class BoltToc extends withContext(BoltElement) {
...
+    if (event.type === 'scrollStart') {
+      this.scrolling = true;
+    } else if (event.type === 'scrollStop' || event.type === 'scrollCancel') {
+      this.scrolling = false;
+    }
+  }
 
-    this.ssrPrepped = true;
+  onActivate(event) {
+    if (event?.detail?.activeItem) {
```
sghoweri: +1

packages/components/bolt-toc/src/toc.js
```diff
@@ -29,43 +43,230 @@ class BoltToc extends withContext(BoltElement) {
...
+    );
+    document.body.appendChild(marker);
+    BoltToc.debuggerAdded = true;
+  }
+
+  get items() {
+    return this.querySelectorAll('bolt-toc-item');
+  }
+
+  get itemData() {
```
sghoweri: +1 Good use of a getter

packages/components/bolt-toc/src/toc.js
```diff
@@ -29,43 +43,230 @@ class BoltToc extends withContext(BoltElement) {
...
+    // Listen for smooth-scroll events
+    this.logScrollEvent = this.logScrollEvent.bind(this);
+    document.addEventListener('scrollStart', this.logScrollEvent, false);
+    document.addEventListener('scrollStop', this.logScrollEvent, false);
+    document.addEventListener('scrollCancel', this.logScrollEvent, false);
+
+    // Dislay waypoint boundaries on screen if debug is `true` (only ever IN_DEV!)
+    if (IS_DEV && BoltToc.debug && !BoltToc.debuggerAdded) {
+      this.addWaypointDebugger();
+    }
```
sghoweri: Oooooo - gotcha. I was about to ask about the debugger until I saw this... not something to solve right this second but we should start to think about how debug modes get handled across components.

packages/components/bolt-toc/src/_toc-item.js
```diff
@@ -44,6 +49,60 @@ class BoltTocItem extends withContext(BoltElement) {
...
   }
 
+  static get observedContexts() {
+    return ['activeItem'];
+  }
+
+  connectedCallback() {
+    super.connectedCallback && super.connectedCallback();
+
+    if (this.url && this.url.indexOf('#') === 0) {
```
sghoweri: @danielamorse todo reminder (post MLP): we need to make sure `this.target` is updated whenever the URL prop changes, otherwise clicking on this will continue to scroll to the original URL (vs the new updated URL)

packages/components/bolt-toc/src/_toc-item.js
danielamorse: Nice catch @sghoweri, left over method from an earlier version of the component. Now removed.

packages/components/bolt-toc/src/_toc-item.js
danielamorse: 👍 added a todo

sghoweri:
PR was released with v2.17.0

PR #1714, opened 2020-01-28
Hotfix/wwwd 4533 web component icon background match twig

remydenton:
## Jira

http://vjira2:8080/browse/WWWD-4533

## Summary

Removes background on web component icons when background="none"

## Details

This fixes a regression (see http://vjira2:8080/browse/WWWD-4533) where a background="none" prop actually resulted in a background being added.  

### Confirm the bug
- On master, add this icon in pattern lab (as a web component): `<bolt-icon name="chevron-right" background="none"></bolt-icon>`
- Confirm that a background is added

### Confirm the fix
- On this branch, confirm that the same behavior results in no background being added.


sghoweri:
PR was released with v2.16.2

PR #1717, opened 2020-01-28
Fix Step removal event not firing due to removal of Animate parent

remydenton:
Version of #1716 rebased for hotfix.


sghoweri:
PR was released with v2.16.3

PR #1720, opened 2020-01-29
Fix/link whitespace

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1980

## Summary

Fix regression in Link component where extra whitespace is added around inline links.

## Details

Line breaks were added to the Link JS file that resulted in extra whitespace. Our test didn't catch it because, though the tolerance is relatively low, the visual difference was still too small to fail the test. The new test makes the font-size huge so extra whitespace will fail the test (verified).

## How to test

- Review the code changes, updated tests
- Verify all tests are passing
- Verify extra whitespace around link is removed: https://fix-link-whitespace.boltdesignsystem.com/pattern-lab/?p=components-link-display-variations


sghoweri:
PR was released with v2.17.0

PR #1728, opened 2020-01-31
fix: remove padding on last list item

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1989

## Summary

Remove right padding on last list item.

## Details

CSS that should be removing this padding is not scoped properly. This fixes the scoping.

## How to test

- Review code changes.
- Check Inline List demo and verify last item has no right padding.


sghoweri:
PR was released with v2.17.0

PR #1729, opened 2020-02-03
Fix block() and include() autoescaping

remydenton:
## Jira

http://vjira2:8080/browse/BDS-1988

## Summary

Fix twig `block()` and `include()` function usage to work with autoescaping

## Details
This PR has two parts:
1. Replace the twig `include()` function in the blueprints package with a macro that prints the contents of the include in order to play nice with autoescaping (credit goes to @sghoweri, thanks!)
2. Update places where the twig `block()` syntax is used to ensure that the block is always printed before setting its contents to a variable (also to play nice with autoescaping and remove the need for the `|raw` filter).  This can take several forms:
    - Most simply, ensuring that where the block contents is printed, the ` {{ block('block-name') }}` syntax (and not a printed variable) is used.
    - In more complex cases-- especially when embeds result in multiple layers of blocks, potentially with conflicting names-- setting a variable representing the block and its contents using the following syntax:
```
{% if block("body") %}
  {% set body_block %}
    {{ block("body") }}
  {% endset %}
{% endif %}
```

## How to test
- Confirm that replacement card, toolbar, and all blueprint patterns can be printed with autoescaping enabled (in particular, in Drupal) without raw markup displaying.  Unfortunately, Drupal lab needs some updates to be functional, which I omitted from this PR for clarity (but will work on in separate PRs).
- Confirm no regressions in pattern lab.  In particular, confirm that when a block is either not present or empty, any conditional wrappers do not print.  For example, both the following should result in cards that do not have a `<bolt-card-replacement-media>` element within them

```
{% embed "@bolt-components-card-replacement/card-replacement.twig"  only %}
  {% block body %}
    hello world
  {% endblock %}
{% endembed %}
```
```
{% embed "@bolt-components-card-replacement/card-replacement.twig" only %}
  {% block media %}{% endblock %}

  {% block body %}
    hello world
  {% endblock %}
{% endembed %}
```

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/03-blueprints/01-atoms/forms/_checkbox.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/03-blueprints/01-atoms/forms/_checkbox.twig
@@ -1,7 +1,9 @@
 {% extends "@bolt-components-form/form-element.twig" %}
 
+{% import "@bolt-blueprints/macros.twig" as macros %}
+
 {% set label %}
-  {% include("@bolt-components-form/form-label.twig") with {
+  {% macros.include("@bolt-components-form/form-label.twig") with {
     "title": labelTitle,
     "displayType": "inline-checkbox",
     "attributes": {
@@ -11,7 +13,7 @@
 {% endset %}
 
 {% set children %}
-  {% include("@bolt-components-form/form-input.twig") with {
+  {% macros.include("@bolt-components-form/form-input.twig") with {
     "attributes": {
       "type": "checkbox",
       "id": elementId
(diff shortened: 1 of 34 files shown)


remydenton:
Shoot, I was indeed testing in Drupal lab, but it looks like I may have broken toolbar right before I pushed.  Investigating now.

sghoweri:
> Shoot, I was indeed testing in Drupal lab, but it looks like I may have broken toolbar right before I pushed. Investigating now.

@remydenton I put in https://github.com/boltdesignsystem/bolt/pull/1730 based on local testing however I'd totally be open to alternatives to get us over the finish line!

remydenton:
@sghoweri, thanks for the assistance.  

Re: toolbar, it looks like adding the `|default()` filter makes the block contents get treated as an untrusted string.  Can you have a look at my updates to the toolbar logic to see if that part of the problem is fixed?

Regarding the raw filters on typography and li, those are unfortunately pretty critical ones in terms of security (they're used everywhere for all sorts of things, including likely user content).  I think the answer here is that we'll just have to update the content in the blueprint pages to make sure content is being printed before it's stored to a variable.  I'll see how far I can get on that now.

sghoweri:
PR was released with v2.17.0

PR #1737, opened 2020-02-06
Fix Button `disabled` prop

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1993

## Summary

Fixes button `disabled` prop, adds test coverage.

## Details

When we switched Button over to LitElement we did not add `reflect: true` to the `disabled` prop. This is required for boolean HTML attributes to be reflected as internal props. Otherwise, they will return an empty string or `undefined`. This PR adds the `reflect` option and two tests to prevent future regressions.

As a general rule, when using LitElement, boolean schema props must be configured with `reflect: true`. This allows external changes to the attribute to be reflected in the internal prop and vice versa.

If the prop is internal-only (see [dotted](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-tooltip/src/tooltip.js#L20), [expanded](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-band/src/band.js#L7)) there's no need to use `reflect`. The reason we still add these as internal props (even though they're not in the schema) is because we can leverage LitElement's built-in observers to do something when that prop value changes or to automatically trigger a re-render.

## How to test

- Review files changed.
- Check `/pattern-lab/?p=components-button-style-variations` are verify disabled buttons have the proper semi-transparent style.


remydenton:
What does "boolean props like `disabled`" mean more specifically?  There are plenty of [boolean](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-band/src/band.js#L7) [props](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-tooltip/src/tooltip.js#L20) that aren't reflected.

danielamorse:
> What does "boolean props like `disabled`" mean more specifically? There are plenty of [boolean](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-band/src/band.js#L7) [props](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-tooltip/src/tooltip.js#L20) that aren't reflected.

@remydenton good call. That was a pretty vague. I updated "Details" with more info. Does that clear things up for you?

remydenton:
As discussed on Slack, we actually need to update all LitElement properties from this syntax:
```
static get properties() {
    return {
        url: String,
        disabled: Boolean
    }
}
```
to this one:
```
static get properties() {
    return {
        url: { type: String },
        disabled: { type: Boolean }
    }
}
```
Based on the [LitElement](https://lit-element.polymer-project.org/guide/properties) docs, the first syntax isn't supported.

danielamorse:
@remydenton updated to use proper syntax ([see commit](https://github.com/boltdesignsystem/bolt/pull/1737/commits/baac30c4840d82ea8c59fb2fda8c81dab2f152a6)).

I also verified that once we're using proper syntax `reflect` is not required to observe `disabled` attribute.

sghoweri:
> As discussed on Slack, we actually need to update all LitElement properties from this syntax:
> 
> ```
> static get properties() {
>     return {
>         url: String,
>         disabled: Boolean
>     }
> }
> ```
> 
> to this one:
> 
> ```
> static get properties() {
>     return {
>         url: { type: String },
>         disabled: { type: Boolean }
>     }
> }
> ```
> 
> Based on the [LitElement](https://lit-element.polymer-project.org/guide/properties) docs, the first syntax isn't supported.

Per the comment I just left, @remydenton @danielamorse can we chat about this?

Unless something changed, this shorthand syntax should still work for simply props (and I could swear used to be in their docs). 

Other LitElement components [out in the wild](https://github.com/ing-bank/lion/blob/bbc79af066167141321eccdc051facd588af4d09/packages/form-system/stories/helper-wc/h-output.js#L8) also use the same syntax so I'd very much prefer for us to understand what syntax is actually still supported before globally changing code that might not need to get changed...

remydenton:
@sghoweri, do you have a reference that mentions the shorthand syntax?  I was going to report a bug about observed attributes not working properly with the shorthand syntax, but then I couldn't find any mention of it in the LitElement docs, so @danielamorse and I came the conclusion it was never supported (and should therefore be updated for all our other components as well).

sghoweri:
> @sghoweri, do you have a reference that mentions the shorthand syntax? I was going to report a bug about observed attributes not working properly with the shorthand syntax, but then I couldn't find any mention of it in the LitElement docs, so @danielamorse and I came the conclusion it was never supported (and should therefore be updated for all our other components as well).

@remydenton lemme check - hang tight!

remydenton:
Thanks @sghoweri.  FYI, the bug is that this "converter" code won't be run if you don't have a `type` key in your prop definition (which you never will if you're using the shorthand syntax).  Makes for weird behavior with non-string types, especially booleans.

https://github.com/Polymer/lit-element/blob/master/src/lib/updating-element.ts#L138-L166

danielamorse:
Discussed with @sghoweri and we're good to merge this as is. We will handle a global syntax update in a future PR.

Nice job diving deep into this one @remydenton. I'm really glad we got to the bottom of it.

sghoweri:
PR was released with v2.18.0

PR #1738, opened 2020-02-07
fix(editor): rearrange step w content

EvanLovely:
## Tasks

- Asana: [Move up/down buttons breaks when applied to pathway steps](https://app.asana.com/0/1126340469288208/1160381693199881/f)
- JIRA: [WWWD-4544](http://vjira2:8080/browse/WWWD-4544)

## Summary

Fixes Editor re-arrangement of Micro Journey Steps

## Details

(Explain the changes in enough detail fo reviewers to understand.  Raise any questions for reviewers to consider.)

## How to test

- Go to https://boltdesignsystem.com/pattern-lab/?p=experiments-micro-journeys and click 'Edit'
- Select one of the `bolt-interactive-step` elements in the layers palette
- Click the `move up` or `move down` button in the editor

**Resolved**: After doing this, the contents of the step you moved will be gone (and won't come back after saving).

# WIP Remaining

- [x] Fix core bug involving Steps moving up & down & retaining content
- [x] Editor UI feedback for first/last items
- [x] Ensure pathway re-arranging is smooth too

PR #1742, opened 2020-02-11
Fix micro journey up and down buttons

remydenton:
Rebased version of #1738.


sghoweri:
PR was released with v2.17.1

PR #1743, opened 2020-02-11
Fix broken PL pages

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-2009

## Summary

Rename Blueprint templates to avoid conflict with other pattern lab templates.

## Details

With the addition of `_default-page-template.twig` to the Blueprints folder we now have two files with the same name in the `@pl` namespace. The one in Blueprints is winning out and breaking pages outside of Blueprints which used to point to another copy of this file, the one in `04-pages` ([see file](https://github.com/boltdesignsystem/bolt/blob/master/docs-site/src/pages/pattern-lab/_patterns/04-pages/_default-page-template.twig)).

I considered excluding Blueprints from the @pl name space ([here](https://github.com/boltdesignsystem/bolt/blob/master/docs-site/.boltrc.js#L37-L47)) but that seemed like it would complicate what is now a pretty simple config. So, the simpler solution was simply rename the ones in Blueprints.

No downstream side-effects expected, as Academy team is not referencing these layouts directly.

## How to test

- Tests are passing
- Review files changed. Ok with the new filenames?
- Pages and Blueprints are working as expected on feature branch, for example:
  - https://fix-broken-pl-pages.boltdesignsystem.com/pattern-lab/?p=blueprints-dashboard
  - https://fix-broken-pl-pages.boltdesignsystem.com/pattern-lab/?p=blueprints-homepage
  - https://fix-broken-pl-pages.boltdesignsystem.com/pattern-lab/?p=blueprints-search-results-page


remydenton:
+1 for a nice simple solution!

sghoweri:
PR was released with v2.18.0

PR #1748, opened 2020-02-19
Fix wrapping links in Safari

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-2010

## Summary

Fixes links in a list that are wrapping when they should not be.

### Underlying issue
There is a bug in Safari with our `bolt-list` component that occurs when an inline list is placed inside certain containers. So far we've observed it happening when placed inside a flex grid cell and nested inside `bolt-list`. The bug causes inline elements to wrap to a new line when they have space to fit on a single line.

It appears to be a sub-pixel rounding issue because when we pad the list `width` by 1 pixel (add `+ 1px` to [this](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-list/src/list.scss#L80-L96) calc function) it fixes itself. However, it causes a separate problem in FF where it's then 1px too wide and can cause scroll bars to appear if the list sits inside a container with overflow `auto` on it.

### Workaround
We couldn't find a fix to the root problem ([made a ticket](http://vjira2:8080/browse/BDS-2024)), so for now this is a workaround. Simply switch to a nested grid instead of using `bolt-list`. Change that and it works fine.

I also adjusted the padding around the cell below the grid to reduce overall spacing.

## How to test

- Review files changed
- See Dashboard template in Safari (and Chrome and FF) and verify the "Show all" and "Add Missions" buttons are not wrapping.
  - `/pattern-lab/patterns/03-blueprints-05-pages-dashboard-dashboard-dashboard/03-blueprints-05-pages-dashboard-dashboard-dashboard.html`

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/03-blueprints/03-organisms/collections/dashboard/_dashboard-band.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/03-blueprints/03-organisms/collections/dashboard/_dashboard-band.twig
@@ -3,7 +3,7 @@
 {% set useCarousel = useCarousel ?? true %}
 
 {% set content %}
-  {% grid "o-bolt-grid--flex o-bolt-grid--matrix" %}
+  {% grid "o-bolt-grid--flex o-bolt-grid--matrix o-bolt-grid--middle" %}
     {% if sectionTitle %}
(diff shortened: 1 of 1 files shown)


sghoweri:
PR was released with v2.18.0

PR #1752, opened 2020-02-20
fix: Correct icon color rules

remydenton:
## Jira

http://vjira2:8080/browse/BDS-2030

## Summary

Fixes icon color regression in Edge

## Details

While it seems to work in most browsers (which is to say, nothing breaks),
the line removed in this PR:

1) is unnecessary, because the same rule is repeated withing the SVGs
2) breaks in earlier versions of Edge, e.g. Edge 18.  My best guess is
that that's because `color: currentColor` is circular/redundant.
`var(--bolt-theme-icon)` also appears to work in all browsers, but why
not just remove it if it's not doing anything the SVG itself isn't also
doing?

### Reproduce the bug
- In Edge 18, go to https://v-2-18-0.boltdesignsystem.com/pattern-lab/?p=components-icon-theme-variation
- White icons will look white on a white background:
Screen Shot 2020-02-20 at 1 40 09 PM

### Confirm the fix
- In Edge 18, go to /pattern-lab/?p=components-icon-theme-variation on this branch and confirm it looks like this instead
Screen Shot 2020-02-20 at 1 38 44 PM
- Confirm no other icon regressions in modern browsers and no other differences in Edge 18.


sghoweri:
PS. I fixed your failing Jest tests 😉

sghoweri:
> Thanks @remydenton.
> 
> Tested and confirmed to be working in Chrome, Safari, Firefox, new versions of Edge, and Microsoft EdgeHTML 17.17134 / Microsoft Edge 42.17134.1098.0.
> 
> Double checking on Edge 18 and if that also checks out I'm going to merge down and get this hotfix published.

Checks out on Edge 18 as well 👍 

image

sghoweri:
PR was released with v2.18.1

PR #1760, opened 2020-02-25
Fix OL number bugs in Firefox

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-2028

## Summary

Fixes alignment and incrementing bugs on OL in Firefox.

## Details

See alignment and incrementing bugs:

image

This appears to be caused by a recent Firefox update. These issues don't exist in older versions.

Setting the `bolt-li` to `display: block` fixes the number position issue. Custom elements are `inline` by default, and somehow that threw off the relative position in Firefox.

Number incrementing was fixed by removing an extra `number-increment` rule which applied only to list items in the shadow DOM. This was causing the counter to be reset for each encapsulated item (so always `1`). Again, not sure why this just started to be an issue in Firefox, but retested and now works cross-browser.

- View [OL demo on master](https://boltdesignsystem.com/pattern-lab/patterns/02-components-ol-15-ol-nested-items/02-components-ol-15-ol-nested-items.html) to see bugs.
- View OL demo on feature branch (waiting to deploy) and verify bugs are fixed.


sghoweri:
PR was released with v2.19.0

PR #1775, opened 2020-03-09
Fix/broken video in tabs

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-2058

## Summary

Fixes broken videos in inactive Tabs.

## Details

It looks like changes to Preact caused this regression, and there was no test in place to cover this exact use case. There is now a test to prevent future regressions.

In Bolt v2.13 we upgraded Preact, which changed the way components (like Video) re-render. Something about the new way of re-rendering doesn't fully destroy and re-create the element like it used to. That's good in some ways but bad for Brightcove which expects a clean slate when it initializes or the player may not properly load.

Our old fix re-initialized the video player when a hidden video became visible. While debugging, I found that I can remove the old fix entirely and the video loads fine, even when it is initially hidden. This tells me that support for loading hidden videos has improved on the Brightcove side because that was the whole reason for the original fix.

So, I removed the old fix and added the missing test.

Also, I updated the version of Chrome used by our Nightwatch tests to match the current version we all run locally for consistency between the two environments.

## How to test

- Verify tests are passing
- View [Tabs Content](https://fix-broken-video-in-tabs.boltdesignsystem.com/pattern-lab/?p=components-tabs-content) and [Accordion Content](https://fix-broken-video-in-tabs.boltdesignsystem.com/pattern-lab/?p=components-accordion-content-variations) pages and verify video loads properly when you open that panel.
  - Note: same underlying issue affects Accordion as Tabs.

### DIFF
--- a/packages/components/bolt-tabs/__tests__/tabs.e2e.js
+++ b/packages/components/bolt-tabs/__tests__/tabs.e2e.js
@@ -73,12 +73,8 @@ module.exports = {
       .assert.elementPresent(video)
       .assert.not.visible(video)
       .execute(function(data) {
-        document
-          .querySelector('bolt-tabs')
-          .renderRoot.querySelector(
-            '.c-bolt-tabs__nav > .c-bolt-tabs__label:nth-child(4)',
-          )
-          .click();
+        // Opens video tab
+        document.querySelector('bolt-tabs').setAttribute('selected-tab', 4);
       })
       .assert.visible(video)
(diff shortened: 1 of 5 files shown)


remydenton:
Sounds like it would be best to make this a 2.19.1 since otherwise it will likely take an additional pega.com sprint before it gets pulled in.

sghoweri:
PR was released with v2.19.1

PR #1788, opened 2020-03-23
Prevent empty `href` attribute on web component version of bolt-link

adamszalapski:
## Jira

http://vjira2:8080/browse/BDS-1921

## Summary

Fix the issue and write a test to check this. 

## How to test

Run this code locally. Add `bolt-link` web component version without `url` prop or provide content without `href` attribute like:
`<bolt-link>Link without URL prop</bolt-link>` or `<bolt-link><a target="_blank">Link without href</a></bolt-link>` and check if you clikc on that link page will reload.
You can inspect if `href` is added to the content inside `bolt-link`

This is a small fix and i think that only one approve is enough to merge it.


sghoweri:
PR was released with v2.21.0

PR #1789, opened 2020-03-24
BDS-1573 action element schema

remydenton:
## Jira

http://vjira2:8080/browse/BDS-1573

## Summary

Add schema for BoltActionElement

## Details

- Create a schema to document BoltActionElement
- Uses the BoltActionElement schema in bolt-button (including removing usage of deprecated props from pl documentation)
- Update bolt-link schema to be js-based
- Uses the BoltActionElement schema in bolt-link (including removing usage of deprecated props from pl documentation)

## How to test

- Confirm that the schema docs for bolt-button and bolt-link both show the inherited props from BoltActionElement (including `href` and `url`)
- Confirm that the schema docs for bolt-link does not have a `disabled` prop (follow up ticket in http://vjira2:8080/browse/BDS-2072)
- Confirm no regressions (tests passing, etc-- this PR shouldn't modify functionality).


sghoweri:
@remydenton have we confirmed what happens with the global data store if more than one schema gets exported in that `element.schemas.js` file?

remydenton:
Thanks @danielamorse, I think that's a great addition.  I'll use your updates for link and make similar ones for button.

@sghoweri, I've got questions about the global data store, I'll follow up on Slack.

sghoweri:
> Thanks @danielamorse, I think that's a great addition. I'll use your updates for link and make similar ones for button.
> 
> @sghoweri, I've got questions about the global data store, I'll follow up on Slack.

@remydenton Ask away!

remydenton:
@danielamorse, I realized once I started that changing the order isn't really as appropriate for button since `url` are `target` are relatively less important.  You probably came to the same conclusion before I did 😄 .  I merged in your updates for link.

@sghoweri, I've confirmed that you can access the schemas from the element package like this:
`{% set schema = bolt.data.components["@bolt-element"].schema.boltActionElement %}`

If more schemas that get added to bolt element in the future, you can just change the last part of that expression to reference the one you want.

sghoweri:
PR was released with v2.21.0

PR #1799, opened 2020-03-26
fix: Rework logic for grid ordering in hero component

remydenton:
## Jira

http://vjira2:8080/browse/BDS-2079

## Summary

Fixes a bug where hero `reverseOrder` prop does not work when `imageValign` is `bottom`

## Details

I set out only to fix the original bug.  However, I found the existing twig logic for `hero` to be pretty difficult to read, in part because there were a number of logic errors.  For example:

```
{% set _isImageFirstOnMobile = (imageValign != 'bottom') or (imageValign == 'top') %}
{% set _isImageLastOnMobile = (imageValign != 'top') or (imageValign == 'bottom') %}
```

First, both of these statements are logically identical if you remove the second conditions (if something is "not bottom", you don't need to also check for whether it could be "top").
```
{% set _isImageFirstOnMobile = (imageValign != 'bottom') %}
{% set _isImageLastOnMobile = (imageValign != 'top') %}
```

Digging in further, what happens if the value of `imageValign` is `middle`?  Well, both of the above statements evaluate to `true`, so you end up with image on mobile being both first _and_ last. 🙃 

So I ended up re-working a lot of the logic here.  Hopefully it's equivalent to what was there before without the additional bugs.

### Reproduce the bug
- Check out any other branch
- Edit docs-site/src/pages/pattern-lab/_patterns/02-components/hero/20-hero--reverse-order.twig
Change `imageValign` to `bottom`
View the pattern at /pattern-lab/?p=components-hero--reverse-order
Note that the image is on the right (expected on the left).

### Confirm the fix
- Check out the branch in this PR
- Confirm that the steps above give the expected results
- Confirm that all other scenarios give expected results

### DIFF
--- a/packages/components/bolt-hero/hero.twig
+++ b/packages/components/bolt-hero/hero.twig
@@ -17,9 +17,16 @@
 {% set _imagePaddingTop = imageValign == "top" ? "none" : "large" %}
 {% set _imagePaddingBottom = this.data["image_valign"] == "bottom" ? "none" : "large" %}
 
-{% set _isImageFirstOnMobile = (imageValign != 'bottom') or (imageValign == 'top') %}
+{# Image order on desktop is always determined by the reverseOrder prop; order on mobile is always set automatically #}
 {% set _isImageFirstOnDesktop = reverseOrder == true %}
-{% set _isImageLastOnMobile = (imageValign != 'top') or (imageValign == 'bottom') %}
+{% set _isImageFirstOnMobile = imageValign != 'bottom' %}
+
+{#
+  Image and content will be printed in the order they should appear on mobile.  For desktop, leave it the same unless:
+    - An image is present, in which case check to see if mobile and desktop ordering should be different
+    - No image is present, in which case simply honor the reverseOrder prop.
+#}
+{% set _reverseGrid = image ? (_isImageFirstOnDesktop != _isImageFirstOnMobile) : reverseOrder %}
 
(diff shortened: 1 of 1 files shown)


packages/components/bolt-hero/hero.twig
```diff
@@ -54,11 +61,17 @@
   {% endset %}
 {% endif %}
 
+{#
+  Image and content are printed in the order they should appear on mobile.  For desktop, leave it the same unless:
```
sghoweri: @remydenton isn't this entire block of code (lines 64 to 69) duplicate of what's already on lines 24 to 29?

sghoweri:
@remydenton yeah, I think these couple of quirks are side-effects from the original implementation allowing for desktop and mobile order to both be independently configured before getting consolidated at the 11th hour... 🤦‍♂ 

image

sghoweri:
@remydenton reminder: make sure you add a label based on the SEMVER release type so things get automatically published to correct version (plus add the milestone release version so we can track what work is expected to go out in which release!)

packages/components/bolt-hero/hero.twig
remydenton: I was trying to figure out which was the better place to put it right before I pushed, and mistakenly left it in both 🤦‍♂.  Good catch.

sghoweri:
PR was released with v2.21.0

PR #1810, opened 2020-04-03
fix: Allow card url attributes to be empty

remydenton:
## Jira

N/A

## Summary

Allows card url attributes to be empty

## Details

null arguments to the `create_attributes()` twig function result in PHP errors in Drupal.  Based on existing Bolt conventions, you should provide an empty default argument unless you know the argument will not be null.


sghoweri:
PR was released with v2.21.0

PR #1818, opened 2020-04-08
Feature/auto-prop support

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1266

## Summary

Adds proper JSON Schema support to our custom LitElement-based BoltElement base class, and updates Menu component to demonstrate usage.

This PR lays the groundwork for adding schema validation to LitElement. Specifically, it:

1. Adds a new decorator to `@bolt/element` which uses static schema data set on a component to:
    - Auto-generate most of the component property data (ie. map schema data to LitElement properties)
    - Automatically add default component prop values based on the JSON schema
2. Updates the Menu component to demonstrate actually using this new functionality

### How To Use
```
class MyComponent extends BoltElement {
  static schema = schema; // --> 1. add imported schema data to a static `schema` property

  static get properties() {	  
    return {
      ...this.props, // --> 2. then add the auto-generated `this.props` property data to your component
     
      // 3. optionally add any extra props not included in the component schema 
      onClickTarget: {
        type: String,
        attribute: 'on-click-target',
      },
      ...
  }
}
```

## How to test
- Review code changes, especially the decorator updates to LitElement
- Smoke test [Menu](https://feature-auto-prop-support.boltdesignsystem.com/pattern-lab/?p=viewall-components-menu) to confirm everything continues to work as expected given CI tests are all passing


packages/base-element/src/lib/decorators.js
```diff
@@ -298,3 +300,115 @@ export const renderAndRenderedEvents = () => classOrDescriptor =>
...
+            // @todo: re-evaluate this switch default for handling `any` + multi-types
+            default:
+              propType = Object;
+              break;
+          }
+
+          props[propName] = {
+            type: propType,
+            reflect:
+              property.type === 'boolean' || property.reflect ? true : false,
```
remydenton, line 382: I still have this question from #1795:

> Why are booleans forced to reflect? Seems like at the least property.reflect should take priority if it's set. But even then, I don't know why booleans should have a different default here.

sghoweri: Hmmm. I could have sworn that we previously ran into issues with Booleans not correctly triggering an attributeChanged event in LitElement unless reflect was set, however given that Button works just fine without this... maybe that was a bug that was fixed? 🤷 

I guess we should try _not_ auto-setting Booleans to reflect but still allow the `reflect` flag to get added in the schema itself. 

Give it a try and just keep an eye on this? How's that sound?

danielamorse: @remydenton @sghoweri updated so that booleans are not forced to reflect ([see commit](https://github.com/boltdesignsystem/bolt/pull/1818/commits/6cd783eafc9ccca02cea3d5d594112c7b922d216)).

I think the issue we ran into was that a boolean prop we expected to be reflected (and used as a CSS selector) was not showing up because `reflect` wasn't set at all. This change should allow us to set it as needed.

sghoweri: Ah, @danielamorse that would make sense. Thanks for updating that!

sghoweri:
@remydenton this should be ready for you to re-review!

packages/base-element/src/lib/decorators.js
remydenton: Thanks for the updated @danielamorse, much more sensible that way (allows you to override the default).

Not to beat this to death, but are you guys sure this isn't related to #1737?  That was based on the following premise, which it turned out is NOT true, and sounds similar to the reason Salem first gave.

> As a general rule, when using LitElement, boolean schema props must be configured with reflect: true. This allows external changes to the attribute to be reflected in the internal prop and vice versa.

In reality, the problem was we were just using the wrong litElement syntax.

If there is a case where a boolean is used as a CSS selector though, reflecting is probably important.

danielamorse: @remydenton @sghoweri yes, #1737 is the one I had in mind. The quoted premise is indeed NOT true. The resolution was to use the correct syntax.

BUT, if you want to set a prop internally and have it reflect as an attribute, you must still set `reflect: true`. Example: when a component is loaded, you set `this.ready = true` and you get a `[ready]` attribute.

Had to double-check my understanding here. I think this is worth clarifying. Let me know if I got anything wrong.

remydenton: > if you want to set a prop internally and have it reflect as an attribute, you must still set reflect: true

Yep, 100% agree.  Thanks for working through that with me.

sghoweri:
PR was released with v2.22.0

PR #1819, opened 2020-04-08
IE11 deprecation testing updates

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-2033

## Summary

Remove uses of `browser.sendKeys()` from our tests, only needed if we're supporting IE11.

## How to test

- Review changes
- Tests are passing

### DIFF
--- a/docs-site/__tests__/pattern-lab.e2e.js
+++ b/docs-site/__tests__/pattern-lab.e2e.js
@@ -23,13 +23,6 @@ module.exports = {
   //     .assert.elementPresent('.js-c-typeahead__input')
   //     .click('.js-c-typeahead__input'); // click on the PL search input
 
-  //   // type "Components-Card" in the input field. Adjust command based on browser support
-  //   if (browser.sendKeys) {
-  //     browser.sendKeys('.js-c-typeahead__input', 'Components-Card');
-  //   } else {
-  //     browser.keys('Components-Card');
-  //   }
-
   // browser.saveScreenshot(
   //   `screenshots/pattern-lab/pattern-lab-search-input--${browser.capabilities
   //     .browserName || 'chrome'}.png`,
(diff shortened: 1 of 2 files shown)


sghoweri:
@danielamorse FYI, you accidently removed the `.keys` Selenium command command which broke the latter half of the Typeahead E2E test -- just pushed up a fix reverting this change!

image

danielamorse:
@sghoweri, ah thanks for catching that!

This was originally based off of `master`. Then I rebased off of `feature/ie11-deprecation` because I thought that might be why the test failed. Do you agree that we should change this back to `master`, since it's not a breaking change, and it's one less thing we have to wait to release?

sghoweri:
> @sghoweri, ah thanks for catching that!
> 
> This was originally based off of `master`. Then I rebased off of `feature/ie11-deprecation` because I thought that might be why the test failed. Do you agree that we should change this back to `master`, since it's not a breaking change, and it's one less thing we have to wait to release?

@danielamorse I'd definitely vote getting this merged down to master ASAP (ie. not having to get merged into the other branch first) -- yes it's part of the broader initiative but there's no reason why this has to wait for the rest of those updates to land (even if they are both probably going to get merged in today 😄)

sghoweri:
> > @sghoweri, ah thanks for catching that!
> > This was originally based off of `master`. Then I rebased off of `feature/ie11-deprecation` because I thought that might be why the test failed. Do you agree that we should change this back to `master`, since it's not a breaking change, and it's one less thing we have to wait to release?
> 
> @danielamorse I'd definitely vote getting this merged down to master ASAP (ie. not having to get merged into the other branch first) -- yes it's part of the broader initiative but there's no reason why this has to wait for the rest of those updates to land (even if they are both probably going to get merged in today 😄)

Errrrr nope. On second thought, I take that back. Let's merge this into the `feature/ie11-deprecation` to avoid an extra rebase!

PR #1823, opened 2020-04-09
Fix: add only to twig includes

mikemai2awesome:
## Jira

http://vjira2:8080/browse/BDS-2102

## Summary

Fixes certain twig includes without `only` which would cause prop options to trickle down.

## How to test

Run the branch locally and check all Academy Blueprints, make sure they match exactly with their counterparts on `master`.


docs-site/src/pages/pattern-lab/_patterns/03-blueprints/02-molecules/cards/_horizontal-card.twig
```diff
@@ -2,12 +2,25 @@
...
     {% include "@bolt-blueprints/_icon-banner.twig" with {
-      gradient: cardGradient,
+      cardGradient: cardGradient,
+      gradientStart: gradientStart,
+      gradientMiddle: gradientMiddle,
+      gradientEnd: gradientEnd,
+      cardIcon: cardIcon,
       horizontal: true
-    } %}
+    } only %}
```
sghoweri, line 23: Adding two sets of the same variables to the same file (which didn't explicitly require defining and passing these down before) has a code smell to it.

I understand why this update was necessary once the `only` keyword was added -- I just don't 100% understand why all of these `only` updates are required to be added in the first place... is something breaking downstream which requires these updates?

IMHO, rather than introducing this sort of cruft, we'd be better off working through how we'd want to better manage these different variations. For example, we could pre-define certain Card variations used here and eliminate having to pass down certain config options entirely.

@danielamorse could use a 2nd opinion on this one.

remydenton: Thanks for weighing in on this @sghoweri.

>  I just don't 100% understand why all of these only updates are required to be added in the first place... is something breaking downstream which requires these updates?

Omitting `only` is a problem because it can result in lots of subtle, hard to track down bugs.  The best example is when you have an `attributes` variable intended for one pattern that unintentionally leaks through to others, adding unintentional classes, duplicate IDs, etc.  That's just one example, and they aren't theoretical-- Academy omitted `only` in a lot of places and has run into tricky problems (or used messy workarounds) as a result.  I'm putting my foot down about this one-- always use `only` unless you know what you're doing (and even then, there's rarely a good reason).

> Adding two sets of the same variables to the same file (which didn't explicitly require defining and passing these down before) has a code smell to it.

If we didn't use twig blocks, and therefore also didn't use embeds, you would only see these variables once.  It's one of several reasons for not using twig blocks in our code at all, which is that item that's on our dev huddle agenda.

Overall, I agree that this commit is starling, but what you're seeing isn't a problem introduced by the new code; it's a surfacing of problems that already exist and were previously just hidden (e.g. that you can pass a variable called `gradientStart` to `_horizontal-card.twig` and that it will actually work).

> we'd be better off working through how we'd want to better manage these different variations. For example, we could pre-define certain Card variations used here and eliminate having to pass down certain config options entirely.

I agree that we should work to improve what's going on here.  But this branch isn't the place-- it's just the messenger, don't shoot the messenger!

docs-site/src/pages/pattern-lab/_patterns/03-blueprints/04-templates/_blueprints-landing-page-template.twig
```diff
@@ -24,7 +24,7 @@
       size: "none",
       valign: "start",
       content: mainContentBlock
-    } %}
+    } only %}
```
danielamorse, line 27: This one cuts off the `backgroundBlock` variable which is used on the Certification Exam page to add the blue radial background. It's missing on [this branch](https://boltdesignsystem-o84g1jw1d.now.sh/pattern-lab/?p=blueprints-certification-exam-page) compared to the [latest release](https://boltdesignsystem.com/pattern-lab/?p=blueprints-certification-exam-page). I'm inclined to remove this `only` just so the template looks right. I see no simple way to pass the background through without passing a non-schema prop to Band - Band lets you do this BTW without a validation error 🤔. I figure passing a random prop (`backgroundBlock`) to Band would be a worse example than omitting this `only`. Thoughts?

docs-site/src/pages/pattern-lab/_patterns/03-blueprints/02-molecules/cards/_horizontal-card.twig
danielamorse: @sghoweri @remydenton here's a good example of a subtle bug that occurs when we don't use `only`.

On the Mission Dashboard,
```
{% include "@bolt-components-card-replacement/card-replacement.twig" with {
  body: {
    content: "foo"
  }
} %}
```
outputs:
```
<bolt-card-replacement>
  foo
</bolt-card-replacement>
```
Whereas,
```
{% include "@bolt-components-card-replacement/card-replacement.twig" with {
  body: {
    content: "foo"
  }
} only %}
```
outputs:
```
<bolt-card-replacement>
  <bolt-card-replacement-body>
    foo
  </bolt-card-replacement-body>
</bolt-card-replacement>
```
I found this while debugging a spacing-related issue caused by the introduction of the `bolt-card-replacement-body` element.

docs-site/src/pages/pattern-lab/_patterns/03-blueprints/04-templates/_blueprints-landing-page-template.twig
remydenton: Good catch, thank you.

As to why there's no validation error, that's because json schema has an `additionalProperties` keyword that defaults to "true".  Basically, that means that if the schema doesn't mention a prop that you're passing, that's OK.  See https://json-schema.org/understanding-json-schema/reference/object.html.

As to this particular problem, I just pushed an update that allows us to skip the block entirely.  Part of my campaign to get rid of blocks 😉.  Let me know if that feels better.

As to whether it's a good idea to list out properties in the embed that aren't actually part of the component being embedded (because they're needed in an interior block), I think it's the lesser evil.  Open for debate, but at least then you're aware of what data the blocks need, and there's no way for variables to sneak through unannounced (like the gradients in toolbar).  Best of all, don't use any blocks at all, and therefore don't use embeds, and you never run into this type of problem (and your component level code stays simpler as a bonus).

danielamorse: I feel good about your solution^. I didn't realize `additionalProperties` was a feature not a bug. Let's go with that.

I fixed one more related issue - renaming variable to `background` causes the footer template to take on the radial background. Adding `only` fixes it.

Once tests pass, I'm good to merge this.

docs-site/src/pages/pattern-lab/_patterns/03-blueprints/02-molecules/cards/_horizontal-card.twig
sghoweri: 😨  ok, I'm sold

docs-site/src/pages/pattern-lab/_patterns/03-blueprints/04-templates/_blueprints-landing-page-template.twig
remydenton: Thank's for catching that too!  I guess another example of how easy it is to make mistakes when `only` isn't there 😉 .

sghoweri:
PR was released with v2.22.0

PR #1826, opened 2020-04-10
Feature/update component generator

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-2097

## Summary

Update component generator to use latest conventions, some other minor improvements.

## Details

- Convert schema from YML to JS
  - Remove description from schema, not required as docs reference `package.json`
- Update component JS to use latest conventions, including auto-props
- Remove Polyfill loader, not required since dropping IE11 support
- Add conditional prop example to component Twig template
- Add title comment to SCSS, fix formatting
- Remove boilerplate component description. We tend to include this on every component and it doesn't add much. Make authors describe the component.
- Update Jest tests to use latest conventions
  - Update generator test snapshot
  - Remove dupe Jest test file

## How to test

- Review changes
- Tests are passing
- Checkout locally and make a component (`yarn cc`)
  - Verify tests pass on that component


packages/generators/yeoman-create-component/generators/component/templates/component.test.js
```diff
@@ -69,67 +52,59 @@ describe('<%= props.name.noCase %>', async () => {
     expect(results.html).toMatchSnapshot();
   });
 
-  test('Default <bolt-<%= props.name.kebabCase %>> w/o Shadow DOM renders', async function() {
-    const renderedComponentHTML = await page.evaluate(() => {
-      const component = document.createElement(
-        'bolt-<%= props.name.kebabCase %>',
+  test('renders w/ Shadow DOM', async function() {
```
adamszalapski, line 55: A typo here should be `renders w/o Shadow DOM`

packages/generators/yeoman-create-component/generators/component/templates/component.schema.js
```diff
@@ -0,0 +1,17 @@
...
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  title: '<%= props.name.titleCase %>',
+  type: 'object',
+  properties: {
+    attributes: {
+      type: 'object',
+      description:
+        'A Drupal attributes object. Applies extra HTML attributes to the &lt;bolt-<%= props.name.kebabCase %>&gt; tag.',
+    },
+    disabled: {
```
adamszalapski, line 11: Correct me if I'm wrong but I think that `disabled` prop has default value and this is false. Because we check if this prop is true or not and determine how the component should behave/looks like.

mikemai2awesome: I second this.

packages/generators/yeoman-create-component/generators/component/templates/component.test.js
danielamorse: "w/" is an abbreviation for "with", but no need to abbreviate here. I'm going to replace with full word for clarity.

packages/generators/yeoman-create-component/generators/component/templates/component.schema.js
danielamorse: Good call, fixed

danielamorse:
> @danielamorse good job i have only two comments. I noticed not in your code a bug because file generated in `__tests__` folder is called `index.js` but should be called like component eg: `bolt-test.js`.
> I fixed this in my commit.

@adamszalapski good call. Now it is consistent with how we name our tests (and how the generator names the test). Thanks for updating 👍

danielamorse:
> @danielamorse
> 
> Regarding @adamszalapski's last commit, does that make it `__tests__/test.js` or `__tests__/[component-name].js`? If it's the latter then it's all good.

It's the latter. The component-generator Jest test creates a component named "test", so that's the name it looks for. Should be all good.

sghoweri:
PR was released with v2.22.0

PR #1830, opened 2020-04-17
Truncate Chip List

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-1455

## Summary

Add truncation support to Chip Lists.

## Details
This PR allows users to truncate long Chip Lists by setting the `truncate` prop.

### No-JS
Before JS loads we want the list to be truncated so that there's no FOUC, but we still want the list to be expandable. We do this using the `label` + `input:checked` method, combined with a Twig-generated `<style>` tag.

### With JS
Once the WC loads, it strips out the `<style>` tag and replaces the original `label` with a Trigger.

### Additional changes
- Add `truncate` prop to Chip List, which truncates list after a given number of items.
- Add `canCollapse` prop to Chip List. By default, once the list expands it cannot be collapsed. Set this prop to `true` to allow it to be collapsed again.
- Output `<bolt-chip-list>` tag, previously just rendered `<bolt-list>`.
  - Allows us to contain the `<style>` tag used for no-JS functionality.
  - Allows us to add ID to Chip List and scope our no-JS styles.
- Add Web Component JS.
  - There is no WC demo page because Chip List is still just a wrapper around a Bolt List that supports  truncation. If you put a `bolt-list` inside `bolt-chip-list` WC it does not automatically apply the expected props to `bolt-list`, as you might expect. This may be useful in the future but was beyond the scope of this task. For this reason there's no WC demo page.
- Add `iconOnly` prop to Chips to make Show more Chip (`...` icon) possible. Add test and demo for `iconOnly`.
- Add `type` to Chip List `size` prop, previously missing.
- Add Twig support for passing attributes to `bolt-list-item`, required to flag certain list items as `[truncate-trigger]`

## How to test
- Review changes
- Tests are passing
- Review demos on feature branch:
  - No regressions to basic [Chip list](https://feature-truncate-chip-list.boltdesignsystem.com/pattern-lab/?p=components-chip-list)
  - No regressions to [List](https://feature-truncate-chip-list.boltdesignsystem.com/pattern-lab/?p=viewall-components-list)
  - `truncate` works as expected on [demo page](https://feature-truncate-chip-list.boltdesignsystem.com/pattern-lab/?p=components-chip-list-truncate)
  - `canCollapse` works as expected on [demo page](https://feature-truncate-chip-list.boltdesignsystem.com/pattern-lab/?p=components-chip-list-can-collapse)
  - `iconOnly` works as expected on [demo page](https://feature-truncate-chip-list.boltdesignsystem.com/pattern-lab/?p=components-chip-icon-variations)


packages/components/bolt-chip-list/chip-list.schema.yml
```diff
@@ -12,11 +12,27 @@ properties:
...
+    type: string
     description: "Sets the size used for all of the chips (if size isn't specified on the individual chip)"
     enum:
       - xsmall
       - small
       - medium
+  truncate:
+    type: number
+    description: 'Sets the max number of chips to show before truncating, 0 or greater'
+  canCollapse:
```
mikemai2awesome, line 24: We can use `collapsible` for this.

danielamorse: Much better name, done.

danielamorse:
@mikemai2awesome addressed both issues.

sghoweri:
PR was released with v2.22.0

PR #1851, opened 2020-04-22
Spacing and Typography | Core | Switch to regular spacing system

mikemai2awesome:
## Jira

To be created when vpn works.

## Summary

Reduces spacing and font-size overall.

## Details

1. Redefined the whole spacing scale.
2. Redefined all the font-size maps and variables.
3. Kept the Japanese specific system as its own system and not related to base font-size and spacing.
4. Created SASS variables for the wrapper padding and used them in wrapper, page header, navbar, and toolbar. These variables are using absolute px value so the page max width and padding are immune to font-size and spacing changes.

## How to test

Run the branch locally and QA all pages, make sure things still line up and there is no awkward white space.

## Notes

Drupal needs to update the mega nav's padding. I should be doing this, will need @remydenton's help.


packages/components/bolt-text/__tests__/text.js
```diff
@@ -157,7 +157,7 @@ describe('<bolt-text> Component', () => {
     ).toBe(false);
 
     expect(image).toMatchImageSnapshot({
-      failureThreshold: '0.01',
+      failureThreshold: '0.015',
```
adamszalapski, line 160: It's a reason to increase the Threshold?

mikemai2awesome:
@adamszalapski @remydenton @sghoweri @danielamorse 

Since I need to update all tests and snapshots, there will be conflicts as long as master keeps changing. So you can ignore those for now and review the actual work. Before we merge into master, I will update all tests and snapshots again. I can't keep repeating this right now and waste time on it.

colbytcook:
PR was released with v2.24.0

PR #1860, opened 2020-04-30
Globally wire up remaining Bolt components to use new lazy queue / on-demand loading pattern

sghoweri:
## Jira
http://vjira2:8080/browse/BDS-2105

## Dependencies**
**to avoid any rendering issues and keep the # of changes to a minimum

- ~[Migrate Accordion to LitElement](https://github.com/boltdesignsystem/bolt/pull/1858)~
- ~[Add Classes to Slotted Children](https://github.com/boltdesignsystem/bolt/pull/1859) decorator work~
- ~[Migrate components to use `<ssr-keep>`](https://github.com/boltdesignsystem/bolt/pull/1857)~
- ~Switch from _eagerly_ loading components to [lazily loading components](https://github.com/boltdesignsystem/bolt/pull/1854)~

## Summary
In a nutshell, this PR takes the new lazy queue web performance updates from https://github.com/boltdesignsystem/bolt/pull/1821 and rolls it out to our components globally -- which should result in _significantly_ improved client-side performance.

For example, with these updates (plus the two separately PR'd updates mentioned below), our main JS bundle goes from being around **`1.9MB`** in size (minified but not gzipped) down to around **`50KB`** (again, minified but not gzipped).

image

> Even unoptimized, these updates start to show the full perf improvements seen on the full POC branch (which tests out all of these updates, combined)

<hr>

## Performance Impact
- All dynamic imports in Bolt are now bundled and loaded separately ("lazy") vs eagerly bundled together as they were previously
- All JS bundles in Bolt are now loaded with the new lazy queue package added in #1821 -- this means that only the components required on a page are loaded and components that get injected later will load on demand
- When compiled for production (without making any local code changes), the main `bolt-global.modern.js` JS file size should now be around `360KB` (minified but not gzipped)

Note: once the work from https://github.com/boltdesignsystem/bolt/pull/1855 and https://github.com/boltdesignsystem/bolt/pull/1846 get in, the fully compiled JS from Bolt (the initial JS that gets loaded) should be closer to 50KB (minified) or around 14KB (minified + gzipped)!

## How To Test
As weird as this might sound, I'd actually recommend we don't focus too much on the actual code for this (at least initially) and initially focus our efforts on spot checking the staged docs site.

Besides confirming that our tests are all passing (which is super easy to do), the main thing that we need to do here is just poke around Pattern Lab to do another across-the-board gut check (especially on heavier demos like some of the Academy page Blueprints or the Tabs / Accordion demos) and verify that things all seem to be working exactly as expected.

Components are loading fine, nothing is rendering weirdly, no flashes of unstyled UI, etc.

Some cross-browser spot checks would also be wise but overall I'm not expecting many (if any) surprises!


docs-site/src/pages/pattern-lab/_patterns/04-pages/99999-bolt-dev-sandbox/-05-ckeditor-5.twig
```diff
@@ -9,28 +9,28 @@
...
-        <bolt-text headline="true" font-size="xsmall" tag="h6">Bolt headline 6</bolt-text>
-        <bolt-text subheadline="true" font-size="xxlarge" tag="p">Bolt subheadline 1</bolt-text>
-        <bolt-text subheadline="true" font-size="xlarge" tag="p">Bolt subheadline 2</bolt-text>
-        <bolt-text subheadline="true" font-size="large" tag="p">Bolt subheadline 3</bolt-text>
-        <bolt-text eyebrow="true" tag="p">Bolt eyebrow</bolt-text>
-        <bolt-text eyebrow="true" tag="a">Bolt link</bolt-text>
-        <bolt-blockquote>Bolt blockquote</bolt-blockquote>
-      {% endspaceless %}
-    </textarea>
+    <bolt-ck-editor>
```
danielamorse, line 12: Does this PR add a new `bolt-ck-editor` component?

packages/components/bolt-tabs/src/tabs.js
```diff
@@ -10,7 +10,7 @@ import {
   waitForTransitionEnd,
 } from '@bolt/core-v3.x/utils';
 import { withLitHtml } from '@bolt/core-v3.x/renderers/renderer-lit-html';
-import { smoothScroll } from '@bolt/components-smooth-scroll';
+import { smoothScroll } from '@bolt/components-smooth-scroll/src/smooth-scroll';
```
danielamorse, line 13: Why this change?

packages/experimental/editor/src/editor.twig
```diff
@@ -4,16 +4,18 @@
...
-    {% autoescape 'js' %}
-      {{ config | json_encode | raw }}
-    {% endautoescape %}
-  </script>
-  <bolt-button color="primary" size="xsmall" class="js-pega-editor__trigger">Edit</bolt-button>
-  <bolt-button color="secondary" size="xsmall" class="js-pega-editor__feedback">Submit Feedback</bolt-button>
-  <br>
-  <div class="js-pega-editor__space">
-    {{ content }}
+<bolt-editor>
```
danielamorse, line 7: Does this PR add a new `bolt-editor` component?

packages/experimental/editor/src/main.js
```diff
@@ -0,0 +1,364 @@
+import { query } from '@bolt/core-v3.x/utils';
```
danielamorse, line 1: Is this related to lazyQueue?

packages/components/bolt-tabs/src/tabs.js
sghoweri: The `main` key In the package.json points to the file that contains lazyQueue code vs the file that we're pointing to here contains the JS that we need to import directly.

packages/experimental/editor/src/editor.twig
sghoweri: No, we're just using that "fake" tag selector (the first param in lazy queue) to know if / when to run the editor code. 

Without a selector to check for, our lazyQueue library will always load the relevant code (ie. load the code even when it's not needed which is most of the time).

packages/experimental/editor/src/main.js
sghoweri: Nope. All the code in main.js here is 1:1 copied over from the current index.js file

docs-site/src/pages/pattern-lab/_patterns/04-pages/99999-bolt-dev-sandbox/-05-ckeditor-5.twig
sghoweri: Same goes with this one. We're just using this tag to know when / when not to load the CK Editor demo code.

sghoweri:
> One thing I notice is that JS loads super fast, but images take a bit longer than before. Have you noticed that? Fast JS is still preferable, but I wonder if it's blocking images.

@danielamorse I didn't notice that myself but will take another look at this in the morning to see if there's anything we need to tweak on master before we cut the release.

Thank you very much for taking the time to go through this -- I know this was a lot but I'm optimistic that this'll be more than worth it from the perf gains! 🙏

sghoweri:
PR was released with v2.23.0

PR #1867, opened 2020-05-04
Micro Journey intersection observer-related fix to prevent "restarts" when scrolling away

sghoweri:
## Jira
http://vjira2:8080/browse/WWWD-4726

## Summary
Updates the Micro Journey JS logic in the Interactive Pathways subcomponent to only have the `beginItAll` (aka init() function) method fire _once_ so the animation sequence won't be unintentionally restarted. 

## Details
This small patch also cleans up any existing observers present which also helps to ensure that animations (or previous interactions the user has taken) aren't unintentionally restarted.

## How to test
This JS update is 1:1 with what was already tested and approved downstream ([see this comment for reference](http://vjira2:8080/browse/WWWD-4726?focusedCommentId=79974&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-79974)) but in general with these updates any panels that have already been opened (or scrolled away from before scrolling back) won't unexpectedly restart.

### DIFF
--- a/packages/experimental/micro-journeys/src/interactive-pathways.js
+++ b/packages/experimental/micro-journeys/src/interactive-pathways.js
@@ -76,10 +76,10 @@ class BoltInteractivePathways extends withLitContext {
   }
 
   connectedCallback() {
-    super.connectedCallback();
+    super.connectedCallback && super.connectedCallback();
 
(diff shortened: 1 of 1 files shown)


sghoweri:
PR was released with v2.23.0

PR #1868, opened 2020-05-04
Link as popover trigger

danielamorse:
## Jira

http://vjira2:8080/browse/BDS-2057

## Summary

Allows Link to be used as a trigger for Popover.

## Details

Currently, if you pass `<bolt-link>` as a trigger for Popover it results in nested anchor tags, which is invalid and breaks the layout. This happens because Twig wraps the trigger in an anchor to make the Popover accessible when JS is off.

We want to use Link with Popover, and we want to keep our current no-JS solution. So, the fix was to make Link internally render a `<button>` when no `url` is provided. Semantically, this is better than outputting an `<a>` that is not actually an anchor, and it solves the nested anchor problem.

## How to test
- Review code changes
- See the [Link with Button element](https://feature-link-as-popover-trigger.boltdesignsystem.com/pattern-lab/?p=components-link-with-button-element) demo page
- Locally, open the [basic popover demo](https://github.com/boltdesignsystem/bolt/blob/master/docs-site/src/pages/pattern-lab/_patterns/02-components/popover/05-popover.twig#L9-L12) and replace the button trigger with a link without `url`:
```
    text: "This triggers a popover"
  } only %}
```
- Verify it works (with and without JS)
- Locally, open the [web component popover demo](https://github.com/boltdesignsystem/bolt/blob/master/docs-site/src/pages/pattern-lab/_patterns/02-components/popover/999-popover-with-web-component.twig#L23-L25) and in the `popover_demo` example, replace `<bolt-button>` with:
```
  <bolt-link >
    This triggers a popover
  </bolt-link>
```
- Verify it works (with JS only)
- Review [Link demos](https://feature-link-as-popover-trigger.boltdesignsystem.com/pattern-lab/?p=viewall-components-link) and verify no other regressions

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/02-components/link/28-link-with-button-element.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/02-components/link/28-link-with-button-element.twig
@@ -1,0 +1,6 @@
+<h3>Link with Button element</h3>
+<p>When you want the appearance of Link but the on-click behavior is handled separately (by Tooltip for example), omit the <code>url</code> property and the Link component will internally render a <code>&lt;button&gt;</code>.</p>
+
+{% include "@bolt-components-link/link.twig" with {
+  text: "Link with button element"
+} only %}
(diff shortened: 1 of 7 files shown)


sghoweri:
PR was released with v2.23.0

PR #1873, opened 2020-05-14
docs: resources & helpful links

sghoweri:
Docs pointing to helpful resources + sources of inspiration


danielamorse:
@remydenton @mikemai2awesome @colbytcook, these are useful docs that I don't quite know how to fit into our existing docs site.

For now, I'd like to move them into `docs-site/src/pages/docs/` and make them hidden. We can refer back to them when we one day clean up our docs. Sound good?

remydenton:
Yes, great call @danielamorse.  Docs organization is hugely important in my mind and I'm glad you're considering it.

PR #1875, opened 2020-05-14
Add demo of using Bolt with a vanilla Webpack build

sghoweri:
Fully functional demo of compiling Bolt web components via a vanilla Webpack build (ie. without using the Bolt Build Tools)


danielamorse:
@colbytcook feel free to merge this if you think it'll be helpful.

PR #1878, opened 2020-05-20
Feature/device viewer enable video embed

relimck:
## Jira

WWWD-4750
https://pegadigitalit.atlassian.net/browse/BDS-35

## Summary

Added an option of embed a video player inside a device viewer component.
Only images could be placed there.


packages/components/bolt-device-viewer/src/device-viewer.twig
```diff
@@ -61,6 +61,9 @@
           {% if image %}
             {% include "@bolt/image.twig" with image only %}
           {% endif %}
+          {% if video %}
+            {% include "@bolt-components-video/video.twig" with video only %}
+          {% endif %}
```
remydenton, line 66: I would like to see this here instead:
```
{% if media %}
  {{ media }}
{% endif %}
```
The reason for this is that we're trying to avoid enforcing a specific data format whenever possible during component composition.  I realize this isn't how image is handled, but the way image is handled is outdated (and I would consider it wrong for any new work).

Additionally, don't forget to add the new prop to the schema for device viewer.

relimck: @remydenton thank you for feedback. I updated a code based on your suggestion.

PR #1879, opened 2020-05-22
Move layout change event to be dispatched after bolt-interactive-path…

relimck:
Move layout change event to be dispatched after bolt-interactive-pathway step rendered.

## Jira

WWWD-4784

## Summary

A page height might be updated more frequently than on pathway change, but also when step change.

I moved `bolt:layout-size-changed` to be triggered after `setActiveStep()` rendered for pathway which support both cases.


colbytcook:
PR was released with v2.24.0

PR #1882, opened 2020-05-27
Allow autoplay video for iPhone/iPad when controls disabled

relimck:
## Jira

WWWD-4750

## Summary

Video autoplay does not work on iPhone and iPad.
Because IOS policy `<video>` tag require extra parameter `playsinline` added.

## Details

Refer docs:
- https://medium.com/code-divoire/autoplaying-inline-videos-on-iphone-ios-10-using-angular-d4e2eaba2164
- https://webkit.org/blog/6784/new-video-policies-for-ios/


remydenton:
As I understand it, this actually fixes autoplay on iOS.  If so, we should remove the note on this docs page: https://boltdesignsystem.com/pattern-lab/?p=components-autoplay-and-loop-hide-controls

> Note: videos are not expected to autoplay on iOS.

colbytcook:
PR was released with v2.24.0

PR #1883, opened 2020-05-28
BDS-2118: Investigate options for Brightcove error styling/UX

colbytcook:
## Jira

BDS-2118: Investigate options for Brightcove error styling/UX

## Summary

Included an update to the display of the Brightcove error to make it closer to brand

## Details

This is a hotfix to make the Brightcove video display look better before PegWorld. This is a quick fix and a better solution should be figured out post PegaWorld.

## How to test

1. Navigate to https://feature-bds-2118.boltdesignsystem.com/pattern-lab/?p=components-video-with-error-message
2. Observe the video is now displaying with an indigo background color and white text horizontally and vertically centered.

### DIFF
--- a/packages/components/bolt-video/src/_video-tag.twig
+++ b/packages/components/bolt-video/src/_video-tag.twig
@@ -71,7 +71,7 @@
 {% set attributes = attributes.setAttribute("share-description", share_description|default('Share This Video'|t)) %}
 
 {% if errorMessage %}
-  {% set attributes = attributes.setAttribute("error-message", errorMessage) %}
+  {% set attributes = attributes.setAttribute("error-message", errorMessage|default("This video didn't load correctly. Refresh page to view."|t)) %}
 {% endif %}
 
 {# @todo: `isBackgroundVideo` will be removed with Bolt v3.0 #}
--- a/packages/components/bolt-video/src/_video-tag.twig
+++ b/packages/components/bolt-video/src/_video-tag.twig
@@ -70,6 +70,10 @@
 
 {% set attributes = attributes.setAttribute("share-description", share_description|default('Share This Video'|t)) %}
 
+{% if errorMessage %}
+  {% set attributes = attributes.setAttribute("error-message", errorMessage) %}
+{% endif %}
+
 {# @todo: `isBackgroundVideo` will be removed with Bolt v3.0 #}
 {% if isBackgroundVideo %}
   {% set attributes = attributes.setAttribute("is-background-video", isBackgroundVideo) %}
--- a/packages/components/bolt-video/src/_videojs-enhancements.scss
+++ b/packages/components/bolt-video/src/_videojs-enhancements.scss
@@ -369,21 +369,12 @@ $bolt-video-js-button-shadow-level-hover: 'level-40';
         text-align: center;
       }
       .vjs-errors-headline{
+        @include bolt-font-size(small);
         display: block;
         width: 100%;
         padding-right: 0;
-        line-height:0;
-        font-size:0;
-        &:after{
-          @include bolt-font-size(small);
-          // @todo: We need to come up with a better solution to handle this through the Brightcove JS API
-          content: "This video didn't load correctly. Refresh page to view.";
-          display:block;
-        }
         @include bolt-mq(medium) {
-          &:after {
-            @include bolt-font-size(medium);
-          }
+          @include bolt-font-size(medium);
         }
       }
 
--- a/packages/components/bolt-video/src/_videojs-enhancements.scss
+++ b/packages/components/bolt-video/src/_videojs-enhancements.scss
@@ -342,7 +342,7 @@ $bolt-video-js-button-shadow-level-hover: 'level-40';
 
   &.vjs-error{
     .vjs-error-display{
-      background: rgba(bolt-color(blue), 1);
+      background: rgba(bolt-color(indigo), 1);
       &:before{
         content:'';
         display:none;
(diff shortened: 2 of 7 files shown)


packages/components/bolt-video/src/_videojs-enhancements.scss
```diff
@@ -339,8 +339,58 @@ $bolt-video-js-button-shadow-level-hover: 'level-40';
...
+        bottom: 15px;
+        text-align: center;
+      }
+      .vjs-errors-headline{
+        @include bolt-font-size(medium);
+        display: block;
+        width: 100%;
+        padding-right: 0;
+        &:after{
+          content: '. Refreshing the page may help resolve this issue.'
```
mikemai2awesome, line 375: This is not translatable. We should avoid generating actual content with CSS, only use it for decorations.

packages/components/bolt-video/src/_videojs-enhancements.scss
```diff
@@ -339,8 +339,58 @@ $bolt-video-js-button-shadow-level-hover: 'level-40';
       transform: translate3d(-50%, -50%, 0);
     }
   }
+
+  // Error styling
+  &.vjs-error{
+    .vjs-error-display{
+      background: rgba(bolt-color(blue), 1);
```
mikemai2awesome, line 346: There's no transparency here, you don't need to use rgba. And `bolt-color(blue)` has been phased out pretty much. I think you are looking for `bolt-color(indigo)`.

mikemai2awesome:
@colbytcook Can you create a demo page and a test for this?

packages/components/bolt-video/src/_videojs-enhancements.scss
```diff
@@ -339,8 +339,58 @@ $bolt-video-js-button-shadow-level-hover: 'level-40';
...
+      .vjs-errors-dialog{
+        top: 0;
+        right: 0;
+        bottom:0;
+        left: 0;
+        padding: 0;
+        border:0;
+      }
+      .vjs-errors-content-container{
+        @include bolt-padding(40px, 40px);
```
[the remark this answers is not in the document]
colbytcook: Spacing tokens are ideal, I just didn't find them and wanted to get this in front of everyone sooner than later

packages/components/bolt-video/src/_videojs-enhancements.scss
colbytcook: I completely agree but this is to get a fast fix for Paul and team to review. Unfortunately, this needs to be resolved before PegaWorld and I am not sure we have enough time to look into doing it the correct way (hooking into the Brightcove API and determining the messaging based on the error return). But this is something that has to be followed up on immediately after PegaWorld.

I will certainly make sure that everyone is aware that this is a quick and dirty fix that will not be translated. If this becomes a problem then I can revisit this. For a bandaid we can hook into the lang tag and just update the content value base don it (I know, the sins...).

mikemai2awesome: I don't think the lang tag works well with our current Drupal translate setup. @remydenton has more insights into that. For the time being, please add a `@todo` so we remember we did this.

remydenton:
One additional change: is there a way we can still show the error code and session ID, maybe in very small/unobtrusive font?  That info is helpful for troubleshooting video problems when someone sends a screenshot.

packages/components/bolt-video/src/_videojs-enhancements.scss
remydenton: If we can add the text using javascript instead of CSS, we can pass it through the Drupal t function.  We're doing something similar for the "Share this video" text (which was also originally a CSS pseudo element).  Text in twig templates can also be passed through the t function-- it's just CSS pseudo elements that don't support translation.

packages/components/bolt-video/src/video.standalone.js
```diff
@@ -228,6 +229,15 @@ class BoltVideo extends withPreact {
     player.on('ended', function() {
       elem.onEnded(player);
     });
+
+    player.on('error', () => {
+      let errorMessage = elem.getAttribute('error-message');
```
danielamorse, line 234: @colbytcook, some tweaks I'd make here:
- Since `errorMessage` is a prop you can get it via `elem.props.errorMessage`
- If you only expect one element you can use `querySelector()` instead of `querySelectorAll`
- You might want to test `headlineHTML` exists before using it

All together would look like this:
```
const headlineHTML = elem.querySelector('.vjs-errors-headline');
if (headlineHTML) {
  headlineHTML.innerHTML =
    elem.props.errorMessage ||
    "This video didn't load correctly. Refresh page to view.";
}
```

packages/components/bolt-video/src/video.standalone.js
```diff
@@ -228,6 +229,15 @@ class BoltVideo extends withPreact {
...
+      const headlineHTML = elem.querySelectorAll('.vjs-errors-headline')
+      if (!errorMessage) {
+        errorMessage = "This video didn't load correctly. Refresh page to view."
+      }
+      headlineHTML[0].innerHTML = errorMessage
```
danielamorse, line 239: Do you have the [Prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) installed for VS Code? That would add a semi-colon to the end of this line.

packages/components/bolt-video/src/_video-tag.twig
```diff
@@ -70,6 +70,10 @@
 
 {% set attributes = attributes.setAttribute("share-description", share_description|default('Share This Video'|t)) %}
 
+{% if errorMessage %}
+  {% set attributes = attributes.setAttribute("error-message", errorMessage) %}
+{% endif %}
```
remydenton, line 75: If you set the default for the errorMessage string here, you can use the Drupal `|t` function available in twig, which will then automatically expose this for translation.  See this example a few lines above:

`share_description|default('Share This Video'|t)`

In your case:

`errorMessage|default('This video didn't load correctly. Refresh page to view.'|t)`

Yes, it means we'll have the string twice in our codebase.  But it means as long as video is being used with twig on the Drupal side, it becomes translatable without anyone having to think about it.

Full explanation here (it starts with PHP, twig stuff is about half way down): https://chromatichq.com/blog/drupal-code-standards-t-function

packages/components/bolt-video/src/video.standalone.js
```diff
@@ -231,6 +231,7 @@ class BoltVideo extends withPreact {
     });
 
     player.on('error', () => {
+      console.log(player)
```
remydenton, line 234: I'm assuming you didn't mean to keep to `console.log()`?

colbytcook:
PR was released with v2.24.0

PR #1889, opened 2020-06-15
WWWD-4862/BDS-2157 broken watch icon

remydenton:
## Jira

http://vjira2:8080/browse/WWWD-4862
http://vjira2:8080/browse/BDS-2157

## Summary

Restores broken watch icon

## Details

This appears to have been broken with the SVG size optimizations introduced in https://github.com/boltdesignsystem/bolt/pull/1861/files#diff-60fb4b2c53d51d3d8aab9999372af25a

Specifically, the problem seems to be the with the `convertPathData` settings.  This PR changes a few of the sub settings back to match [the defaults](https://github.com/svg/svgo/blob/master/plugins/convertPathData.js).  I just used guess-and-check, so I don't understand much of the details.

@mikemai2awesome, @danielamorse, @danielamorse, @colbytcook, if any of you know more about SVG and SVGO, I'd be curious to hear your thoughts on this.  Could a better fix be to re-create the watch icon and keep our SVGO optimizations?


remydenton:
Closing in favor of #1892

PR #1896, opened 2020-06-24
Feature/ds 101

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-101](https://pegadigitalit.atlassian.net/browse/DS-101)

## Summary

Matched version dependencies and ran yarn upgrade to resolve issue with node 12.18.1

## Details

After installing Node 12.18.1 I noticed that the build was throwing errors. I ran it all again and caught that there was some version mismatching and a warning about caniuse-lite needing to be updated via `yarn upgrade`. After making all the mentioned changes, it seems like Node 12.18.1 is now working.

## How to test

I would recommend start with the master branch, installing Node v12.18.1 and trying to run the bolt setup commands to confirm that the error is actually repeatable. After that, pull this branch and try again to confirm the issue is resolved.

### DIFF
--- a/packages/website-ui/uikit-workshop/package.json
+++ b/packages/website-ui/uikit-workshop/package.json
@@ -140,11 +140,11 @@
   },
   "optionalDependencies": {
     "fibers": ">= 3.1.0",
-    "preact": ">=10 || ^10.0.0-alpha.0",
+    "preact": "^10.2.1",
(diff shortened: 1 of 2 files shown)


colbytcook:
We are going to hold off on this, it isn't something we should get into at the moment.

PR #1903, opened 2020-07-06
Component generator updates

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-114

## Summary

Fix component generator regressions, update to reflect recent architectural changes.

## Details

- Check that changelog file exists before using it in Twig.
- Don't assume latest Bolt release version number matches latest Bolt Core version number. In the case of a hotfix they might be different, and this causes generator to add core dependency at a version number that doesn't exist.
- We've consolidated how we include Bolt components in `.boltrc`. Instead of adding each component individually to [docs-site `.boltrc`](https://github.com/boltdesignsystem/bolt/blob/master/docs-site/.boltrc.js#L87-L119), we add them to [Starter Kit `.boltrc`](https://github.com/boltdesignsystem/bolt/blob/master/packages/base-starter-kit/.boltrc.js#L3-L68). Updated generator to reflect this change ([see file](https://github.com/boltdesignsystem/bolt/pull/1903/files#diff-ceada4237c607da1d4e97d1d83ed9351)).
- Update docs and tests.

## How to test

- Checkout branch and create a new component.
- Verify the changes to Starter Kit `.boltrc` are correct.

### DIFF
--- a/packages/generators/yeoman-create-component/generators/component/index.js
+++ b/packages/generators/yeoman-create-component/generators/component/index.js
@@ -15,6 +15,10 @@ const { addBoltPackage } = require('./add-bolt-package');
 const currentBoltVersion = require('../../../../../docs-site/package.json')
   .version;
 
+// When there's a hotfix, Core may not match current Bolt version
+const currentBoltCoreVersion = require('../../../../../packages/core-v3.x/package.json')
+  .version;
+
 program
   .version(currentBoltVersion)
   .option('-N, --name [name]', 'button')
@@ -54,6 +58,7 @@ module.exports = class extends Generator {
     };
 
     this.boltVersion = currentBoltVersion;
+    this.boltCoreVersion = currentBoltCoreVersion;
 
     if (program.test) {
       this.testData = {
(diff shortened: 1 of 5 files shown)


cjwhitedev:
This solved the problems I encountered. Great work @danielamorse!

packages/generators/yeoman-create-component/generators/component/index.js
```diff
@@ -15,6 +15,10 @@ const { addBoltPackage } = require('./add-bolt-package');
 const currentBoltVersion = require('../../../../../docs-site/package.json')
   .version;
```
colbytcook, line 16: This seems like a very odd syntax, shouldn't we move this to the line above?

packages/generators/yeoman-create-component/generators/component/index.js
```diff
@@ -15,6 +15,10 @@ const { addBoltPackage } = require('./add-bolt-package');
...
 
+// When there's a hotfix, Core may not match current Bolt version
+const currentBoltCoreVersion = require('../../../../../packages/core-v3.x/package.json')
+  .version;
```
colbytcook, line 20: Same comment as above.

packages/generators/yeoman-create-component/generators/component/index.js
danielamorse: @colbytcook this is done by Prettier, which enforces max line lengths.

PR #1905, opened 2020-07-08
DS-123: Scrubber not appearing on embedded video [Commented out Tabs and Navbar VRTs]

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-123

## Summary

Updated the tab styling to prevent BC from hiding controls for videos in a inactive tab at load

## Details

* Copied the . vjs-layout-tiny style block that was being printed in the head and pasted them in the tab.scss partial to control when that style declaration is being rendered (not on tablet or desktop specifically)
* Commented out failing VRT tests in navbar and tabs until we can solve these issues

## How to test
Accordion
1. Navigate to https://boltdesignsystem-oht78pqe9.vercel.app/pattern-lab/?p=components-accordion-content-variations
2. Click on the "Expand to see a video" accordion
3. Observe that the volume and scrubber is **visible** on the Brightcove video after you play it
4. Reduce the device size to a mobile size and refresh the page
5. Again, click on the "Expand to see a video" accordion
6. Observe that the volume and scrubber is **not visible** on the Brightcove video after you play it
7. Resize back to desktop device size and refresh the page

Tabs
1. Navigate to https://boltdesignsystem-oht78pqe9.vercel.app/pattern-lab/?p=components-tabs-content
2. Click on the "A video" tab
3. Observe that the volume and scrubber is **visible** on the Brightcove video after you play it
4. Reduce the device size to a mobile size and refresh the page
5. Again, click on the "A video" tab
6. Observe that the volume and scrubber is **not visible** on the Brightcove video after you play it


mikemai2awesome:
@colbytcook @danielamorse This is working as described, but we need to double check if this is a change we should make. Our CEO has a history of requesting the video player to not hide any core functions. Let's make sure this is the desired outcome first.

danielamorse:
@colbytcook approved.

For more background on this type of Brightcove problem, here's how we solved a similar one: [see commit](https://github.com/boltdesignsystem/bolt/commit/a7af0ed4b).

In this case, the video would never load if it had been initially hidden (in an accordion or tab, for example). The fix was to reinitialize the video when it was revealed. We removed this fix when Brightcove apparently fixed this on their end (it just started working). It looks like while they've fixed the video loading issue, it is still sizes the video incorrectly.

I haven't verified that these are known issues with Brightcove. This is just what I gleaned from my own testing. I agree that at this point the CSS fix is the lesser of two evils.

@mikemai2awesome this change actually prevents controls from being hidden that are supposed to be visible. It fixes an apparent Brightcove bug.

PR #1919, opened 2020-07-16
feature/DS-105: Remove unnecessary package dependency.

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-105](https://pegadigitalit.atlassian.net/browse/DS-105)

## Summary

Removing a package I'm pretty sure we don't need.

## Details

While looking into our automatic release notes for simplifying github labels, Remy came across release-it in our package.json. I spent a little time looking into when it was added and the current consensus is that it's just some accidental nonsense that was never supposed to have made it into the repo. I removed it from the package.json, removed my node_modules directory, and started the setup over again with no issues.

## How to test

- pull this branch
- delete your node_modules folder or otherwise clean our your installation
- do a fresh setup
- ???
-profit


remydenton:
Looks good to me!  Now we cross our fingers that it doesn't affect anything in some secret corner of the release process 🤞

PR #1922, opened 2020-07-20
DS-142: Security Modal: "Unlocking" content with webform

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-142

## Summary

Updated the modal to supported being a gated modal

## Details

- Added a gated modal schema option
- Gated option disables close functionality and hides the close button
- Provides an example for a return button

## How to test

1. Navigate to https://feature-ds-142.boltdesignsystem.com/pattern-lab/?p=components-modal-usage-gated
2. Open the model
3. Observe that there is a link present in the modal labeled "No thanks! Return to last page"
4. Observe there is not a close button
5. Click off the modal
6. Observe that the modal does not close and the cursor does not change to a pointer.

### DIFF
--- a/packages/components/bolt-modal/src/modal.js
+++ b/packages/components/bolt-modal/src/modal.js
@@ -73,7 +73,6 @@ class BoltModal extends withLitHtml {
 
     if (this.open) {
       this.focusTrap.active = true;
-
       this.setFocusToFirstItem(this.renderRoot);
       this.ready = true;
     }
@@ -248,11 +247,9 @@ class BoltModal extends withLitHtml {
       '.c-bolt-modal__content',
     );
 
-    if (!this.persistent) {
-      // If event target contains modal content, assume it is an "outside" click and close
-      if (e.target.contains(modalContent) && !this.persistent) {
-        this.hide();
-      }
+    // If event target contains modal content, assume it is an "outside" click and close
+    if (e.target.contains(modalContent) && !this.persistent) {
+      this.hide();
     }
   }
 
--- a/packages/components/bolt-modal/src/modal.js
+++ b/packages/components/bolt-modal/src/modal.js
@@ -29,12 +29,8 @@ class BoltModal extends withLitHtml {
     theme: props.string,
     scroll: props.string,
     uuid: props.string,
-    gated: props.boolean,
-    // @todo: persistent - this is here to set up the future prop, which is commented out in the schema right now. For now, this will always be false until it's introduced. The same applies for all the other persistent logics below.
-    persistent: {
-      ...props.boolean,
-      ...{ default: false },
-    },
+    persistentReturnUrl: props.string,
+    persistent: props.boolean,
     open: {
       ...props.boolean,
       ...{ default: false },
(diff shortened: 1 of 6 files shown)


packages/components/bolt-modal/src/modal.js
```diff
@@ -29,6 +29,7 @@ class BoltModal extends withLitHtml {
     theme: props.string,
     scroll: props.string,
     uuid: props.string,
+    gated: props.boolean,
     // @todo: persistent - this is here to set up the future prop, which is commented out in the schema right now. For now, this will always be false until it's introduced. The same applies for all the other persistent logics below.
     persistent: {
       ...props.boolean,
```
danielamorse, line 35: Nix this bracketed syntax and use the same syntax as `gated`.

We used this `{...props.boolean, ...}` syntax to set a default when a prop is not officially in the schema (usually an internal-only prop).

mikemai2awesome:
@colbytcook What Dan said. `persistent` is the standard term.

Additionally, I have the following feedback:

1. When hitting the `Esc` key, it needs to act the same as the "return link", since the "return link" is the only way to close the modal.
1. Once the modal is open, focus needs to shift to the "return link", similar to how the focus is shifted to the close button once a regular modal is open.
1. We are missing the activate modal on page load and activate modal after X seconds demos, which is needed for restricted content.
1. Documentation needs work, I can take care of it once all the functions are finalized.

danielamorse:
@mikemai2awesome,

> 1. When hitting the `Esc` key, it needs to act the same as the "return link", since the "return link" is the only way to close the modal.
> 2. Once the modal is open, focus needs to shift to the "return link", similar to how the focus is shifted to the close button once a regular modal is open.
> 3. We are missing the activate modal on page load and activate modal after X seconds demos, which is needed for restricted content.
> 4. Documentation needs work, I can take care of it once all the functions are finalized.

1. Is this for accessibility? I'd assume being able to click the "return" link would suffice?
2. The content of the modal is intentionally not in our control. I told @colbytcook to leave that open so that we weren't locking down a particular design or layout. Instead, we just provide the `persistent` feature. With the current setup, the best we could do would be to ensure the _first_ focusable element in the modal gets focus. Otherwise, it's a lot of rework.
3. The "activate on load" stuff is independent of the `persistent` feature. That works the same for any modal, so I don't think it needs to be demo'd more than once. Good enough to just point the team to the existing demo?

remydenton:
I agree with Mike on disabling the esc key.  Otherwise, you can get around the persistent modal by just hitting escape. 
 Should be a straightforward update here to match what you're already doing for external clicks: https://github.com/boltdesignsystem/bolt/pull/1922/files#diff-e9ed8b7cdeb9041e1d95511e4755cf08R227-R234

I agree with Dan on leaving the content open though.  Whoever is implementing just needs a basic understanding of the best practices-- maybe we provide some suggestions in the documentation.

Final question, is reloading the page the only way to get rid of a persistent modal?  i.e. after filling out the form, you still need the page to reload to dismiss the modal?  I can't think of obvious scenarios where that's not the case, but it seems plausible...

colbytcook:
@remydenton there are bits of this ticket that are going to be handled on the Drupal side (the form showing/closing being one of them). For the UI team work we needed to provide a schema option that would display the close functionality (which was pretty straight forward). The 'Esc' key now activates the return url and the user is brought to that url.

PR #1926, opened 2020-07-27
Feature/ds 166 topside labels

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-166](https://pegadigitalit.atlassian.net/browse/DS-166)

## Summary

Add option to put form input labels before the input.

## Details

Edit the twig and scss to allow for a form input label variation where the label can come before the input. 

## How to test

- Pull this branch
- run yarn start
- check out the new form label position (will link once build is ready)

### DIFF
--- a/packages/components/bolt-form/src/elements/form-element.twig
+++ b/packages/components/bolt-form/src/elements/form-element.twig
@@ -17,25 +17,33 @@ Available variables:
 {% set attributes = attributes.addClass('c-bolt-input-list__item') %}
 
 <div{{ attributes }}>
-  {% if labelDisplay in ['before', 'invisible'] %}
+  {% if labelDisplay == 'before' %}
     {{ label }}
-  {% endif %}
-
-  {{ children }}
-
-  {% if labelDisplay == 'after' %}
+    <div class="c-bolt-block-input-wrapper">
+      {{ children }}
+      {% if errors %}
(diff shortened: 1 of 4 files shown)


cjwhitedev:
ok @mikemai2awesome these requests are resolved. 
> 1. Style the block label. Use small font-size, headline font-family, and xsmall margin bottom.
> 2. Style the validation message. See my comment from JIRA: https://pegadigitalit.atlassian.net/browse/WWW-26?focusedCommentId=10742

packages/components/bolt-form/src/elements/form-element.twig
```diff
@@ -6,7 +6,6 @@ Available variables:
   - labelDisplay (enum) - Where the label should be displayed
       before
       after
-      invisible (Not yet supported, defaults to 'before')
```
remydenton, line 9: I know it says `invisible` is  "not yet supported" here, but technically this pattern still handles that possibility, at least through graceful fallback.  It's also a real option in [the Drupal template this is replacing](https://git.drupalcode.org/project/drupal/-/blob/9.0.x/core/themes/stable/templates/form/form-element.html.twig#L22), which means it's possible someone is using it, maybe even without knowing since this can be selected by content authors in some cases.  I think this would be a little more clear if we had proper schema for this component.

TLDR, I'd like to keep this options both to avoid breaking anything and to remind us about it.  Make sense?

PR #1861, opened 2020-04-30
Migrate Icon component over to LitElement; reduce overall SVG file size by 2.5x

sghoweri:
## Jira
http://vjira2:8080/browse/BDS-2127

## Summary
Switches our Icon component over to render via LitElement + significantly reduce SVG file sizes by updating (and simplifying!) the Icon-portion of our build process.

Updates the Jest snapshots of components with inlined SVG assets to reflect the now smaller file sizes.

## Details
Before our Icon JS bundle clocked in at around `390.7KB` (uncompressed) -- now it's down to around `146KB`! 🎉 

## How to test
- Reference icon docs to confirm everything continuing to work exactly as expected
- Double-check that Icons render correctly in Accordion (since that's one of the few places where Icons render inside the Shadow DOM of the parent component)

PR #1928, opened 2020-07-29
DS-169: Fix broken icons in Micro Journeys

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-169

## Summary

Fix broken icons in Micro Journeys GrapesJS editor.

## Details

The bug was introduced when Icon was refactored to use `BoltElement` as its renderer in #1861.

The underlying issue was not `BoltElement` itself but one of the custom decorator functions used by `BoltElement`.

[The `jsonSchemaProps` function](https://github.com/boltdesignsystem/bolt/blob/master/packages/base-element/src/lib/decorators/json-schema-props.js#L33-L88) loops through the schema and automatically adds schema props to Web Components. It was also adding Twig-only props like `attributes`, `content`, `items`. Everywhere else, having these Twig-only props defined on our Web Components is harmless. However, GrapesJS uses its own `attributes` prop to store data on each element. When the editor loads, GrapesJS tries to write to `element.attributes` which LitElement is already using as a prop. GrapesJS fails to write its data to the element and/or the web component's own internal schema blows up, or some of both.

We intended to [filter out these props](https://github.com/boltdesignsystem/bolt/blob/master/packages/base-element/src/lib/decorators/json-schema-props.js#L42) anyway. So that's what I did and it fixes the issue.

## How to test

To see the bug, go to [Micro Journeys demo on latest release](https://boltdesignsystem.com/pattern-lab/?p=experiments-micro-journeys). Click the "edit" button and notice how the "asset-presentation" icon disappears (it's next to the text, "It’s not mind-reading. It’s signals").

Checkout this branch and view [the same page locally](http://localhost:3000/pattern-lab/?p=experiments-micro-journeys). Verify the icon does not disappear and it persists on save.

### DIFF
--- a/packages/base-element/src/lib/decorators/json-schema-props.js
+++ b/packages/base-element/src/lib/decorators/json-schema-props.js
@@ -33,13 +33,17 @@ const jsonSchemaPropsDecorator = clazz => {
       for (const key in this.schema.properties) {
         let property = this.schema.properties[key];
 
-        // skip any schema properties marked as being deprecated
-        if (
-          !property.title ||
-          (!property.title.includes('deprecated') &&
-            !property.title.includes('DEPRECATED'))
-        ) {
-          // @todo: skip any twig only schema properties such as `attributes`, `content`, `items`
+        // Bolt uses `title` to mark properties as deprecated
+        const isDeprecated = property.title
+          ?.toLowerCase()
+          .includes('deprecated');
+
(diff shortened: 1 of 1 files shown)


remydenton:
Heads up, I also included this as part of the hotfix to 2.26.  I'm also merging it normally here.

PR #1931, opened 2020-08-04
DS-41: Make Accordion transparent  by default

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-41](https://pegadigitalit.atlassian.net/browse/DS-41)

## Summary

Set Accordion default background to transparent

## Details

Make Accordion transparent by default. When box_shadow is set true, the background color will be set to match theme. 

## How to test

Visit the Accordion page on pattern lab.


cjwhitedev:
OK @danielamorse I have resolved the transparency issues. I knew exactly what had happened as soon as you mentioned it. I also set the text color and removed the example page.

PR #1935, opened 2020-08-06
Feature/ds 134 navbar title

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-134](https://pegadigitalit.atlassian.net/browse/DS-134)

## Summary

Set navbar default text and add `hide_title` option.

## Details

Set the navbar default text to read "Go to" if no other text is provided. Only allow that text to be visible under `xsmall` breakpoint unless otherwise specified with `hide_title`. If `hide_title` is set to `from-#{$breakpoint-name}-bp`, only display title text under that breakpoint.

## How to test

- visit the navbar section of pattern library and test different widths.


packages/components/bolt-navbar/navbar.schema.js
```diff
@@ -32,6 +32,13 @@ module.exports = {
             },
           },
         },
+        iconOnly: {
+          type: 'boolean',
+          description:
+            'Set this to true if you do not wish to show title text.',
+          default: false,
+          enum: [true, false],
```
remydenton, line 40: - You don't need the `enum` key for booleans (true and false are always the only possibilities).  
- Since it is new, the prop name should be snake case ("icon_only").
- Did you even mean to call it "icon_only"?  I only proposed that as an example from the button component.  Maybe `hide_title` would be a more descriptive prop?  I'm unsure about this one, just proposing it.
- I think the description would be more accurate if it said, "Set this to true if you do not wish to show title text above the mobile breakpoint."

packages/components/bolt-navbar/src/_navbar-title.twig
```diff
@@ -12,8 +14,8 @@
       } only %}
     </div>
   {% endif %}
-  <{{ tag }} class="c-bolt-navbar__title-text">
-    {{ text }}
+  <{{ tag }} class="c-bolt-navbar__title-text {{ ( not text or iconOnly) ? "c-bolt-navbar__title-text--mobileOnly" : "" }}">
+    {{ text ? text : "Go to" }}
```
remydenton, line 18: - Any text hard coded into templates must be translatable.  You do this using the [t function in twig](https://www.drupal.org/docs/8/api/translation-api/overview#s-translation-in-twig-templates).
- You could make this more concise with the default filter, something like this: `{{ text|default("Go to"|t) }}

packages/components/bolt-navbar/src/_navbar-title.twig
```diff
@@ -1,8 +1,10 @@
 {% set attributes = create_attribute(attributes|default({})) %}
+{% set iconOnly = iconOnly is sameas(true) or iconOnly is sameas(false) ? iconOnly : scheuma.properties.iconOnly.default %}
 {% if url %}
   {% set attributes = attributes.setAttribute("href", url) %}
 {% endif %}
 
+
```
remydenton, line 7: Nitpick: avoiding adding extra blank lines in PRs.

packages/components/bolt-navbar/src/_navbar-title.twig
remydenton: Also, consider using an attributes object for `c-bolt-navbar__title-text` and other classes that should be applied to it.  That would 1) make the logic more clear and 2) potentially avoid you having to update jest tests due to the extra whitespace (not a good reason by itself, but it does draw attention to the `attributes` part).

remydenton:
Good work overall.  I left some comments about feedback to address.

packages/components/bolt-navbar/navbar.schema.js
cjwhitedev: i saw enum set elsewhere, so if we don't need it we should remove it from everywhere.

cjwhitedev:
@remydenton i think I've hit all your points in this last commit. Thanks for the feedback!

packages/components/bolt-navbar/navbar.schema.js
remydenton: @danielamorse , do you know why this is?  I see enums for booleans in a lot of places, like band, image, accordion, etc.  Maybe there is a reason for it... does it have to do with web components?

remydenton:
@colbytcook , good catch on the 420px, we should fix before merging.

I'm seeing the 400-419px issues with the existing component as well (looking at https://boltdesignsystem.com/pattern-lab/?p=viewall-components-navbar), so unless it's a relatively minor change, I'm OK with merging this PR as-is and addressing that separately.

cjwhitedev:
@colbytcook @remydenton I was able to fix it by unifying what breakpoint the mq uses in navbar. some were using xsmall and some where using $bolt-navbar-xsmall-bp which is that 20px difference we were seeing.

docs-site/src/pages/pattern-lab/_patterns/02-components/navbar/05-navbar-hide-title.twig
```diff
@@ -0,0 +1,30 @@
+<p>Setting <code>"hide_title": true</code> will hide the title text on tablet and desktop.</p>
+
+{% include "@bolt-components-navbar/navbar.twig" with {
+  "title": {
+    "tag": "h2",
+    "text": "Title Text",
+    "hide_title": true,
```
mikemai2awesome, line 7: A boolean prop is not scalable, for what breakpoint range to hide the title is dependent on the specific layout. If we set it to `hide_title: from-xsmall-bp,`, we can add more options as the need arises.

`hide_title: true` also means the title will always be hidden, it's not intuitive for this case.

packages/components/bolt-navbar/navbar.schema.js
danielamorse: @remydenton @cjwhitedev I don't think there's a good reason to keep using `enum`. My best guess is that we used it for our demos, so we could loop over the values of Booleans (`[true, false]`) like we do an array of strings.

I'd say, in a separate ticket, remove all the `enum`s and fix any demos or tests that expect `enum`.

cjwhitedev: [Ticket](https://pegadigitalit.atlassian.net/browse/DS-181) Made.

cjwhitedev:
@mikemai2awesome 
@colbytcook 
@remydenton 
**I think I got it all!**

packages/components/bolt-navbar/navbar.schema.js
```diff
@@ -32,6 +32,12 @@ module.exports = {
             },
           },
         },
+        hide_title: {
+          type: 'string',
+          description:
+            'Set the breakpoint at which you wish to hide the title text.',
+          enum: ['xsmall', 'small', 'medium', 'large', 'xlarge'],
```
mikemai2awesome, line 39: I know this is short but people have 2 tracks of thinking, so we need to use the same method as in our mixin. Utilize the `from` and `until` keywords. In this case, we are doing `from`. So the naming should be `from-xsmall-bp` and `from-small-bp`. Let's eliminate the other options and only keep 2 for now. When the need arises, we can add more.

packages/components/bolt-nav-priority/nav-priority.js
```diff
@@ -34,7 +34,6 @@ class BoltNavPriority extends withLitHtml {
   connecting() {
     Promise.all([customElements.whenDefined('bolt-navlink')]).then(_ => {
       this.isOpen = false;
-      this.offsettolerance = 5; // Extra wiggle room when calculating how many items can fit
```
remydenton, line 37: Good find!  There probably was some edge case use for this, but I think you did the right thing by removing it rather than trying to add even more complexity.

packages/components/bolt-navbar/src/_navbar-title.twig
```diff
@@ -25,4 +37,4 @@
   <div class="c-bolt-navbar__title--static">
     {{ navbar_title }}
   </div>
-{% endif %}
+{% endif %}
```
remydenton, line 40: Nitpick, don't forget to add newlines at the end of all your files (and try to review for non-functional changes when committing)

packages/components/bolt-nav-priority/nav-priority.js
cjwhitedev: @remydenton I am glad you agree. Typically I try not to remove something unless it makes total sense to.

packages/components/bolt-navbar/src/_navbar-title.twig
cjwhitedev: @remydenton this should be resolved with the last commit

packages/components/bolt-navbar/src/_navbar-title.twig
```diff
@@ -1,4 +1,16 @@
 {% set attributes = create_attribute(attributes|default({})) %}
+{% set title_attributes = create_attribute({}) %}
+
+{% set schema = bolt.data.components["@bolt-components-navbar"].schema %}
+{% set hideOptions = schema.properties.title.properties.hide_title.enum %}
+{% set hide_title = hide_title in hideOptions ? hide_title %}
+
+{% set title_classes = [
+  'c-bolt-navbar__title-text',
+  hide_title ? "c-bolt-navbar__title-text--hideTitle_#{hide_title}" : ""
```
danielamorse, line 10: I think this should be: `c-bolt-navbar__title-text--hide-title-#{hide_title}`. We don't usually use camelCase in our classnames.

packages/components/bolt-navbar/src/_navbar-title.twig
```diff
@@ -7,7 +7,8 @@
 
 {% set title_classes = [
   'c-bolt-navbar__title-text',
-  hide_title ? "c-bolt-navbar__title-text--hide-#{hide_title}" : ""
+  hide_title ? "c-bolt-navbar__title-text--hide-#{hide_title}" : "",
+  (not hide_title and not text) ?  "c-bolt-navbar__title-text--hide-from-xsmall-bp"
```
danielamorse, line 11:
```suggestion
  (not hide_title and not text) ? "c-bolt-navbar__title-text--hide-from-xsmall-bp"
```

PR #1943, opened 2020-08-12
DS-160: Calculator JS (v2)

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-160

## Summary

Update to how the input type formatting JS is working to be more accessible 

## Details

This update has changes to the HTML, JS and CSS

## How to test

(List clear steps to test that the changes are working)


danielamorse:
@colbytcook, JS is looking good. I just have some suggestions on naming. Some parts of the code took me a minute to understand and I'm hoping renaming a few things will help. I pushed my suggestions to [`feature/DS-160--js-edits`](https://github.com/boltdesignsystem/bolt/commit/90e63bc590783e81ea7453b44141dd4e3043d1e4). Take what you like and feel free to modify as you see fit.

- `inputType` > `inputFormat`, makes it clear we're not talking about the "type", e.g. "checkbox".
- `typeBehaviors` > `onInputChange`, again "type" was ambiguous, input "type" or user "typing"?
- `addThousands` > `formatNumber`, could also be `formatThousands` if it's specific to that piece of formatting.
- `getElementsByClassName(...)[0]` > `querySelector(...)`, the latter returns the first match, no need to get an array and then get the first item.

The last part that was a little confusing was `let thousandsDecimal = [',', '.'];` but I left that alone because I figured you were setting up for future internationalization, and I didn't want to step on your toes. I'd just say if you're going to store specific values, i.e. a "thousands symbol" and a "decimal symbol", it might be better to use key value pairs or variables instead of an array. Something like `formattingSymbols.decimal` would be easier to understand than `formattingSymbols[1]`.

Overall this thing is looking awesome! This is the first time I've seen it with the animation. Nice job guys.

colbytcook:
@danielamorse yea don't look too closely at the currency format it is still a work in progress. I was debating changing the array into an object (the array was a first 'thought' because i didn't know if I would need it for some functionality). I am checking out your changes now, thank you for checking it out.

mikemai2awesome:
We need to strip special characters once the input is out of focus, if the user types with formatting right now such as $25, the error message comes up and says it’s not a number. And update the form doc page.

PR #1952, opened 2020-08-19
Feature/ds 182 blockquote renderer

cjwhitedev:
## Jira

[DS-182](https://pegadigitalit.atlassian.net/browse/DS-182)

## Summary

Convert Bolt-Blockquote to the latest renderer. 

## Details

In an attempt to unify and improve components, I am updating the Blockquote component to use the latest web component renderer. 

## How to test

- Open the blockquote portion of pattern-lab and compare with the origional.
- Review code changes to confirm the conversion is a success.


packages/components/bolt-blockquote/src/blockquote.js
```diff
@@ -1,80 +1,76 @@
...
+    return {
+      ...this.props,
+      authorName: String,
+      authorTitle: String,
+      authorImage: String,
+    };
+  }
+
+  // https://github.com/WebReflection/document-register-element#upgrading-the-constructor-context
+  constructor() {
```
danielamorse, line 35: Remove `constructor()` if we're not doing anything inside it. It will be inherited from the base Class. Note: true of `BoltElement`, not true of `withLitHtml`.

packages/components/bolt-blockquote/src/blockquote.js
```diff
@@ -1,80 +1,76 @@
...
-  customElement,
 } from '@bolt/element';
+
+import classNames from 'classnames/bind';
 import styles from './blockquote.scss';
-import schema from '../blockquote.schema';
 import { AuthorImage, AuthorName, AuthorTitle } from './Author';
-
-let cx = classNames.bind([styles, textStyles]);
+let cx = classNames.bind([styles]);
```
danielamorse, line 13: Put variables after imports please.

PR #1954, opened 2020-08-21
DS-182: Convert Blockquote to latest renderer

cjwhitedev:
## Jira

[DS-182](https://pegadigitalit.atlassian.net/browse/DS-182)

## Summary

Convert Bolt-Blockquote to the latest renderer.

## Details

In an attempt to unify and improve components, I am updating Blockquote to use the latest web component renderer as well as updating, adding, and changing some other relevant concepts.

> Note: Updating [the mutation observer](https://github.com/boltdesignsystem/bolt/pull/1954/files#diff-7cf25cf3960df57948ccf0d192ec1709R34-R55) in the Blockquote component required a [change to Slotify](https://github.com/boltdesignsystem/bolt/pull/1954/files#diff-9314574ce6412c13f9a56d5735da16d9). When building the `slotMap` Slotify now checks to see if an element is already present in the array before adding it. Previously, it would add duplicates if re-run.

## How to test

- Open the blockquote portion of pattern-lab and compare with the origional.
- Review code changes to confirm the conversion is a success.


danielamorse:
@remydenton @cjwhitedev I had to comment out many tests.

We're running into a problem we've seen on other tests, once we upgrade to the new base class. The Promises that tell us the component is ready are not working inside Puppeteer. Since we cannot wait for the component to be ready, render, and return the HTML we want to snapshot, the HTML it returns is essentially the exact same HTML we feed into it. So, there's no point in having these tests. Note: these tests are not VRT - they are the snapshots of Web Components once they've booted up.

The testing helper `renderWC` was supposed to solve this problem for us but it is not working. It times out, never resolving the promise that tells us the component is ready.

I've spent a lot of time, now and in the recent months, debugging this and haven't solved the issue. For now, I'm just commenting out these tests.

PR #1955, opened 2020-08-24
DS-193: fix list wrapping bug in safari

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-193

## Summary

Fixes a bug where inline list pushes its last list item to the next line even if there's enough space to fit all items.

## Details

1. Safari specific CSS added to fix the wrapping issue.
2. CSS is commented stating it's needed for Safari.

## How to test

Run the branch locally and check the list docs. Also, view this [band doc](http://localhost:3000/pattern-lab/?p=components-band-with-pinned-content) and make sure the share tool's last item doesn't get pushed to the next line. It should be visually the same across Safari, Chrome, and Firefox.


mikemai2awesome:
@danielamorse If we can find that case again, then put the overflow hidden util on the element that contains the list, would that fix that one case? As it currently is right now, it breaks in more places than it fixes.

danielamorse:
> @danielamorse If we can find that case again, then put the overflow hidden util on the element that contains the list, would that fix that one case? As it currently is right now, it breaks in more places than it fixes.

I'm pretty sure the use case was on [this Academy page](https://master.boltdesignsystem.com/pattern-lab/?p=pages-acd-completed-training), either the "Sort by: Latest Activity" row or the content inside the horizontal cards. Both have been reworked using grid, so whatever the issue was, it's no longer relevant here.

I leave it up to you. I agree it breaks more than it fixes to leave as is. On the other hand, if we have to add overflow hidden somewhere, tooltips and other positioned content gets complicated.

mikemai2awesome:
@danielamorse That academy page actually is [broken on master in Safari](https://master.boltdesignsystem.com/pattern-lab/?p=pages-acd-completed-training), this fixes it. We have to pick one of these to live with for the time being:

1. Having a 1px overflow which will create horizontal scroll bar
2. Having the last item wrap to the next line even though there's enough space to fit all items.

1 is not easy to recreate, while 2 can happen in a lot of places. I would pick to live with 1 for now.

danielamorse:
@mikemai2awesome I agree. Bug 1 is rare, if it occurs at all in the wild. Let's merge the fix and deal with any 1-type bugs as they arise.

danielamorse (approved):

PR #1958, opened 2020-08-26
DS-144: Accordion deep linking

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-144

## Summary

Add deep linking functionality to Accordion component.

## Details

Deep linking setup is modeled after [same feature on Tabs](https://boltdesignsystem.com/pattern-lab/?p=components-tabs-deep-linking). It uses the query param `?selected-item=ITEM_ID` to deep link to an Accordion item.

- Added `scroll_offset_selector`, and `scroll_offset` props to Accordion
- Added `id` prop to Accordion Item
- Deprecated `uuid` from Accordion Item. It's only used internally for the no-JS functionality, does not need to be customized.

## How to test
- Review code
- Review two new demo pages
- Verify no regressions to other Accordion demos

## Release notes

`uuid` prop is deprecated on Accordion Item. I do not believe this prop is in use on any properties.


cjwhitedev:
The Code looks good. The two new demo pages show this update working. I didn't find any regressions with other demos. 

I think the answer is no, but should this work with JS disabled?
I also wonder if we should call `-item` something less indistinct. for example `-bellows`. I am worried that we will need to deep link another component and have to use `selected-item` elsewhere.

danielamorse:
> I think the answer is no, but should this work with JS disabled?

Good question. No, it's a JS-only feature.

> I also wonder if we should call `-item` something less indistinct. for example `-bellows`. I am worried that we will need to deep link another component and have to use `selected-item` elsewhere.

Also, a good call. Tabs uses `?selected-tab=`. Maybe accordion should use `?selected-accordion-item=`? @mikemai2awesome  you're good naming things - any suggestions?

PR #1963, opened 2020-08-27
Feat/editor tests

remydenton:
Replaces #1718


remydenton:
Tests are broken, which seems to be because they're brittle (and not because anything is actually broken).  Closing.

PR #1965, opened 2020-08-28
Convert Bolt-List to Latest Renderer

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-45](https://pegadigitalit.atlassian.net/browse/DS-45)

## Summary

Convert bolt-list to the latest renderer.

## Details

In an attempt to resolve a race condition issue, convert list to the newest renderer and use `wc-context` for context between the components.

## How to test

To test the conversion:
- Review changes
- Test List demos with and without JS on

@remydenton if you think it makes sense you could test this PR on .COM with the JS patch removed to see if it resolves the `ssr-keep` race condition. But I understand that may be more trouble than it's worth.


packages/components/bolt-list/src/list.js
```diff
@@ -1,91 +1,67 @@
...
-      [`c-bolt-list--spacing-${spacing}`]: spacing !== 'none',
-      [`c-bolt-list--separator-${separator}`]: separator !== 'none',
       [`c-bolt-list--align-${align}`]: align,
-      [`c-bolt-list--valign-${valign}`]: valign,
+      [`c-bolt-list--display-${display}`]: display,
       [`c-bolt-list--inset`]: inset,
       [`c-bolt-list--nowrap`]: nowrap,
+      [`c-bolt-list--separator-${separator}`]: separator !== 'none',
+      [`c-bolt-list--spacing-${spacing}`]: spacing !== 'none',
+      [`c-bolt-list--valign-${valign}`]: valign,
```
remydenton, line 58: +1 for alphabetizing

packages/components/bolt-list/src/list.js
```diff
@@ -1,91 +1,67 @@
...
-    separator: props.string, // none | solid | dashed
-    inset: props.boolean, // true | false
-    nowrap: props.boolean, // true | false
-    align: props.string, // start | center | end
-    valign: props.string, // start | center | end
-  };
+  static get providedContexts() {
+    return {
+      align: { align: schema.properties.align.default },
+      display: { value: schema.properties.display.default },
```
remydenton, line 22: What determines whether you use `value` or the name of the prop (e.g. `align`) as they key in `providedContexts()`?

packages/components/bolt-list/src/list.js
```diff
@@ -1,91 +1,67 @@
...
+    const spacing = this.spacing || schema.properties.spacing.default;
+    const tag = this.tag || schema.properties.tag.default;
+    const valign = this.valign || schema.properties.valign.default;
+
+    this.updateProvidedContext('align', align);
+    this.updateProvidedContext('display', display);
+    this.updateProvidedContext('inset', inset);
+    this.updateProvidedContext('separator', separator);
+    this.updateProvidedContext('spacing', spacing);
+    this.updateProvidedContext('tag', tag);
```
remydenton, line 49: This section seems redundant.  For each property, you set a const using the schema default:
`const align = this.align || schema.properties.align.default;`
 then call this.updateProvidedContext().
`this.updateProvidedContext('align', align);`

What's this doing that the `providedContexts()` method doesn't handle?

remydenton (commented):
Thanks @cjwhitedev .  I'm going to hold off on testing in Drupal for now-- that testing will be easier once this is in a release, and this is good for other reasons (and should be stable)in the meantime.

A few inline comments, mostly questions about how you guys are handling the wc upgrade.

packages/components/bolt-list/src/list.js
danielamorse: Noticed this too! 🙏

packages/components/bolt-list/src/list.js
danielamorse: @remydenton this is an oversight. They should all be `value:` if we're using this approach. Nice catch. However, when I dug into the docs I found a better approach. I'll share below.

packages/components/bolt-list/src/list.js
danielamorse: @remydenton It's making sure that we always pass down either the default value or the latest prop value, never an empty value, so that the consuming component always receives a valid value. `providedContexts()` could be overridden by, say, and empty string if we didn't have a default here in render.

This was the original intent^. Turns out this too is not necessary. I will explain below.

danielamorse:
@remydenton @cjwhitedev I discovered a couple things:

1) `wc-context` can automatically [hook into LitElement props](https://github.com/blikblum/wc-context#lit-element-and-skatejs-integration). This means instead of manually setting defaults, you [set a pointer to your prop](https://github.com/boltdesignsystem/bolt/pull/1965/files#diff-8196554fef1aa69a1e8044f4c1c19d02R19-R28). That's it. You don't need to [manually update context](https://github.com/boltdesignsystem/bolt/pull/1965/files#diff-8196554fef1aa69a1e8044f4c1c19d02L66-L73) in the `render` function. How cool is that!? Sorry @cjwhitedev, wish I'd realized this sooner. We'll have to update some components 😬 
2) `BoltElement` automatically [sets default prop values](https://github.com/boltdesignsystem/bolt/blob/master/packages/base-element/src/lib/decorators/json-schema-props.js#L49-L51) if they're provided in the schema. I might've known this at one point in time but I completely forgot. So, we've been unnecessarily setting defaults in our JS on components that use the new base class.

I think part of the confusion is _because_ we have so many components at various stages of evolution. This just tells me that we're on the right track standardizing these components. @cjwhitedev we can review this in tomorrow's meeting. @remydenton if you have any questions just let me know or we can discuss in dev huddle.

danielamorse (approved):

remydenton:
This looks really good!  I would hit approve and merge except for one more question:

Reading the wc-context docs, shouldn't we be using the lit-element-specific class?  Namely, instead of
[import { withContext } from 'wc-context';](https://github.com/boltdesignsystem/bolt/pull/1965/files#diff-8196554fef1aa69a1e8044f4c1c19d02R3)
we should use this?
[import { withContext } from 'wc-context/lit-element'](https://github.com/blikblum/wc-context#lit-element-and-skatejs-integration)

danielamorse:
> ...shouldn't we be using the lit-element-specific class?

@remydenton yes, I missed this. Updated the code.

Before this fix context worked as expected on initial load, but not on subsequent prop changes. `wc-context/lit-renderer` is required to tell consumers when they should update after that initial load.

Also removed `align` which, as you pointed out, doesn't do anything.

cjwhitedev:
This is some great stuff! Thanks @danielamorse and @remydenton.

remydenton (approved):

PR #1972, opened 2020-09-10
Fix issue with play icon on Safari

adamszalapski:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-209

## Summary

Use `clip-path` to render the play button.

## Details

Below I add a picture showing the before and after effect on Safari.
play-button-glitch-2

I check if the play button is rendered correctly on Safari, Chrome, Edge, Firefox, on iOS, and Android. In all places, the button was rendered correctly. Here is a link to Can I Use showing that this solution should work on all browsers that we support: https://caniuse.com/css-clip-path

## How to test

Go to the video component page and check if the play button is rendered properly especially in Safari but we should check that on all browsers that we support.

## Release notes

(Optionally, add details that will be added to the release notes. Use this to call out deprecations. [Read more](https://intuit.github.io/auto/docs/generated/changelog#additional-release-notes))


mikemai2awesome (approved):
@adamszalapski Works perfectly! Thank you.

PR #1975, opened 2020-09-14
DS-198: Rename `update` to `updateLayout`

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-198

## Summary

Rename `update` to `updateLayout` to prevent conflict with the [Lit Element's `update` method](https://lit-element.polymer-project.org/guide/lifecycle#update).

## Details

We have been using a custom method called `update` on components that need to "update" their layout when a parent component changes state. For example, when an Accordion item that contains a Carousel opens, the Accordion component calls `update` on the Carousel to tell it to resize itself.

LitElement already has a method called `update`, so I'm renaming ours to `updateLayout`. Tabs is the first component where there will be a conflict between the two. So, I'm making this change separately before including the Tabs renderer upgrade changes.

## How to test
- Checkout code locally.
- Go to Accordion demo: `/pattern-lab/patterns/40-components-accordion-40-accordion-content-variations/40-components-accordion-40-accordion-content-variations.html`
- Go to Tabs demo: `/pattern-lab/patterns/40-components-tabs-30-tabs-content/40-components-tabs-30-tabs-content.html`
- On each open the Tabs/Items that contain Accordion, Tabs, and Carousel. Verify this content renders properly and fits into the parent component as it does currently on master (or better, as in the case of Carousel inside Accordion where it fixes a bug).

### DIFF
--- a/packages/components/bolt-accordion/src/accordion.js
+++ b/packages/components/bolt-accordion/src/accordion.js
@@ -180,7 +180,7 @@ class BoltAccordion extends withContext(BoltElement) {
       const elementsToUpdate = this.querySelectorAll('[will-update]');
       if (elementsToUpdate.length) {
         elementsToUpdate.forEach(el => {
-          el.update && el.update();
+          el.updateLayout && el.updateLayout();
         });
       }
     });
--- a/packages/components/bolt-carousel/src/carousel.js
+++ b/packages/components/bolt-carousel/src/carousel.js
@@ -409,6 +409,11 @@ class BoltCarousel extends BoltElement {
     }
   }
 
+  // Public method called by Tabs and Accordion when DOM changes and Carousel needs to update
+  updateLayout() {
+    this.swiper?.update();
+  }
+
   onSlideChange() {
     this.disableSwipingIfAllSlidesAreVisible();
   }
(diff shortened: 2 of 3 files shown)


cjwhitedev (changes requested):
@danielamorse the bug you mentioned with carousel inside a tab or accordion element: do you mean that it should be showing two slides? that's the primary difference i see between this branch and master. 

if that is the intended result, then this has my approval.

danielamorse:
> do you mean that it should be showing two slides? that's the primary difference i see between this branch and master.

@cjwhitedev yep, that's one part of it. Also, you'll notice on master the carousel's pagination dots aren't showing. On master it fixes itself when you resize the window. On the feature branch, it's set up properly as soon as you open the tab/item.

cjwhitedev (approved):

remydenton (approved):
Looks good.  

I still see calls to update() [in accordion, for example](https://github.com/boltdesignsystem/bolt/blob/edb336e513ac8e033d425bdc14832848ae843749/packages/components/bolt-accordion/src/accordion.js#L54), but am assuming those are correctly calling the litElement function, not our custom layout update function.

danielamorse:
> I still see calls to update()...

Yep, there are a handful of cases where we do override the litElement [`update()` function](https://lit-element.polymer-project.org/guide/lifecycle#update). We also use the [`updated()` function](https://lit-element.polymer-project.org/guide/lifecycle#updated) which runs _after_ a prop has updated and the component has re-rendered.

Accordion is actually calling `update()` on the Handorgel (accordion plugin) instance not the renderer.

PR #1976, opened 2020-09-15
DS-45: Add missing 'wc-context' dependencies

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-45

## Summary

Add missing `wc-context` dependency to recently converted components, including List.

## Details

Forgot to add `wc-context` as a dependency, previously import came from core, so this was missed. Adding it now.

## How to test

Review code changes, build is passing.


remydenton (approved):
Looks good, thank you @danielamorse.

PR #1979, opened 2020-09-16
DS-213 Modal Conversion

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-213](https://pegadigitalit.atlassian.net/browse/DS-213)

## Summary

Convert bolt-modal to the latest renderer.

## Details

In an attempt to resolve a race condition issue, convert modal to the newest renderer.

## How to test

To test the conversion:

- Review changes
- Test demos

>Note: The `:focus` behavior is the same as before, i.e. it still takes two tabs to focus on the close button and you still can't use `[autofocus]` to set initial focus on inner modal content. These issues will be handled in https://pegadigitalit.atlassian.net/browse/DS-149.

## Release notes

Removed `no-shadow`, `no_shadow`, `disabled`, and `slots` props from schema. `no-shadow` and `no_shadow` are included automatically by base class. `disabled` is not used. `slots` is not actually a prop, was included as way to document slot names, but does not belong in schema.


danielamorse (approved):
@cjwhitedev nice job! I'll explain a few fixes I made:
- Added contructor back in because that's the place to set initial values, like `this._noBodyScroll` ([see code](https://github.com/boltdesignsystem/bolt/pull/1979/files#diff-e9ed8b7cdeb9041e1d95511e4755cf08R36-R37)).
- Added missing keypress `bind()` method back in ([see code](https://github.com/boltdesignsystem/bolt/pull/1979/files#diff-e9ed8b7cdeb9041e1d95511e4755cf08R53)). Without it `this` is undefined in the keypress callback. The `bind()` methods previously on `show` and `hide` ([see code](https://github.com/boltdesignsystem/bolt/pull/1979/files#diff-e9ed8b7cdeb9041e1d95511e4755cf08L49-L50)) are unnecessary because those aren't used as event handler callbacks, and you were right to remove them.
- I moved a block of code from `firstUpdated()` to `rendered()` because this code needs to run every time the component updates, not just the first time ([see code](https://github.com/boltdesignsystem/bolt/pull/1979/commits/fb2edf81aaadcc2249040e8e5c6d0f310c593299#diff-e9ed8b7cdeb9041e1d95511e4755cf08R62-R66)).
- Lastly, the close button wasn't getting focus on modal open. That's because we call `focus()` before it is ready ([see code](https://github.com/boltdesignsystem/bolt/pull/1979/commits/1e418a6457cab756373066cb26a7265d42fa443c#diff-e9ed8b7cdeb9041e1d95511e4755cf08R289-R292)). More to come on this particular issue in https://pegadigitalit.atlassian.net/browse/DS-149.

colbytcook (approved):

packages/components/bolt-modal/src/modal.js
```diff
@@ -1,57 +1,44 @@
...
-      ...props.boolean,
-      ...{ default: false },
-    },
-  };
+class BoltModal extends BoltElement {
+  static schema = schema;
+
+  static get properties() {
+    return {
+      ...this.props,
```
remydenton, line 27: There are some weird things in the schema for modal.  For example, the following "advanced" props (`hidden: true` in the schema):
`no-shadow`, `no_shadow`, `disabled`

Now that we're using `...this.props`, are those getting through?

I'm also a little foggy on how `persistent_return_url` gets converted to camel case.  The library we're using doesn't show an example of underscore going to camel.  Maybe it's working anyway but just not documented? https://github.com/blakeembrey/change-case/tree/master/packages/camel-case

remydenton: Answered my own question on persistent_return_url-- looks like it gets converted successfully to camel case, even if the documentation is ambiguous about it.

noShadow appears to work the same as for other components without this in their schema.  I guess I'm still not sure what `no-shadow`, `no_shadow`, and `disabled` are doing in this component schema to begin with...

danielamorse: @remydenton good call. I removed `no-shadow`, `no_shadow`, `disabled`, and also `slots` (which should not be a prop). I verified `noShadow` works without including it in the schema because it's [included on the base class](https://github.com/boltdesignsystem/bolt/blob/master/packages/base-element/src/lib/decorators/conditional-shadow-dom.js#L12). The rest just aren't necessary and only confuse things. Removed!

PR #1980, opened 2020-09-17
DS-200: Convert UL, OL, LI to BoltElement renderer

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-200

## Summary

Convert UL, OL, LI to BoltElement renderer.

## Details

This required some rework of all three components to avoid a race condition which appeared after upgrading to the `BoltElement` renderer. The race condition is expected given the way the components were originally written. They used `this.parentNode` to inspect the DOM around the component ([see code](https://github.com/boltdesignsystem/bolt/pull/1980/files#diff-b4e0088fd7d3cbe4bb3cf1df3f8913dcL33-L41)). Sometimes the `parentNode` would be `bolt-li`, sometimes `ssr-keep` depending upon the order the components loaded. For some reason, using the old renderer the race condition was averted, but it appeared with the upgrade and I fixed it by waiting for the closest `bolt-li` to load before proceeding ([see code](https://github.com/boltdesignsystem/bolt/pull/1980/files#diff-b4e0088fd7d3cbe4bb3cf1df3f8913dcR38)).

In the process, I also refactored out of existence the Core helper function `mapWithDepth`. It was only used on UL, OL, and LI, and it's no longer necessary. I searched on GitLab and found no usage downstream.

## How to test

- Review changes files
- Compare UL and OL demos on feature branch to those on latest release. Verify there are no regressions.

## Release notes

Removed helper function `mapWithDepth` from Bolt Core, previously only used internally on UL, OL, LI components.


mikemai2awesome (changes requested):
Probably unrelated to this PR, but can you fix the Component Name on the Unordered List doc page?

remydenton (approved):
Looks great, good find @danielamorse.

cjwhitedev (approved):
@danielamorse so glad you were able to remove a library in the process.

danielamorse:
@mikemai2awesome fixed ([see commit](https://github.com/boltdesignsystem/bolt/pull/1980/commits/f78ad3828f3455109606cfa5453b6e07558bf963#diff-d0d9522e91e4a6ee39f20b0a68b2a02cL1)). Good catch. The whole schema was actually broken.

mikemai2awesome (approved):

PR #1982, opened 2020-09-18
DS-222: Convert Table to latest renderer

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-222](https://pegadigitalit.atlassian.net/browse/DS-222)

## Summary

Convert bolt-table to the latest renderer.

## Details

Convert bolt-table to the latest renderer.

## How to test

To test the conversion:

- Review changes
- Test table demos with and without JS on


danielamorse (approved):
@cjwhitedev made a couple important changes:
- Turn off Shadow DOM ([see commit](https://github.com/boltdesignsystem/bolt/pull/1982/commits/dcb4888187e6678d8711f6d380449dc933562e0c)). `this.useShadow = false` was removed with the constructor. Turns out that is required. Table always renders to the light DOM. Otherwise, the utility classes in [this demo](https://boltdesignsystem.com/pattern-lab/?p=components-table-cell-attributes) don't work because the CSS can't pierce the Shadow DOM. Using the new renderer, the way to turn off shadow DOM completely is via a `static` variable, like we do [in Carousel](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-carousel/src/carousel.js#L56). You can turn Shadow DOM off on an instance of any component via the `no-shadow` attribute, but `static` variable is a clean way to turn it off internally without reflecting the prop as an attribute in the HTML.
- Move code from `firstUpdated()` to `updated()` ([see commit](https://github.com/boltdesignsystem/bolt/pull/1982/commits/350fc09b594bf508073e241dc7269c8d633d772c)). We need this code to run every time the component changes. This is a good litmus test on each component. Change a prop, for instance add or remove the `borderless` prop via the inspector, and see what happens. In this case, it looks ok visually, but the HTML starts to get messed up. You see something like this (note the `data-attrs` attribute which needs to get cleaned up after every render, not just the first time):
```
<th scope="row" class="c-bolt-table__cell c-bolt-table__cell--header" data-attrs="scope=row|class=c-bolt-table__cell c-bolt-table__cell--header">
```

Easy to miss either of these things! But they're important. You'll likely come across this type of thing again in other components.

colbytcook (approved):

PR #1983, opened 2020-09-23
DS-149 Fix Modal focus

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-149

## Summary

Properly set initial focus when modal first opens.

## Details

On open, modal now focuses on 1) the built-in close button 2) the first element with `autofocus` attribute or 3) the first focusable element, in that order.

I updated the persistent modal demo so that the custom close button has `autofocus`.

> Note: @mikemai2awesome in the persistent demo the custom close button _does_ get focus, but the focus style doesn't seem to apply unless you tab onto it. If I add an explicit `.c-bolt-link:focus` style to the Link SCSS that style _does_ work.

## How to test

- Review code changes
- Review Modal demos and verify the expected element gets focus on open.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/40-components/modal/52-modal-usage-persistent.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/40-components/modal/52-modal-usage-persistent.twig
@@ -9,8 +9,7 @@
     inputAttributes: {
       placeholder: "Enter your email address",
       type: "email",
-      required: true,
-      autofocus: true,
+      required: true
     },
     labelDisplayType: "floating"
   } only %}
@@ -68,6 +67,9 @@
           name: "chevron-left",
           position: "before"
         },
+        attributes: {
+          "autofocus": ""
+        }
       } only %}
     {% endcell %}
   {% endgrid %}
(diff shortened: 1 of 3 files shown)


remydenton (approved):
This looks sensible to me, though please don't merge without approval from @mikemai2awesome.

I was going to remark on the following weirdness, but see you already noted it in the PR description.  I'm curious how our link can be focused but not have focus styles 🤔 .  Sadly I have no insight to offer.
> in the persistent demo the custom close button does get focus, but the focus style doesn't seem to apply unless you tab onto it

docs-site/src/pages/pattern-lab/_patterns/40-components/modal/52-modal-usage-persistent.twig
```diff
@@ -68,6 +67,9 @@
           name: "chevron-left",
           position: "before"
         },
+        attributes: {
+          "autofocus": ""
```
mikemai2awesome, line 71: This is not valid HTML, `autofocus` is only supported in form controls such as `<button>`, `<input>`, `<select>` and `<textarea>`.

mikemai2awesome (changes requested):

docs-site/src/pages/pattern-lab/_patterns/40-components/modal/52-modal-usage-persistent.twig
danielamorse: @mikemai2awesome updated with `data-bolt-autofocus` selector.

mikemai2awesome (approved):

PR #1984, opened 2020-09-23
DS-224: Convert Text to latest renderer

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-224](https://pegadigitalit.atlassian.net/browse/DS-224)

## Summary

Convert bolt-text to the latest renderer.

## Details

Convert bolt-text to the latest renderer.

Adds `a` as a tag option to the schema.

## How to test

To test the conversion:

- Review changes
- Test text demos

## Release notes

Deprecated `cite` from bolt-text.


danielamorse (changes requested):
@cjwhitedev looking good. I approve of removing all that commented-out code 👍 

Only regression I found was the "This is the xsmall headline" demo under "Headlines" [here](https://boltdesignsystem.com/pattern-lab/?p=components-text-typographic-recipes). Run locally and compare. It has more letter-spacing currently on master. Pull the latest before you debug. I made some edits:
- Renamed the original `subComponentValues` function to `setDefaultValue` which better describes what it does.
- I added a check for the `url` in the `a` tag switch statement so that `url="undefined"` is never output.

Be sure to mention any meaningful code changes in the PR description. For example, `a` was added to the schema in this PR. That's good to know as a reviewer. Use your judgement, but that change seemed important to me.

You can avoid commits like "for dan's eyes only" 😆 going into git history with the `git commit --amend` flag ([docs](https://www.atlassian.com/git/tutorials/rewriting-history)). I use it all the time. Let me know and we can review it together.

cjwhitedev:
@danielamorse all of these notes have been acted upon besides amending the old commit. Doing so retroactively seems risky.

cjwhitedev:
@danielamorse I just made another update to this while its open. **You had mentioned yesterday that you need to fix some tests before we merge this**, just documenting that here as well.

danielamorse:
To "fix" the failing tests I had to remove `bolt-text` from other component Jest tests: https://github.com/boltdesignsystem/bolt/pull/1984/commits/34f9b16dbb53ef601f458b949aedf53cefdf0835

These tests do not wait long enough for nested components (like `bolt-text`) to finish rendering, so the results it returns are not correct. I spent a bit of time looking for a way to get the tests to wait, but I did not find a solution that works with our current testing setup.

See here for example: https://github.com/boltdesignsystem/bolt/blob/master/packages/testing/testing-helpers/index.js#L53. `customElements` is undefined inside this `renderWC` function. While, it _is_ defined inside the main `test()` function. So, it will never wait for the Promise to resolve.

This requires more time to debug and potentially rebuild our tests. For now, I'm just removing the failing parts of the test.

cc: @colbytcook @remydenton

packages/components/bolt-text/text.schema.js
```diff
@@ -17,7 +17,19 @@ module.exports = {
...
+        'h2',
+        'h3',
+        'h4',
+        'h5',
+        'h6',
+        'p',
+        'div',
+        'span',
+        'cite',
+        'a',
```
mikemai2awesome, line 31: Remove the anchor tag here. Also the `cite`, no one is using that and it doesn't make sense to be here. Only keep headings, paragraph, div, and span.

mikemai2awesome (changes requested):

mikemai2awesome:
Here're the bugs on Safari locally:

1. Tabs disappearing
2. Text color util classes are not working with bolt-text
3. List is not rendering correctly with icons in the item

Screen Shot 2020-10-20 at 11 22 09 AM

packages/components/bolt-text/text.schema.js
cjwhitedev: @mikemai2awesome these have been removed.

docs-site/src/pages/pattern-lab/_patterns/40-components/text/15-text-tag-variations.twig
```diff
@@ -18,4 +18,14 @@
 
     {% include text_code.code_example(text_demo, true) %}
   {% endfor %}
+
+
+{% set anchor_demo %}
+<bolt-text tag="a" url="https://www.google.com">
```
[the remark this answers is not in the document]
cjwhitedev: @mikemai2awesome I have removed this demo for now. We are going to deprecate the use of URL, but do so in a different PR. I made a ticket [here](https://pegadigitalit.atlassian.net/browse/DS-243).

cjwhitedev:
> Here're the bugs on Safari locally:
> 
> 1. Tabs disappearing
> 2. Text color util classes are not working with bolt-text
> 3. List is not rendering correctly with icons in the item

ok @mikemai2awesome this is resolved as we saw in dev huddle. Do you want to just double check that all is well on your system?

mikemai2awesome (approved):

PR #1988, opened 2020-10-06
Bolt version generator

adamszalapski:
## Jira

Without JIRA ticket

## Summary

Fix some logic and add an updated version of the `.incache` file.

## Details

For every release, we should use the correct way to push the changes or if we create only `tag` in GitHub we should update the `.incache` file to have all versions available in our JSON file. We can achieve that by running those commands:
```
npx json -I -f docs-site/.incache -e 'this["bolt-tags"].expiresOn = "2019-06-14T12:30:26.377Z"'
npx json -I -f docs-site/.incache -e 'this["bolt-urls-to-test"].expiresOn = "2019-06-14T12:30:26.377Z"'
yarn setup (or) yarn start
```
I can run `npx now alias boltdesignsystem.com ${tagSpecificUrl} --token=${NOW_TOKEN}` because I have this error after running the command:
```
 Error! You do not have access to the specified account
```
I assume that my `NOW_TOKEN` doesn't have access to Bolt NOW.SH account. @danielamorse maybe you will have more luck on this.

When we create an alias URL for version `2.27.0` we should update the `.incache` file using the commands above.

## How to test

Check if after running this code locally you see `2.26.4` in the bolt version dropdown.
image


danielamorse (approved):
@adamszalapski my token did not work at first. So, I logged into Now.sh/Vercel and generated a new token and tried again with the new token. That time it worked.

I aliased `v2.27.0` to `v2-27-1.boltdesignsystem.com` because they are exactly the same code. There was an error during the `v2.27.0` release so it was either not deployed or not aliased. `v2.27.1` is the same code, successfully released right after.

Retroactively aliasing is actually pretty tricky. Once we lose the context of "latest deploy URL" we have no way to identify what was deployed where. It's just a blob of static code sent to a URL like "boltdesignsystem-fioimjb7t.vercel.app". There is a rough timestamp but no way to be sure what was deployed without a helpful alias. So, this is something to be aware of in the future if we have a failed release.

I updated `.incache` and `v2.27.0` is now showing in the version selector.

colbytcook (approved):

adamszalapski:
Thanks @danielamorse

PR #1991, opened 2020-10-08
DS-152 Page Header Refactor

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-152

## Summary

Adds Page Header component.

## Details

1. Page header component has been properly built out.
2. It is no longer just a PL demo component.
3. Accessibility and ux enhancements are implemented.
4. Will test out Drupal implementation, starting with .com. (@remydenton)

## How to test

Run the branch locally and check all docs and demos under Page Header in Pattern Lab.


packages/components/bolt-page-header/page-header-nav-ul.schema.js
```diff
@@ -0,0 +1,34 @@
+module.exports = {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  title: 'Page Header Nav Ul',
+  type: 'object',
+  // required: ['content', 'category'],
```
remydenton, line 5: This doesn't need to be commented out now since we've [commented out the validation in the twig file](https://github.com/boltdesignsystem/bolt/pull/1991/files#diff-2145f6c793cca3ae115f7ffdf4ce6ec3738bb5131fedc41b4b2b3a948db3696eR1-R9).  The problem isn't that these props aren't required, it's that we haven't yet figured out how to validate multiple sub component files yet.  So we should make the schema correct even if there's no validation happening.

remydenton (changes requested):
I've given most of my feedback previously in dev huddles, Slack, etc. and it looks like it's all been addressed.

One last thing I just noticed, we need to move the header component from [the pattern lab boltrc file](https://github.com/boltdesignsystem/bolt/blob/feature/ds-152-page-header/docs-site/.boltrc.js#L94) to [the starter kit boltrc file](https://github.com/boltdesignsystem/bolt/blob/feature/ds-152-page-header/packages/base-starter-kit/.boltrc.js) since it is no longer just an internal component.

I also added a comment with one nitpick about schema while we're making changes.

remydenton (approved):

danielamorse (approved):

PR #1993, opened 2020-10-09
[DO NOT MERGE] Remove unused '@bolt/build-tools' dependency from micro-journeys

danielamorse:
## Jira

n/a

## Summary

Remove unused `@bolt/build-tools` dependency from Microjourneys, causing warning when building Bolt.

## Details

If a package includes a dependency with `bolt` in the name and that dependency is not included in that environment's `.boltrc` file (or as a dependency of a package in that file), a warning is logged:
```
Warning! Some of the components being compiled have Bolt dependencies and/or peerDependencies that appear to be missing from your build configuration.

To prep for the upcoming Bolt v3.0 release, you'll need to add these packages to the components: { global: [] } section in your /Users/morsd/sites/bolt-backup/docs-site/.boltrc.js file to ensure that Twig namespaces are properly registered, any related Sass and JS code is compiled and loaded, and to have the fastest front-end web performance.

[ '@bolt/build-tools' ]
```

Removing `@bolt/build-tools` from `@bolt/micro-journeys` fixes the warning. `@bolt/build-tools` does not appear to be used anywhere in `@bolt/micro-journeys` so this seems to be a win-win.

Even if `@bolt/build-tools` is required downstream it should be required separately. @remydenton can you think of any reason not to remove this?

## How to test

- Run locally and verify there is no warning on `yarn start`
- Verify micro-journeys works as expected

## Release notes
Unused `@bolt/build-tools` dependency removed from `@bolt/micro-journeys`.


mikemai2awesome (approved):

remydenton:
I believe this was added by @relimck as part of allowing micro journeys to be built as a separate entry point.  The commit is da88caa.

@relimck, can you shed some light on why build tools is required by micro journeys and whether there's a better way around this based on @danielamorse's description?

remydenton:
Should we be adding build tools to the .boltrc file that gets used to do the build on the Drupal side?  That rc file gets created [here](https://gitlab.com/pegadigital/bolt/bolt-release-build/-/blob/master/lib/tasks/get-bolt-package-metadata.js) and doesn't look like it currently includes build tools.  I admit I don't really understand that warning though and the ramifications of adding the build tools to the rc file 🤔 .

danielamorse:
@remydenton, good question. Had to dig into the manifest code to answer it.

No, we shouldn't add `build-tools` to `.boltrc`. [`manifest.js`](https://github.com/boltdesignsystem/bolt/blob/master/packages/build-tools/utils/manifest.js#L45) finds any dependencies with `bolt-` in the name that are missing from `.boltrc`. It auto-adds those dependencies to the manifest, a requirement for Twig namespacing and compiling component JS/CSS. Since `built-tools` isn't a component and doesn't need any of these things, it doesn't need to be in the manifest.

It _does_ make sense to add `build-tools` to the whitelist of packages not to auto-add to the manifest, which would also squash the warning we're seeing. I just made this change ([see commit](https://github.com/boltdesignsystem/bolt/pull/1993/commits/d66187ee29fd547812ac5746d2b8f42af0512d8b)).

Salem [planned to retire](https://github.com/boltdesignsystem/bolt/blob/master/packages/build-tools/utils/manifest.js#L80-L85) this dependency auto-adder in `v3.0`. That's what the warning is for. I agree we should remove it, but prefer to remove it later to not over-complicate `v3.0`.

If there's a good reason to keep `build-tools` in `micro-journeys` that's fine. I just didn't see how it was being used. @relimck if you could point out where it's used, I'd be curious to know 🙏

danielamorse:
Keeping this PR open. Let's test it as part of https://pegadigitalit.atlassian.net/browse/DS-288 later in Q1, Q2.

danielamorse:
Closing in lieu of #2086.

PR #1992, opened 2020-10-09
DS-235 Remove ie11 polyfills and unused packages

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-235

## Summary

Remove IE11 polyfills and unused packages.

## Details
Removed IE11 polyfills from:
- `@bolt/polyfills` (provides only the custom-elements polyfill required by Edge < 79)
- `@bolt/core-v3.x`
- `@bolt/micro-journeys`
- `@bolt/analytics-autotrack`

Specific polyfill dependencies removed:
- `@ungap/weakset`
- `@webcomponents/shadycss`
- `@webcomponents/shadydom`
- `@webcomponents/template`
- `@webcomponents/url`
- `@webcomponents/webcomponentsjs`
- `core-js`
- `custom-event-polyfill`
- `document-register-element`
- `es6-promise`
- `get-own-property-symbols`
- `promise-polyfill`
- `element-closest`
- `mdn-polyfills`
- `smoothscroll-polyfill`
- `whatwg-fetch`
- `regenerator-runtime/runtime`

Deleted the following unused packages:
- `@bolt/core` (aka 'core-v2.x')
- `@bolt/components-critical-fonts`
- `@bolt/components-critical-css`

> Note: these details, along with the other `v3.0` PR details, will be compiled in the release notes and upgrade guide.

## How to test

There are 200+ changed files. The best way to understand the code changes is to look at each commit, where related changes are grouped together in a way that's more digestible than viewing the full diff.

- Review files changed.
- Build is passing
- No regressions (including Edge 18)

## Release notes

Rename any dependencies on `@bolt/core` to `@bolt/core-v3.x`. Remove any use of `@bolt/components-critical-fonts` and `@bolt/components-critical-css`.

PR #1994, opened 2020-10-09
DS-235 Remove IE11 polyfills from PL

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-235

## Summary

Remove IE11 polyfills from the PatternLab build, similar to #1992.

## Details

This PR drops IE support in the Pattern Lab build itself. It has no effect on code downstream, just removes the parts we no longer need.

## How to test

- Check out locally.
- Verify PL builds without issue
- No regressions to PatternLab viewer or custom PL components.


packages/website-ui/uikit-workshop/webpack.config.js
```diff
@@ -53,24 +53,22 @@ module.exports = function(apiConfig) {
     // Allow external flags for modifying PL's prod mode, on top of the .patternlabrc config file
     const config = Object.assign({}, defaultConfig, customConfig, apiConfig);
 
-    function getBabelConfig(isModern = false) {
+    function getBabelConfig() {
```
remydenton, line 56: So the `isModern` flag is going away and we're just keeping whatever you'd get if you set it to true?

remydenton (approved):
All looks pretty sensible.

packages/website-ui/uikit-workshop/webpack.config.js
danielamorse: Exactly, no more legacy vs modern. One single config that is modern (with Edge support for custom elements).

colbytcook (approved):

PR #1999, opened 2020-10-19
DS-156 add text link element and doc page template

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-156

## Summary

Introduces the first Bolt element named Text Link.

## Details

1. Text Link element is added to packages and PL docs.
2. New doc page template created to keep individual pages consistent. The template mandates a particular information architecture.
3. `bolt-button-native-styles-reset` mixin is created for scenarios where all styles need to be stripped for a `<button>` element.
4. `bolt-text-link` mixin is created for scenarios where a regular text link style should be used.

## How to test

Run the branch locally and read through everything under PL > Elements > Text Link. Make sure all the docs make sense. Create a dummy page and try to use the twig include and HTML, ensure all props work as intended.


packages/elements/bolt-text-link/src/text-link.scss
```diff
@@ -0,0 +1,108 @@
...
+
+.e-bolt-text-link--reversed-underline {
+  text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
+.e-bolt-text-link--block {
```
danielamorse, line 48: I love this feature but it's not really "block". What if we had a `block` option that was just `display: block` and another that was called `cover` or `fill` (or something like that)?

packages/elements/bolt-text-link/src/text-link.scss
```diff
@@ -0,0 +1,108 @@
...
+    right: 0;
+    bottom: 0;
+    left: 0;
+  }
+}
+
+/**
+ * These are error messages for when a developer uses <a> and <button> incorrectly with `.e-bolt-text-link`.
+ */
+a.e-bolt-text-link[type],
```
danielamorse, line 62: I'm not sure we should do this.

1) We need to eliminate any CSS that's not absolutely necessary for performance reasons.

2) In the event someone makes a mistake or some errant JS puts the wrong attribute on an element, it will look very unprofessional on a production site.

I know bad accessibility is a serious problem, but I worry putting our validation in CSS and making it publicly visible will come back to haunt us.

packages/elements/bolt-text-link/src/text-link.twig
```diff
@@ -0,0 +1,38 @@
...
+{% set schema = bolt.data.components['@bolt-elements-text-link'].schema %}
+{% if enable_json_schema_validation %}
+  {{ validate_data_schema(schema, _self)|raw }}
+{% endif %}
+
+{# Variables #}
+{% set this = init(schema) %}
+{% set attributes = create_attribute(attributes|default({})) %}
+
+{% if button %}
```
danielamorse, line 24: Could replace this if/else with:
```
{% set tag = button ? 'button' : 'a' %}
```

packages/elements/bolt-text-link/src/text-link.twig
```diff
@@ -0,0 +1,38 @@
...
+
+{% set classes = [
+  'e-bolt-text-link',
+  reversed_underline ? 'e-bolt-text-link--reversed-underline',
+  display == 'block' ? 'e-bolt-text-link--block',
+] %}
+
+{# Template #}
+{# [1], [2] #}
+{% spaceless %}<{{ tag }} {{ attributes.addClass(classes) }}>{% if icon_before %}<span class="e-bolt-text-link__icon-before" aria-hidden="true">{{ icon_before|spaceless }}</span>{% endif %}{{ content }}{% if icon_after %}<span class="e-bolt-text-link__icon-after" aria-hidden="true">{{ icon_after|spaceless }}</span>{% endif %}</{{ tag }}>{% endspaceless %}
```
danielamorse, line 38: This outputs the same thing as far as I can tell (if you consider it an improvement):
```
{% spaceless %}<{{ tag }} {{ attributes.addClass(classes) }}>{% spaceless %}  
  {% if icon_before %}<span class="e-bolt-text-link__icon-before" aria-hidden="true">{{ icon_before|spaceless }}</span>{% endif %}
  {{- content -}}
  {% if icon_after %}<span class="e-bolt-text-link__icon-after" aria-hidden="true">{{ icon_after|spaceless }}</span>{% endif %}
{% endspaceless %}</{{ tag }}>{% endspaceless %}
```

packages/elements/bolt-text-link/text-link.schema.js
```diff
@@ -0,0 +1,50 @@
...
+  type: 'object',
+  required: ['content'],
+  properties: {
+    attributes: {
+      type: 'object',
+      description:
+        'A Drupal-style attributes object with extra attributes to append to this component.',
+    },
+    content: {
+      type: 'string',
```
danielamorse, line 18: Should this be `string` or `object`? `object` would work with both strings and a Drupal render array, right @remydenton?

danielamorse (changes requested):
@mikemai2awesome great work! I'm really liking the simplicity.

The new focus styles are nice, too. On the reversed-underline header example, the focus styles overlap a bit. Are you ok with this? Seems kind of unavoidable with this style. Just pointing it out.
image

I left some inline comments and questions.

docs-site/src/pages/pattern-lab/_patterns/20-elements/text-link/05-text-link.twig
```diff
@@ -0,0 +1,44 @@
+{% set description %}
```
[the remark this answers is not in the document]
mikemai2awesome: Moved it to second.

packages/elements/bolt-text-link/src/text-link.scss
mikemai2awesome: removed

packages/elements/bolt-text-link/text-link.schema.js
mikemai2awesome: changed to `object`.

packages/elements/bolt-text-link/src/text-link.twig
mikemai2awesome: i think it's ok to keep it in one line and use one less spaceless.

mikemai2awesome:
@danielamorse I updated the focus outline, that overlap only happens in Firefox now, looks good in chrome and safari. And the underline only turns gray if it's a visited link.

packages/elements/bolt-text-link/src/text-link.scss
mikemai2awesome: changed to `expand-click-target`

danielamorse (approved):

remydenton:
This looks great.  Nice work @mikemai2awesome.

I'm approving, but I have a philosophical question-- does it really make sense to have twig template be the recommended solution for Drupal?  I personally think we should recommend the bare element, `<a class="e-bolt-text-link"></a>`, in all cases.  Here's why:
* We do officially support the bare element, so the genie is already out the bottle.
* Having two different ways of doing something makes it confusing.  Without looking at the docs, how do you know which one is preferred and which situations to use which one in?  It's one more thing to keep track of for developers who already struggle with Bolt and its quirks.
* We have a strong line in the sand about NOT using `c-bolt-*` elements outside bolt.  The fact that you CAN use `e-bolt-*` elements is already an exception to that, so the clearer we can make the line between the two, the better.  Otherwise, the rule against `c-bolt-*` may not be quite as clear (and that one is very important).
* One major goal of twig is to make components easy to update.  But the core philosophy behind elements is that they are simple, and we officially support their usage outside of twig, so they should be updated very little (i.e. we both shouldn't have much reason to AND it would be a lot of work).  So that reduces the usefulness of twig.
* Another major goal of twig is SSR.  That doesn't apply to elements though.

remydenton (approved):

adamszalapski:
Everything looks good but I have one question because the component section on my end is empty and all components are in the elements section. See image below

image

Except for this everything is ok. I agree with @remydenton that maybe we don't need twig for elements because they should be the simplest part of our design system.

adamszalapski (approved):

mikemai2awesome:
As discussed, we are hiding the elements section for now till we decide on keeping twig template for elements or not.

PR #2000, opened 2020-10-20
Feature/ds 237 move no shadow demos

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-237](https://pegadigitalit.atlassian.net/browse/DS-237)

## Summary

Move all no-shadow demos in pattern lab to their own directory and hide them.

## Details

Move all no-shadow demos in pattern lab to their own directory and hide them. This also involved renaming the expirements directory so that tests could be placed last permanently and hidden according to its position (last).

## How to test

- make sure all no-shadow demos are removed from the components listing
- make sure the tests listing is actually hidden
- verify that the experiments listing isnt broken from the rename
- make sure that the tests still run from their new location

## Release notes

no-shadow testing updated and experiments directory moved.


cjwhitedev:
@mikemai2awesome I wanted to pull you in specifically because of how I handled the hiding of the tests directory, and having to rename the experiments directory. I would love your feedback anywhere you can provide it though.

packages/website-ui/uikit-workshop/src/scripts/components/pl-nav/nav-list.scss
```diff
@@ -36,3 +36,9 @@
   flex-grow: 1;
   width: 100%;
 }
+
+pl-nav > .pl-c-nav__list > .pl-c-nav__list-item {
+  &:nth-last-child(2) {
```
mikemai2awesome, line 41: What is this hiding? @cjwhitedev

mikemai2awesome (changes requested):
@cjwhitedev Thank you for moving these, the organization makes much more sense now. Can you take care of the following?

1. Can we remove the `<form>` tags wrapping each accordion test on accordion-no-shadow.twig? I am not sure why the tags are used.
2. For the experiments section, let's go with 60-experiments.

danielamorse:
> * Can we remove the `<form>` tags wrapping each accordion test on accordion-no-shadow.twig? I am not sure why the tags are used.

@mikemai2awesome `<form>` is a workaround for `no-shadow`. When a component is in a form it renders to Light DOM. You could add the `no-shadow` attribute to every "bolt-accordion" and "bolt-accordion-item" element in the test instead, but I'd just leave the `<form>` tag there for simplicity. Maybe add a comment "components inside form always render to Light DOM".

packages/website-ui/uikit-workshop/src/scripts/components/pl-nav/nav-list.scss
cjwhitedev: @mikemai2awesome the `Tests` section in the pattern lab navigation.

mikemai2awesome (approved):
PL compiled correctly after `yarn build`.

danielamorse (approved):
Nicely done!

PR #2001, opened 2020-10-21
DS-128: Tooltip | a11y | Enable ESC key to close and style no-js version

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-128

## Summary

Updated the tooltip to meet accessibility requirements

## Details

* Added a 'Esc' key event to close the tooltip
* Added functionality to highlight the tooltip trigger

## How to test

* Disable JS and check to see if the border is present
* Tab to a tooltip trigger and press "Esc" to close the flyout

### DIFF
--- a/packages/components/bolt-tooltip/__tests__/__snapshots__/tooltip.js.snap
+++ b/packages/components/bolt-tooltip/__tests__/__snapshots__/tooltip.js.snap
@@ -2,7 +2,7 @@
 
 exports[`<bolt-tooltip> Component Advanced usage: adding tooltip to button 1`] = `
 <bolt-tooltip uuid="12345">
-  <div class="c-bolt-tooltip c-bolt-tooltip--top  ">
+  <div class="c-bolt-tooltip c-bolt-tooltip--top   c-bolt-tooltip--dotted">
     <div role="button"
          tabindex="0"
          aria-describedby="js-bolt-tooltip-12345"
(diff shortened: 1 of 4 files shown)


mikemai2awesome:
@colbytcook I also fixed the no-js CSS, now the tooltip can be seen on focus when js isn't running.

As discussed, the dotted underline should only apply if the trigger content is plain text. This logic is in the JS and not in the twig, that's why the underline doesn't appear when no-js, the `.c-bolt-tooltip--dotted` wasn't added properly based on the trigger content type.

danielamorse (approved):
Verified escape key closes tooltip 👍 

@mikemai2awesome as you mentioned, the dotted line is not possible currently without JS. The best we could do is add the class only when `trigger.type == 'text'`, [see here](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-tooltip/src/tooltip.twig#L73). But this part of the code appears to just be an adaptor for the old way of passing data. It wouldn't work for the most common use cases, where the `type` is not set. I think this will just have to be a JS-only feature. Is that ok?

Also note, there is a 1px shift when the JS sets the bottom border. Is this something easily avoided? We're trying to eliminate all possible layout shift as is negatively affects performance scores: https://web.dev/cls/. 1px isn't much, but if it's easily fixed I'd say let's do it.

danielamorse:
@mikemai2awesome added dotted-underline to all triggers until JS kicks in. Then it only applies to plain-text.

mikemai2awesome (approved):
Thanks @danielamorse. This is good to go. Just need to fix the failing tests.

cjwhitedev (approved):
👌

PR #2003, opened 2020-10-23
DS-178 Fix Blockquote quotes in Safari

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-178

## Summary

Fixes quotes in Safari which were previously not showing.

## Details

Safari doesn't support the CSS selector we were using to add quotation marks. That selector is `::slotted()` combined with a pseudo-selector, example: `::slotted(p):before { }`. I couldn't find any Safari docs specifically saying this isn't supported, but it's what I found after testing. It supports chaining `::slotted()` with other selectors, just no pseudo-selectors.

Given that limitation of Safari, I had to refactor the CSS to use attribute selectors not classes in order to target the blockquote `<p>` tags. In the end this actually reduced a lot of complexity in the CSS _and_ in the JS.

I switch to use `first-of-type` rather than `.is-first-child` classes and eliminated some unnecessary JS in the process.

I also had to refactor the language-specific quotes so that it worked when the `[lang]` attribute was on the blockquote itself _or_ somewhere up the tree.

## How to test
- Review code changes
- Review all [Blockquote demos](https://bugfix-ds-178-blockquote-quotes-in-safari.boltdesignsystem.com/pattern-lab/?p=viewall-components-blockquote) in Safari

### DIFF
--- a/packages/components/bolt-blockquote/src/blockquote.scss
+++ b/packages/components/bolt-blockquote/src/blockquote.scss
@@ -10,18 +10,34 @@ $bolt-blockquote-image-border-color: rgba(bolt-color(gray), 0.2);
 $bolt-blockquote-image-size: 4rem;
 
 @mixin bolt-lang-specific-quotes($type: 'open') {
-  content: map-get-deep($bolt-quotation-marks, 'en', $type);
-
-  [lang='de'] & {
-    content: map-get-deep($bolt-quotation-marks, 'de', $type);
+  $quote-selector: 'p:not([slot]):first-of-type:before';
+  @if ($type == 'close') {
+    $quote-selector: 'p:not([slot]):last-of-type:after';
   }
 
-  [lang='fr'] & {
-    content: map-get-deep($bolt-quotation-marks, 'fr', $type);
+  #{$quote-selector} {
+    content: map-get-deep($bolt-quotation-marks, 'en', $type);
   }
 
-  [lang='ja'] & {
-    content: map-get-deep($bolt-quotation-marks, 'ja', $type);
+  @at-root [lang='de'] &,
+    &[lang='de'] {
+    #{$quote-selector} {
+      content: map-get-deep($bolt-quotation-marks, 'de', $type);
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


packages/components/bolt-blockquote/src/blockquote.scss
```diff
@@ -10,18 +10,34 @@ $bolt-blockquote-image-border-color: rgba(bolt-color(gray), 0.2);
...
 
-  [lang='de'] & {
-    content: map-get-deep($bolt-quotation-marks, 'de', $type);
+  #{$quote-selector} {
+    content: map-get-deep($bolt-quotation-marks, 'en', $type);
   }
 
-  [lang='fr'] & {
-    content: map-get-deep($bolt-quotation-marks, 'fr', $type);
+  @at-root [lang='de'] &,
```
mikemai2awesome, line 22: Are you sure this is working? I have never been able to use `@at-root` with multiple lines. Sass yells at me and I have to do separate rules.

mikemai2awesome (changes requested):

packages/components/bolt-blockquote/src/blockquote.scss
danielamorse: Yeah, this is the output:
image

It barks at me when `@at-root` is not first in the list. Works fine when it is. In this case it _should_ be first in the list. Think I should put it on its own line to be safe? Or leave out `@at-root` altogether?

colbytcook (approved):

packages/components/bolt-blockquote/src/blockquote.scss
mikemai2awesome: @danielamorse OK if it's working then let's keep it.

mikemai2awesome (approved):

danielamorse:
Merging since we have approvals, but just want to remind @cjwhitedev to check out this PR. Thought you might be interested to see how we solved the bug and that we were able to remove some component JS once the fix was in place.

cjwhitedev (approved):
looks great @mikemai2awesome

PR #2009, opened 2020-11-05
DS-250 remove text underline and fix subpixel border issue

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-250

## Summary

Fixes an issue where the action block text is underlined and a bottom border is showing up at random breakpoints.

## Details

1. Set text-decoration to none in CSS.
2. 1.5px is used to fix the subpixel border issue.
3. Fixed a doc where `url` was not passed to an action block.
4. Fixed a small issue with the PL doc frame template.

## How to test

Run the branch locally and view the action blocks docs. Make sure the text is not underlined and a bottom border on the whole container doesn't show up at random breakpoints.

### DIFF
--- a/docs-site/src/components/pattern-lab-utils/docs.twig
+++ b/docs-site/src/components/pattern-lab-utils/docs.twig
@@ -35,14 +35,14 @@
 
 {% set component_last_updated %}
   <bolt-text font-size="xsmall">
-    Last updated: <strong>{{ component_version }}</strong>
+    <span style="white-space: nowrap">Last updated: <strong>{{ component_version }}</strong></span>
   </bolt-text>
 {% endset %}
 
 {% set component_change_log %}
   {% if has_changelog %}
     <bolt-text font-size="xsmall">
-      <bolt-link target="_blank" url="{{ github_url(change_log_file) }}">Change log</bolt-link>
+      <a target="_blank" href="{{ github_url(change_log_file) }}">Change log</a>
     </bolt-text>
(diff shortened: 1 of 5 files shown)


colbytcook (approved):

docs-site/src/components/pattern-lab-utils/docs.twig
```diff
@@ -35,29 +35,29 @@
...
-    Last updated: <strong>{{ component_version }}</strong>
+    <span style="white-space: nowrap">Last updated: <strong>{{ component_version }}</strong></span>
   </bolt-text>
 {% endset %}
 
 {% set component_change_log %}
   {% if has_changelog %}
     <bolt-text font-size="xsmall">
-      <bolt-link target="_blank" url="{{ github_url(change_log_file) }}">Change log</bolt-link>
+      <a target="_blank" href="{{ github_url(change_log_file) }}">Change log</a>
```
danielamorse, line 45: 👍

danielamorse (approved):
@mikemai2awesome nice catch. Verified fixed and approved 👍 

In the future, would it be a good idea to switch to a CSS approach that doesn't rely on overflow hidden and negative margin? CSS grid perhaps like we did [on Academy](https://github.com/boltdesignsystem/bolt/pull/1996/files#diff-8a9c7515507488b016489e88e09592d0ec682ac00d87069dd6691f053712734e)?

PR #2014, opened 2020-11-12
DS-255: Hero band should support vertical/horizontal alignment for the background

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-255](https://pegadigitalit.atlassian.net/browse/DS-255)

## Summary

Add the ability to align background image in hero.

## Details

Add the ability to align background image in hero by creating a new prop and passing a bolt-image into it.

## How to test

Check out last item in the bolt-hero docs


danielamorse (commented):
@cjwhitedev nice job. This works as expected.

One thing I didn't realize until I reviewed the PR is that passing a `custom_background` requires you to set the `cover` and `lazyload` props yourself, or you won't get either of those features in the bg image (and these are required for it to work properly). Using the original `background` prop sets `cover` and `no_lazy` props [by default](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-background/src/background-item.twig#L22-L24).

We either need to document that the user must set these themselves when using `custom_background` or take another approach.

@mikemai2awesome any thoughts? We could always go the `background_valign` route.

mikemai2awesome:
@danielamorse @cjwhitedev Did you walk Tomasz through these updates yet? Please get his feedback first to make sure this will help him or not.

cjwhitedev:
@mikemai2awesome yes, we talked to him.

cjwhitedev:
I talked to Tomasz some more this morning and he's fine with setting defaults because once they are set in a shared module they won't be changed by other developers. I put a little info in the docs about it just to cover the bases and committed.

mikemai2awesome (changes requested):
As @danielamorse said "Using the original background prop sets `cover` and `no_lazy` props by default", we need to specify this in the doc page. That's something you have to do manually when setting `custom_background`.

mikemai2awesome (approved):

danielamorse (approved):
Thanks @cjwhitedev

PR #2015, opened 2020-11-12
DS-157 add button element

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-157

## Summary

Introduces the second Bolt element named Button.

## Details

1. Button element is added to packages and PL docs.
2. Small updates were made to Text Link element to keep the two elements consistent.
3. Added mode specific CSS vars that will eventually replace theming.

## How to test

Run the branch locally and read through everything under PL > Elements > Button. Make sure all the docs make sense. Create a dummy page and try to use the twig include and HTML, ensure all props work as intended.


packages/elements/bolt-button/src/button.scss
```diff
@@ -0,0 +1,279 @@
...
+    cursor: not-allowed;
+
+    &:before,
+    &:after {
+      display: none;
+    }
+  }
+}
+
+$_bolt-button-hierarchy: primary, secondary, tertiary, transparent;
```
danielamorse, line 108: New convention for private SCSS vars?

danielamorse (changes requested):
Really nice work. Button styles look and work great. The docs are lean and mean.

I left a few inline copy suggestions. Take a look and commit if they make sense to you.

I reviewed in Edge 18, which we half-support, and it all looks good with two exceptions. I think these are acceptable since it's still usable:
1. No box-shadow
image
2. File upload is wonky
image

packages/elements/bolt-button/src/button.scss
mikemai2awesome: [Not new](https://github.com/boltdesignsystem/bolt/wiki/Naming-Conventions#variables-mixins-and-functions), we just didn't use it properly elsewhere.

mikemai2awesome:
@danielamorse Good catch. I need to investigate further on the shadows.

docs-site/src/pages/pattern-lab/_patterns/20-elements/button/05-button.twig
```diff
@@ -0,0 +1,43 @@
...
+  Only text should be passed to the <bolt-code-snippet display="inline">content</bolt-code-snippet> prop. However, <bolt-code-snippet display="inline" lang="html">&lt;span&gt;</bolt-code-snippet>, <bolt-code-snippet display="inline" lang="html">&lt;em&gt;</bolt-code-snippet>, and <bolt-code-snippet display="inline" lang="html">&lt;strong&gt;</bolt-code-snippet> elements can be used.
+{% endset %}
+
+{% set twig_markup %}
+{% verbatim %}
+{% include '@bolt-elements-button/button.twig' with {
+  content: 'This is a button',
+  attributes: {
+    type: 'button'
+  }
```
remydenton, line 24: I understand why you did it this way, but I feel like expecting users to pass some things to go through attributes and others to go through props makes it much more confusing.  Even as someone who knows Bolt, I find it unintuitive and a little bit brain-breaking.

Maybe if the pattern gets more solidified with other elements, it will make more sense.  But my knee jerk reaction is to just tell users to always use markup.  I know we've talked about it before, but I just feel like a lot of things keep pointing in that direction for me.

mikemai2awesome: @remydenton Again, here are the main differences

- **Twig**: We document what you MUST do to get the expected results
- **HTML**: We document what you SHOULD do to get the expected results (because we have no control over what you actually do)

Use whatever fits your use case, there's no hard rule about this. When using the Twig, it comes with first class support, we can maintain and update easily. When using the HTML, while we still support it, maintaining and updating won't be as easy compared to Twig (especially if it involves many instances instead of one edge case), because it depends on how the HTML is modified.

We recommend using templates because that's the best way to keep an ecosystem running, and Twig is the official templating engine we use. Again, **this is just a recommendation**, it is still completely up to you to choose whatever path you want to go down with.

As for props vs attributes. All props are related to visual, and all attributes are HTML attributes. It's the case for both text link and button elements right now. And going forward, I do not want any HTML attributes to ever become props. Always treat attributes as you would with HTML. This is clear and intuitive.

cc @colbytcook @danielamorse @cjwhitedev

danielamorse: This is a tough one, but forced to choose I'd say keep the Twig templates and recommend users use the Twig whenever possible.

I think we're very close to a point of diminishing returns by _including_ Twig because it really just provides an abstraction layer for classnames and the fussy parts of the template (including icons for example), but I think these things are still valuable to us and to devs downstream.

Using the attributes object for things that are strictly HTML attributes is a pattern I see subsequent "elements" using.     I like the transparency - I know exactly what it does. I'm not sure other components (vs elements) can readily adopt this convention. In those components you're not interfacing with the element itself, usually a container. But I think it makes sense for elements. Is it ok that elements are not like the others? So far, I'm ok with it. Though, I hear the arguments against it.

I'm approving. @remydenton if you still have reservations, I'd rather wait and talk again when @mikemai2awesome's back. This is not going out with `v2.29` anyway.

danielamorse (approved):

docs-site/src/pages/pattern-lab/_patterns/20-elements/button/05-button.twig
remydenton: > Is it ok that elements are not like the others?

Yes, but it's also a delicate thing.  There's a lot of power and simplicity in just saying "here's the rule, always do it this way" (as we've done with twig embeds/blocks, for instance).  As soon as the rules are different depending on the situation—especially if there are caveats—it becomes a lot harder to understand and follow the rules well (e.g. you can use element markup directly but you should never use component markup; while you _can_ use element markup, we _recommend_ you don't if you don't have to; html attributes are part of `attributes` for elements but use props for components).

So it is ok, it just comes with a cost.  I want us to a) make sure the differences are fully justified and b) be crystal clear about what the differences are so we can communicate it well.

PR #2017, opened 2020-11-16
WWW-384 Play icon on video players floating separately from buttons on Edge v42 (Pega Edge)

colbytcook:
## Jira

## Summary

## Details

Wrapped the "play" icon CSS in @support to serve up two different CSS blocks depending on if the browser supports clip-path

## How to test

View "/pattern-lab/?p=components-video" in Edge v.18 Browserstack (this browser also doesn't support clip-path)


packages/components/bolt-video/src/_edge-video-updates.scss
```diff
@@ -0,0 +1,34 @@
...
+  .vjs-big-play-button {
+    .vjs-icon-placeholder {
+      @supports (
+        not
+          (
+            (-webkit-clip-path: polygon(0 0, 0 100%, 100% 50%)) or
+              (clip-path: polygon(0 0, 0 100%, 100% 50%))
+          )
+      ) {
+        top: 40% !important;
```
adamszalapski, line 11: @colbytcook we really need `!important` here I try to remove it and it works as well.

packages/components/bolt-video/src/_edge-video-updates.scss
```diff
@@ -0,0 +1,34 @@
...
+        left: 40% !important;
```
adamszalapski, line 12: @colbytcook we really need `!important` here I try to remove it and it works as well.

adamszalapski (approved):
It works perfectly. I add two comments in the code.

packages/components/bolt-video/src/_edge-video-updates.scss
colbytcook: ah ok, let me test that, i was having trouble at one point

colbytcook: Edge inspector is strange, the 40% CSS declaration is being overwritten (according to the inspector anyways) but it is not changing the display at all. This is why I put the !important in originally. Well I am removing it now anyways.

packages/components/bolt-video/src/_edge-video-updates.scss
```diff
@@ -0,0 +1,34 @@
+.video-js {
```
mikemai2awesome, line 1: Let's not create a separate file for this. it's really hard to know there are hacks. keep it in the same file and write a comment.

mikemai2awesome (changes requested):

mikemai2awesome (approved):

PR #2018, opened 2020-11-17
WWW-385 Space/padding missing in between Subnav links when set to 'Use Centered Layout' in Edge v42 (Pega Edge)

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/WWW-385

## Summary

Create Edge ~v.42 specific CSS to handle the navlink spacing

## Details

Wrapped the navlink CSS in @support to serve up two different CSS blocks depending on if the browser supports clip-path (this is used to target Legacy Edge)

## How to test

View "/pattern-lab/patterns/40-components-navbar-05-navbar/40-components-navbar-05-navbar.html" in Edge v.17 Browserstack (this browser also doesn't support clip-path)


adamszalapski (approved):
I don't like that we checking if `clip-path` is supported and fixing calc/CSS vars issue but this is temporary and the fix is working that I'm ok with that. Good job @colbytcook 👍

packages/components/bolt-navlink/_edge-bolt-navlink.scss
```diff
@@ -0,0 +1,15 @@
+.c-bolt-navlink {
+  @supports (
+    not
+      (
+        (-webkit-clip-path: polygon(0 0, 0 100%, 100% 50%)) or
```
mikemai2awesome, line 5: Why are we using clip path as the logic?

packages/components/bolt-navlink/navlink.scss
```diff
@@ -140,3 +143,5 @@ bolt-navlink {
     transform: scale3d(1, 1, 1);
   }
 }
+
+@import '_edge-bolt-navlink';
```
mikemai2awesome, line 147: Don't create a separate file, just keep it in the same file with a comment.

mikemai2awesome (changes requested):

mikemai2awesome (approved):

PR #2021, opened 2020-11-24
DS-265 Bolt-card action buttons fix

adamszalapski:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-265

## Summary

Action buttons start to rendered wrong, see image below

image

## Details

Update the code that was pushed with some icons updates.

## How to test

Go to the `/pattern-lab/patterns/40-components-card-replacement-999-card-replacement-with-web-component/40-components-card-replacement-999-card-replacement-with-web-component.html` and check if the buttons are rendered correctly.

image

## Release notes

(Optionally, add details that will be added to the release notes. Use this to call out deprecations. [Read more](https://intuit.github.io/auto/docs/generated/changelog#additional-release-notes))


mikemai2awesome:
Can we figure out why? What was the issue? What about inline grid that's causing this? Is it the same with inline flex? After understanding it, let's comment in the code.

remydenton:
Looks like the core problem is that the styles for `:host()` in `bolt-icon` are intended only for when bolt-icon is the host, but are spilling over to other elements like `<bolt-card-replacement-action>`

From https://developer.mozilla.org/en-US/docs/Web/CSS/:host():
> The :host() CSS pseudo-class function selects the shadow host of the shadow DOM containing the CSS it is used inside (so you can select a custom element from inside its shadow DOM) — but only if the selector given as the function's parameter matches the shadow host.

It seems potentially dangerous that we don't have `:host()` scoped at all.  Wouldn't it be safer if we used, say, `:host(bolt-icon)`?

danielamorse:
@adamszalapski, I think @remydenton's approach, scoping to `:host(bolt-icon)`, is the safest fix to this issue. I push a change to the branch - let us know if you agree.

Yes, @remydenton it is potentially dangerous not to scope `:host` because our components can render those styles to the "light" dom. If our components only ever rendered to the shadow dom it would be fine, but that's not how our components work. We should consider scoping `:host` rules for all WCs.

In the case of icon, it is designed [_only_ to render to "light" dom](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-icon/src/icon.standalone.js#L60), which means when it renders inside `<bolt-card-replacement-action>`'s shadow dom, the `:host` rule applies to `bolt-card-replacement-action`. Same thing happens to Carousel, only without a visual regression.

I don't think we need a `:host` rule at all for Icon because it is never supposed to render to the shadow dom. When it renders inside another component's shadow DOM, the `bolt-icon { }` selector will match it.

To be safe, I suggest scoping it for now. Then in the next release remove the `:host` rule altogether and re-test everywhere.

adamszalapski:
@danielamorse / @remydenton 💯  agree on this :)

remydenton (approved):

danielamorse (approved):

PR #2022, opened 2020-11-25
WWW-384: Fix responsiveness on Edge 42/44 

adamszalapski:
## Jira

## Summary

## Details

Make play icon responsive

## How to test

View "/pattern-lab/?p=components-video" in Edge v.18 Browserstack (this browser also doesn't support clip-path)
Check it on different screen sizes.

### DIFF
--- a/packages/components/bolt-video/src/_videojs-enhancements.scss
+++ b/packages/components/bolt-video/src/_videojs-enhancements.scss
@@ -95,9 +95,8 @@ $bolt-video-js-button-shadow-level-hover: 'level-40';
 
         &:before {
           position: absolute;
-          top: 0;
-          left: 0;
-          transform: translate(-65%, -30%);
+          top: -33%;
+          left: -50%;
           width: 100%;
           height: 100%;
         }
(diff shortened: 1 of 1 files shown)


colbytcook (approved):

packages/components/bolt-video/src/_videojs-enhancements.scss
```diff
@@ -87,6 +84,23 @@ $bolt-video-js-button-shadow-level-hover: 'level-40';
...
+
+        @include bolt-mq(xxlarge) {
+          font-size: 100%;
+        }
+
+        &:before {
+          position: absolute;
+          top: 0;
+          left: 0;
+          transform: translate(-65%, -30%);
```
danielamorse, line 100: @adamszalapski can we not use `translate(-50%, -50%)` here and use top/left `50%`?

danielamorse (approved):
I can't really test this because the `bolt-ms-edge-42-only` mixin doesn't appear to apply to Edge 18, which is the closest thing I have to 42. So, if @colbytcook reviewed and approved, I'm good with it.

@adamszalapski had just one inline comment/question.

packages/components/bolt-video/src/_videojs-enhancements.scss
adamszalapski: @danielamorse ok i can try to figure out the better numbers for top and left 👍

colbytcook: @danielamorse I have confirmed that "bolt-ms-edge-42-only" does work (remember 42 = 17 in Edge terms).

danielamorse: ah! Good call. Thanks for sharing.

adamszalapski: Done

PR #2023, opened 2020-11-25
DS-211: Disable canary releases when merging to master

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-211

## Summary

Disables canary releases every time a PR (or any branch) is merged to master.

## How to test

Since this only affects Travis, I don't know of a way to test other than to merge this to master and then confirm that future merges to master don't create a canary release as part of their build.  A good look at the code is probably the best alternative before merging this.

### DIFF
--- a/auto-release.js
+++ b/auto-release.js
@@ -8,7 +8,6 @@ const { normalizeUrlAlias } = require('./scripts/utils/normalize-url-alias');
 const { branchName } = require('./scripts/utils/branch-name');
 const { NOW_TOKEN } = process.env;
 
-const isCanaryRelease = branchName === 'master';
 const isFullRelease =
   branchName === 'release-2.x' || branchName === 'release/2.x';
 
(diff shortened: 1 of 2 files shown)


auto-release.js
```diff
@@ -8,20 +8,13 @@ const { normalizeUrlAlias } = require('./scripts/utils/normalize-url-alias');
 const { branchName } = require('./scripts/utils/branch-name');
 const { NOW_TOKEN } = process.env;
 
-const isCanaryRelease = branchName === 'master';
 const isFullRelease =
   branchName === 'release-2.x' || branchName === 'release/2.x';
```
adamszalapski, line 12: We should think to add a condition for release 3.

remydenton: Yep, it's definitely on my radar!

colbytcook (approved):

danielamorse (approved):
This looks good to me. Let's test it out on Travis 👍

PR #2025, opened 2020-11-25
DS-245: Links to external sites in Teaser headlines not opening in new tab

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-245](https://pegadigitalit.atlassian.net/browse/DS-245)

## Summary

Allow links in the headline of bolt-teaser to use the target attribute.

## Details

Allowed a new target attribute on bolt-teaser be passed into bolt-headline, and ultimately set a target on bolt-link to allow for opening links in a new tab.

## How to test

There's a demo in bolt-teaser.

## Release notes

Add target option to bolt-teaser.

### DIFF
--- a/packages/components/bolt-headline/headline.schema.js
+++ b/packages/components/bolt-headline/headline.schema.js
@@ -1,3 +1,6 @@
+const elementSchemas = require('@bolt/element/element.schemas');
+const { url, target } = elementSchemas.boltActionElement.properties;
+
 module.exports = {
   $schema: 'http://json-schema.org/draft-04/schema#',
   title: 'Headline',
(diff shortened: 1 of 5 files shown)


packages/components/bolt-headline/headline.schema.js
```diff
@@ -52,6 +52,12 @@ module.exports = {
         'Automatically shrink the font size used in the `xxxlarge` headline size when 60 or more characters are used.',
       default: true,
     },
+    target: {
+      type: 'string',
+      description: '',
+      default: '_self',
+      enum: ['_blank', '_self'],
```
colbytcook, line 59: _parent and _top are also valid options for the target attribute, should we include these in this list as well? https://www.w3schools.com/tags/att_a_target.asp

danielamorse (changes requested):
@cjwhitedev just a couple inline change requests.

I don't think there should be a default `target` value, but since that's what we're doing elsewhere already, what you did makes sense to me. Let's update this to pull in the `BoltActionElement` schema and then in a future breaking change we'll remove the default `target` value there.

Thank you!

cjwhitedev:
@colbytcook @danielamorse the previous commit resolves that issue for us.

danielamorse (approved):
Perfecto 👌 

@colbytcook feel free to merge if this looks good to you.

PR #2026, opened 2020-11-26
Fix play icon

adamszalapski:
## Jira

## Summary

## Details

The previous fix was based on media queries and it works when we have a video with the same dimensions as the page but if we have video as a part of the page fix was not working correctly. This will work in that case now.

### Previous fix:
image
image

### Correct fix:
image
image

## How to test

View "/pattern-lab/?p=components-video" and "/pattern-lab/?p=components-card-replacement-with-video" in Edge v.18 Browserstack (this browser also doesn't support clip-path)
Check it on different screen sizes.


colbytcook (approved):

danielamorse (approved):
Thanks for catching and fixing this @adamszalapski.

I reproduced the bug on `master` and verified it's fixed on the feature branch.

PR #2029, opened 2020-12-03
WWW-421: Do not show remaining video time in accordion or tabs

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/WWW-421

## Summary

Fixes a bug where time remaining shows on videos that are placed within tabs or accordion

## Details

Screen Shot 2020-12-03 at 2 17 26 PM

## How to test

This _should_ be testable in Bolt on the [tabs content](https://boltdesignsystem.com/pattern-lab/?p=components-tabs-content) and [accordion content](https://boltdesignsystem.com/pattern-lab/?p=components-accordion-content-variations) pages, but it's not.  For some reason, videos in those scenarios get the `vjs-layout-medium` class instead of the `vjs-layout-tiny` class that happens on pega.com production, and for which this CSS was originally written.  Maybe there's a race condition of some sort?  @colbytcook and I are stumped 🤷 .

Edit: it appears that videos inconsistently get either `vjs-layout-medium` or `vjs-layout-tiny`.  It's likely some race condition (FYI @danielamorse), but that isn't a problem for this PR, which only cares about the scenario where videos load with `vjs-layout-tiny` class.

### DIFF
--- a/packages/components/bolt-video/src/video.scss
+++ b/packages/components/bolt-video/src/video.scss
@@ -196,6 +196,11 @@ bolt-accordion-item,
 bolt-tab-panel {
   @include bolt-mq($from: medium) {
     .video-js.vjs-layout-tiny:not(.vjs-fullscreen) {
+      // Remaining time should never be shown.  Other controls that shouldn't be forced to show, but sometimes do show--
+      // such as the subs-caps button-- are typically hidden with an !important declaration when they should be hidden.
+      // This seems to be the exception.
+      //
+      // .vjs-remaining-time,
       .vjs-audio-button,
       .vjs-captions-button,
       .vjs-chapters-button,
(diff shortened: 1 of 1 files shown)


danielamorse:
@remydenton it sounds like a race condition, most likely due to the fact that the video is hidden on load and/or when Brightcove JS loads.

Have all the Brightcove updates you and @colbytcook worked on been rolled out? I'd say, let's see where we're at once those changes are in place.

remydenton:
The Brightcove updates haven't been rolled out yet @danielamorse.  I agree we shouldn't bother troubleshooting any further unless it shows up there too (and even then, this CSS should solve this particular case, so it may not even need any more attention unless other related bugs come out of the woodwork).

colbytcook (approved):

PR #2031, opened 2020-12-07
DS-243: Remove URL attribute on bolt-text

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-243](https://pegadigitalit.atlassian.net/browse/DS-243)

## Summary

Remove bolt-text URL since it is not used.

## Details

Remove URL and the tag alterations associated with it. Also remove old commented-out code.

## How to test

Check out the code I removed, and the demos.
Confirm this is not in use anywhere else. 

## Release notes

`url` prop has been removed from the Text component. To create text link going forward, pass Text Link element into the `text` prop.


danielamorse (approved):
Code changes look good 👍 

I searched the repo for other uses of `<bolt-text>` with `url` and found nothing. Also [searched GitLab](https://gitlab.com/search?group_id=5804310&repository_ref=&scope=blobs&search=%3Cbolt-text&snippets=false) and found nothing.

I know you can add and edit `<bolt-text>` in the Microjourneys editor, but `url` is not included in [the editor config](https://github.com/boltdesignsystem/bolt/blob/master/packages/experimental/editor/src/setup-bolt.js#L404-L429), and it doesn't appear [in the demo](https://boltdesignsystem.com/pattern-lab/patterns/60-experiments-editor-10-editor-simple/60-experiments-editor-10-editor-simple.html) (cc @remydenton).

@cjwhitedev, we're really _removing_ the `url` prop. I'd update the PR and release notes from "deprecate" to "remove". When we deprecate a prop, we still support it but flag it for removal in the future. Normally we'd always deprecate before removing, but since this prop never worked I'm fine skipping that step here.

mikemai2awesome (approved):
What @danielamorse said, let's mention that it's completely removed, not just deprecated.

PR #2035, opened 2020-12-11
DS-273 Remove 100% height from icon component

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-273

## Summary

Fixes a bug where the icon height is as tall as page height.

## Details

Removed a line in css that's making the icon container take up 100% height.

## How to test

Run the branch locally and check the follow pages where icons are used most:

1. Icon docs
2. Product pages
3. Academy completed training page

Then create some dummy code to check if it breaks flex display:

```
.container {
    display: flex;
    flex: 0 1 100%;
    flex-direction: column;
}

<div class="container">
  <bolt-icon name="chevron-up" size="large" background="circle" color="yellow"></bolt-icon>
  <bolt-icon name="chevron-up" size="large" background="circle" color="yellow"></bolt-icon>
</div>
```


colbytcook (approved):

remydenton:
I'm not able to reproduce the problem-- when I follow the steps to reproduce on https://pegadigitalit.atlassian.net/browse/DS-273, the icon looks fine.

That said, `height: 100%` doesn't make sense here, it was added in a sloppy/non-specific commit (a9e3bd8), and I don't see any regressions after your fix, so approved ✅ !

remydenton (approved):

PR #2036, opened 2020-12-11
DS-275 Make the background component absolute positioned

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-275

## Summary

Fixes a bug where the background component does not fill up its parent container.

## Details

Moved the absolute position CSS rules to the outer most container which is `<bolt-background>`.

## How to test

Run the branch locally and check the background docs.

Create some dummy code to test the background component inside a CSS grid:

```
.container {
  display: grid;
  grid-template-columns: repeat(3, 1fr);
  position: relative;
  width: 500px;
  height: 500px;
}

<div class="container">
  <!-- use the background component and set an image here -->
  <p>Some text here.</p>
</div>
```

### DIFF
--- a/packages/components/bolt-background/src/background.scss
+++ b/packages/components/bolt-background/src/background.scss
@@ -43,20 +43,24 @@ $bolt-bg-shape-offset-side-large: $bolt-bg-shape-offset-bottom-large * 1.2;
 }
 
 bolt-background {
-  flex-grow: 1;
+  position: absolute;
+  top: 0;
+  right: 0;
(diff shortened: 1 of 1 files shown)


cjwhitedev (approved):
The other contents would still need position set here, but i think thats how background already works, isn't it? Otherwise looks good to me, approved.

mikemai2awesome:
@cjwhitedev Correct, background component is only used inside another positioned non-static element, currently just bands.

PR #2039, opened 2020-12-11
DS-277 Replace divs with spans in tooltip

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-277

## Summary

Fixes a bug where any inline tooltip does not render correctly inside a paragraph of text when JS is turned off.

## Details

Replaced all `div`s in twig template with `span`s, because `div` can not go inside `p` elements.

## How to test

Run the branch locally and check tooltip docs with JS turned off.


colbytcook (approved):

cjwhitedev (approved):
This all looks good to me!

PR #2043, opened 2020-12-21
Unable to tap/select first step in microjourneys experiences on mobile

relimck:
## Jira

WWW-446 - Unable to tap/select first step in microjourneys experiences on mobile

## Summary

Unable to tap/select first step in microjourneys experiences on mobile

## Details

## How to test

* Open any microjourney experience on mobile.
* Switch to any step (but not first)
* Try to switch back to first step


remydenton (approved):
Interesting— I was curious how we would have caused this regression, but it looks like this has been present since the beginning and this is just the first anyone noticed it!

This looks like the right fix.  `Array.prototype.findIndex()` returns `-1` if no element is found, so 0 should be a valid index.  Nice work @relimck.

adamszalapski (approved):
Good job and good catch @relimck

PR #2044, opened 2020-12-21
DS-278 add page padding css vars

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-278

## Summary

Adds page padding CSS vars to global styles.

## Details

1. Responsive CSS vars are created to handle page padding and page max-width.
2. The method of using a wrapper element is no longer needed. Instead, setting padding using these vars will achieve the same effects.

## How to test

Run the branch locally. Create a dummy element and test out the CSS vars, make sure it works as expected.

```css
.test {
  @include bolt-full-bleed;
  padding-right: var(--bolt-page-padding-x);
  padding-left: var(--bolt-page-padding-x);
  background: red;
}
```

### DIFF
--- a/packages/global/styles/00-vars/_vars-page-width-and-spacing.scss
+++ b/packages/global/styles/00-vars/_vars-page-width-and-spacing.scss
@@ -1,0 +1,18 @@
+/* ------------------------------------ *\
+   Page Padding
+
+   Note: the padding alone will create a max-width for the content, there's no need for a wrapper element when using these vars.
+\* ------------------------------------ */
+:root {
+  --bolt-page-padding-x: #{$bolt-wrapper-padding};
+
+  @include bolt-mq(medium) {
+    --bolt-page-padding-x: #{$bolt-wrapper-padding-at-medium-bp};
+  }
+
+  @include bolt-mq(xxlarge) {
+    --bolt-page-padding-x: calc(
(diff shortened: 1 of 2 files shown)


danielamorse (commented):
Interesting approach. Will this eventually replace the wrapper element which uses max-width? My reservation is that it doesn't work so well if it happens to be nested inside another element that also uses the page padding variables or another element with max-width (if you aren't also setting position relative, negative margin, etc). It wouldn't make sense to nest them as it _is_ a page wrapper, but sometimes templates on the Drupal-side end up nesting things in unexpected ways and a plain max-width/wrapper is more resilient to nesting. Just a thought. If this is not a concern for you I'm fine merging.

Also, this prop isn't yet included in the SCSS index file here: `packages/global/styles/index.scss`. Needed to add it there for before using it.

mikemai2awesome:
@danielamorse That's what the full-bleed mixin does, you can nest as much as you want and still make something full bleed, that's how our bands work. Basically all still works the same way but now we can reduce one layer of markup. We'll have to go through our codebase to remove wrappers from bands and maybe some other components, but that doesn't have to be done right away.

danielamorse (approved):

colbytcook (approved):

PR #2046, opened 2020-12-22
DS-271: Support multiple schemas, nested `items` display

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-271

## Summary

Add support for components to have multiple schemas. Handle displaying multiple schemas and nested `items` props in PL docs.

## Details

- Moved the macros from `schema-docs.twig` into a separate file, as the main file was getting huge.
- Added a macro to output the schema table so it could be called recursively.
- Added a check for props with an `items` key _and_ and `items.properties` key*.
- Output nested `items` in a table below the main schema table, before the "Advanced Schema Options".
- Display multiple schemas inside an accordion.
- Fix Accordion bug that prevented Accordion items from opening when inside inactive Tab

> *Note: This change only applies to props whose `items` are properly structured ([see example](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-toc/toc.schema.js#L28-L53)). There are some components which list `items` properties in (what I believe to be) the wrong way ([see example](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-grid/grid.schema.js#L24-L68)). Note: in the second example, the array has no `items` key, it just lists `properties`. We may want to revisit the schemas, but it's outside the scope of this ticket.

## How to test

- Checkout locally and run `yarn setup`.
- See [Accordion schema docs](https://feature-ds-271-schema-docs-updates.boltdesignsystem.com/pattern-lab/?p=viewall-components-accordion) for an example of the new `items` table.
- See [Page Header docs](https://feature-ds-271-schema-docs-updates.boltdesignsystem.com/pattern-lab/?p=viewall-components-page-header) for an example of multiple schemas.
- Check [other schemas](https://feature-ds-271-schema-docs-updates.boltdesignsystem.com/pattern-lab/?p=viewall-components-card) for regressions.
- Review [Accordion bugfix](https://github.com/boltdesignsystem/bolt/pull/2046/files#diff-c6fc58a7cbfb20aa5fa40a25999303b6f88e09deb69ae74a13bab5409ad8e15eR237).


mikemai2awesome:
@danielamorse I updated it with accordion instead of `<details>` because `<details>` is also being used for some hidden props, I want a clear separation from those. I can't get an accordion item to open by default, but that's not a blocker for this, maybe investigate separately why it doesn't work inside Tabs.

mikemai2awesome (approved):

remydenton:
For future reference, try to separate out changes that are just refactoring (moving code between files, changing formatting, etc) from code that makes functional change.  Otherwise, as in commit 26cd25f, it's very difficult to see what changed functionally.

Otherwise, this looks great, nice work.

remydenton (approved):

PR #2047, opened 2020-12-24
DS-212: Review component use of prop.default and context

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-212](https://pegadigitalit.atlassian.net/browse/DS-212)

## Summary

Make sure all components are using providedContext and prop.default in the ideal way.

## Details

Make sure all components are using providedContext and prop.default in the ideal way. As well as some other clean up to align these rendering files. 

## How to test

Check the demos for visual regressions.
Check file changes.


packages/components/bolt-trigger/src/trigger.js
```diff
@@ -13,19 +13,13 @@ let cx = classNames.bind(styles);
 @customElement('bolt-trigger')
 @convertInitialTags(['button', 'a']) // The first matching tag will have its attributes converted to component props
 class BoltTrigger extends BoltAction {
-  static props = {
```
danielamorse, line 16: Can't update props until you update the renderer. Still uses the old version of `BoltAction`. Update as part of https://pegadigitalit.atlassian.net/browse/DS-280.

danielamorse (approved):
@cjwhitedev really nice work. There's a lot of old, confusing code and conventions cleaned up in the PR 🙏 

I left a bunch of inline comments explaining the changes I'm about to push. If anything doesn't make sense we can review tomorrow.

colbytcook (approved):

PR #2048, opened 2020-12-24
DS-181: Remove enums from booleans in schema files

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-181](https://pegadigitalit.atlassian.net/browse/DS-181)

## Summary

Remove enums from booleans in schema files.

## Details

There is no need to define enums for booleans in the schema files. Removing them to avoid redundancy. 

## How to test

Review the code.


remydenton:
This is definitely an improvement-- it's redundant to have an enum for a boolean prop.

Unfortunately, there are some places in our tests and docs site that run a foreach on this enum.  For example, the action blocks test for borderless ([which is failing](https://travis-ci.com/github/boltdesignsystem/bolt/jobs/464661428#L992)) or the [action blocks docs for borderless](https://github.com/boltdesignsystem/bolt/blob/40e1a9052e9bdd5bbcca34d27011e981b139f66c/docs-site/src/pages/pattern-lab/_patterns/40-components/action-blocks/15-action-blocks-borderless-variations.twig#L37) (which looks like it was attempted at one point to remove this).

IMO, those places should be updated to something like `{% for borderless in [true, false] %}`.  But we'll need to do something like that before this can be merged.

remydenton (changes requested):
See comment above.

danielamorse:
@remydenton I addressed your comment, updating demos and Jest tests. Ready for another look.

remydenton (approved):
Looks great, thanks @danielamorse and @cjwhitedev!

PR #2049, opened 2020-12-29
[DO NOT MERGE] DS-283: Update bolt-animate renderer

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-283](https://pegadigitalit.atlassian.net/browse/DS-283)

## Summary

Update animate to the latest renderer standards.

## Details

Update animate to the latest renderer standards.

## How to test

To test the conversion:

- Review changes
- Test animate demos in experiments

### DIFF
--- a/packages/components/bolt-animate/src/animate.js
+++ b/packages/components/bolt-animate/src/animate.js
@@ -1,10 +1,11 @@
-import { classMap, styleMap, html, customElement } from '@bolt/element';
 import {
-  props,
-  hasNativeShadowDomSupport,
-  convertSchemaToProps,
-} from '@bolt/core-v3.x/utils';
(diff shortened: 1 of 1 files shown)


danielamorse (commented):
@cjwhitedev this works perfectly in isolation, but I noticed some issues when used with Micro Journeys. Specifically, some elements start to animate in and then disappear and restart the animation. This smells like a race condition due to the components in MJs still using the old renderer. We have to hold PR one back until we figure out what to do about MJs and WWO, both of which have a dependency on this component.

I think we'll have to update all the components on MJ and WWO. While we're at it, we may want to move `bolt-animate` out of "Bolt" into these individual projects. Talk more tomorrow.

PR #2053, opened 2021-01-06
Brightcove CSS Update

colbytcook:
## Summary

CSS adjustments to fix display bugs with the social panel overlay

### DIFF
--- a/packages/components/bolt-video/brightcove/_video.scss
+++ b/packages/components/bolt-video/brightcove/_video.scss
@@ -203,6 +203,7 @@ $base-video-meta-transition: $bolt-transition;
   }
 
   .vjs-control.vjs-close-button {
+    z-index: 10;
     width: 2.5rem;
     height: 2.5rem;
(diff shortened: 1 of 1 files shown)


remydenton (approved):

remydenton (changes requested):

packages/components/bolt-video/brightcove/_video.scss
```diff
@@ -446,12 +447,11 @@ $base-video-meta-transition: $bolt-transition;
     }
   }
 
-  &.vjs-controls-disabled {
-    pointer-events: none;
-
+  .vjs-controls-disabled {
```
remydenton, line 450: Shouldn't we keep the ampersand?  The `vjs-controls-disabled` class gets added to the same element as `c-base-video`.  Test by removing the `controls` attribute from `<video-js>`

remydenton (approved):

PR #2054, opened 2021-01-06
DS-282 remove bolt-color() and bolt-theme() usage

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-282

## Summary

Removed all usage of `bolt-theme()` function and most of `bolt-color()` usage.

## Details

1. `bolt-theme()` usage has been converted to `var(--m-bolt-*)` light/dark mode vars.
2. `bolt-color()` usage has been converted to `var(--bolt-color-*)` global color vars.
3. Certain `bolt-color()` usage were kept because transparency was used (for example, `rgba(bolt-color(navy), 0.2)`, we will not refactor at this point and leave them as edge cases: icon, progress bar, and table.
4. Note that `var(--m-bolt-*)` and `var(--bolt-color-*)` vars do not work with `rgba()` or `hsla()`. It is encouraged to use brand colors as is without modifying them. 
5. These are not touched because they were either deprecated or being refactored: old card, old video, nav indicator, nav priority, navbar, and navlink.

## How to test

Run the branch locally and compare all pages with master. There should be no visual differences except for the following:

1. Code snippet: the background gray is slightly different, it is now using the brand color.
2. Button: component is styled exactly like button element now.
3. Link: component is styled exactly like text link element now.
4. Table: the background gray is slightly different, it is now using the brand color. The table headers are now accessible and consistent with all other headlines, using the same brand navy color.
5. Typeahead: focus ring color is slightly different, it is now using the brand navy color.
6. Ordered list and numbered headline: the circle gray is slightly different, it is now using the brand color.


packages/core-v3.x/styles/01-settings/settings-global/_settings-global.scss
```diff
@@ -25,24 +25,13 @@ $bolt-fonts--subset-loaded-class: 'js-fonts-subset-loaded' !default;
 /// Async Default font loaded class.
 $bolt-fonts--loaded-class: 'js-fonts-loaded' !default;
 
-// @todo Salem, we should probably deprecate these so there is only one way to do shadows
-// Shadows
-/// Bolt shadow color
-$bolt-shadow-color: bolt-color(black) !default;
-/// Bolt small shadow
-$bolt-shadow--small: 0 2px 0.15rem rgba($bolt-shadow-color, 0.3) !default;
```
danielamorse, line 33: It looks like this _may_ be in use downstream: https://gitlab.com/search?utf8=%E2%9C%93&snippets=false&scope=blobs&repository_ref=&search=%24bolt-shadow-*&group_id=5804310

However, some of these variables are also set again downstream using the same name so it's hard to be sure.

@remydenton @mikemai2awesome have you dealt with this before? It looks like all the variables here might be replicated in the base theme and then inherited by each property, in which case it is not a breaking change to remove them.

packages/components/bolt-icon/src/icon.scss
```diff
@@ -147,12 +147,6 @@
         opacity: bolt-opacity(100);
         background-color: $backgroundColor;
       }
-
-      &.c-bolt-icon--white {
```
danielamorse, line 151: Not required now that we're using css vars?

packages/components/bolt-form/src/form.scss
```diff
@@ -421,46 +418,6 @@ $bolt-input-transition: $bolt-transition;
...
-      background: linear-gradient(
-        to left,
-        $bolt-input-background-color--invalid,
-        rgba(255, 255, 255, 0)
-      );
-    }
-  }
-}
-
-.c-bolt-input.is-disabled,
```
danielamorse, line 453: Are we not using disabled or invalid styles anywhere?

packages/components/bolt-dropdown/dropdown.scss
```diff
@@ -507,58 +434,3 @@ bolt-dropdown {
     )
   );
 }
-
-// Using component-specific class instead of utility class so we don't have to include the whole height utils library
-.c-bolt-dropdown--height-auto {
-  // !important to overwrite inline styles
-  height: auto !important;
-}
```
danielamorse, line 515: We need this class for Dropdown's `collapse` option.

packages/components/bolt-icon/src/icon.scss
mikemai2awesome: White was deprecated a while ago, the css isn't used anymore.

packages/components/bolt-form/src/form.scss
mikemai2awesome: It's on the new line 455, written differently, that's why this is showing deleted.

packages/components/bolt-dropdown/dropdown.scss
mikemai2awesome: Oops, in a sea of commented out code, this got sucked in. Added it back now.

packages/components/bolt-accordion/src/AccordionItem/accordion-item.scss
```diff
@@ -258,7 +231,6 @@ bolt-accordion-item {
   display: block;
   transform: rotate(0deg);
   padding: var(--bolt-spacing-y-xxsmall) 0; // This is for aligning the icon with the text baseline.
-  will-change: transform;
```
danielamorse, line 261: Do you know what all these `will-change` rules were added for?

danielamorse: Nm, I cross-browser tested Accordion and I see no regressions.

packages/core-v3.x/styles/01-settings/settings-global/_settings-global.scss
danielamorse: The [file where this is used](https://gitlab.com/pegadigital/pegapdn/pdn/-/blob/release/release-49/docroot/themes/custom/pdn_theme/packages/components/components-dita/src/components-dita.scss) imports `@pega_bolt_theme/core`, which includes [this file](https://gitlab.com/pegadigital/bolt/pega_core_theme/-/blob/master/src/styles/01-settings/settings-global/_settings-global.scss) where the variable _is_ defined. So I think this is safe to remove.

colbytcook (approved):
Code and VRT looks good

danielamorse:
Reviewed the code and did a VRT pass. Looks good. Merge once it passes.

danielamorse (approved):

PR #2056, opened 2021-01-07
DS-290: Update Constructor on image and link

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-290](https://pegadigitalit.atlassian.net/browse/DS-290)

## Summary

Updated image and link constructors to fit our standards.

## Details

These were out of date and likely not functioning as intended. I updated the constructors on both of these elements to match our modern standards. 

## How to test

Read the code and check patternlab.

### DIFF
--- a/packages/components/bolt-image/src/image.js
+++ b/packages/components/bolt-image/src/image.js
@@ -40,6 +40,11 @@ class BoltImage extends BoltElement {
   static get properties() {
     return {
       ...this.props,
+      // internal use only
+      initialClasses: {
+        type: Array,
+      },
+      // deprecated
       ratio: {
         type: String,
       },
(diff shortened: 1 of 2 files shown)


packages/components/bolt-image/src/image.js
```diff
@@ -143,14 +140,14 @@ class BoltImage extends BoltElement {
         this.lazyImage.addEventListener('lazyloaded', this.onLazyLoaded);
         // `lazySizes.elements` may be undefined on first load. That's ok - the line below is just to catch JS injected images.
 
-        const self = this; // required so checkIfLazySizesReady has the right scope
+        // const self = this; // required so checkIfLazySizesReady has the right scope
```
colbytcook, line 143: Do we need this line commented out or can it be deleted?

colbytcook (approved):

danielamorse (approved):
@cjwhitedev made a few tweaks:
- `initialClasses` _could_ be a prop but because it's only set once on connect, and isn't expected to trigger a re-render, I think it's better as just data on the component.
- See comment about `ratio`, was throwing an error when default wasn't set.
- `const self = this;` is required because inside `checkIfLazySizesReady()` `this` refers to the function not our Web Component. It's not an arrow function so it binds its own `this`.

PR #2059, opened 2021-01-07
Fix: tabs show more button active color

mikemai2awesome:
## Summary

Fixes a color regression on tabs' show more button's active state caused by the color variable conversion.

## Details

Removed a line of CSS that caused the double background color.

## How to test

Run the branch locally and check Tabs doc page. Make the browser narrow and click on the more button of truncated tabs. Make sure it's background color is the same as the hover state's background color.


danielamorse (approved):
Tested locally and it works as expected.

PR #2063, opened 2021-01-08
DS-294: update action-block twig conventions

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-294](https://pegadigitalit.atlassian.net/browse/DS-294)

## Summary

update action-block twig conventions

## Details

action-block still used the old validation so this updates action-block validation to use `{% set this = init(schema) %}`. also used this time to remove `base_class` variable.

## How to test

- test unexpected props and classes.
- visual test demos against master.

## Release notes

Deprecated Action Blocks `children` prop. Use `content` instead.

### DIFF
--- a/packages/components/bolt-action-blocks/src/action-blocks.twig
+++ b/packages/components/bolt-action-blocks/src/action-blocks.twig
@@ -32,7 +32,6 @@
 
 {# Variables #}
 {% set this = init(schema) %}
-{% set attributes = create_attribute(attributes|default({})) %}
 {% set inner_attributes = create_attribute({}) %}
 
 {# Array of classes based on the defined + default props #}
@@ -60,7 +59,6 @@
   {% endif %}
 {% endfor %}
 
-{# Example component's custom element wrapper #}
 <bolt-action-blocks
   {% if outer_classes %} class="{{ outer_classes|join(' ') }}" {% endif %}
   {{ this.props|without("content")|without("class") }}
--- a/packages/components/bolt-action-blocks/src/action-blocks.twig
+++ b/packages/components/bolt-action-blocks/src/action-blocks.twig
@@ -30,6 +30,11 @@
   {% set items = contentItems %}
 {% endif %}
 
+{# DEPRECATED.  `children` has been renamed to `content` #}
+{% if children %}
+  {% set content = children %}
+{% endif %}
+
 {# Variables #}
 {% set this = init(schema) %}
 {% set inner_attributes = create_attribute({}) %}
@@ -64,9 +69,8 @@
   {{ this.props|without("items")|without("class") }}
 >
   <ul {{ inner_attributes.addClass(inner_classes) }}>
-    {# `children` is not in the schema but is in use on WWW, @see 'featured-partners-row.twig'. Cannot safely remove until removed there. #}
-    {% if children %}
-      {{ children }}
+    {% if content %}
+      {{ content }}
     {% else %}
       {% for item in items %}
         <li class="c-bolt-action-blocks__item">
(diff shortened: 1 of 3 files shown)


packages/components/bolt-action-blocks/src/action-blocks.twig
```diff
@@ -4,11 +4,6 @@
   {{ validate_data_schema(schema, _self) | raw }}
 {% endif %}
 
-{# Variables #}
-{% set base_class = "c-bolt-action-blocks" %}
-{% set attributes = create_attribute(attributes|default({})) %}
```
danielamorse, line 9: @cjwhitedev I removed the `attributes` variable which in most cases is not used and is handled by `init()`.

packages/components/bolt-action-blocks/src/action-blocks.twig
```diff
@@ -63,29 +51,25 @@
...
 
-{# Example component's custom element wrapper #}
 <bolt-action-blocks
-  {% if spacing %} spacing="{{ spacing }}" {% endif %}
-  {% if valign %} valign="{{ valign }}" {% endif %}
-  {% if max_items_per_row %} max-items-per-row="{{ max_items_per_row }}" {% endif %}
-  {% if borderless %} borderless {% endif %}
   {% if outer_classes %} class="{{ outer_classes|join(' ') }}" {% endif %}
-  {{ attributes | without("class") }}
+  {{ this.props|without("items")|without("class") }}
```
danielamorse, line 64: @cjwhitedev changed `without("content")` to `without("items")` because `items` is the schema prop we want to filter out, not `content`.

packages/components/bolt-action-blocks/src/action-blocks.twig
```diff
@@ -63,29 +51,25 @@
...
 >
   <ul {{ inner_attributes.addClass(inner_classes) }}>
+    {# `children` is not in the schema but is in use on WWW, @see 'featured-partners-row.twig'. Cannot safely remove until removed there. #}
```
danielamorse, line 67: Added a note here about the `children` prop for the next dev who comes across this. I made [a ticket](https://pegadigitalit.atlassian.net/browse/DS-301) to remove the use of `children` on .COM. It's pretty low priority but wanted to track it. (cc @remydenton)

danielamorse (approved):
@cjwhitedev approved 👍 

I just made a couple changes. See inline comments.

packages/components/bolt-action-blocks/src/action-blocks.twig
cjwhitedev: good to know!

packages/components/bolt-action-blocks/src/action-blocks.twig
cjwhitedev: yeah if its only used in one place hopefully they can update and we can remove.

remydenton: Based on the ongoing goal of keeping components uncoupled, I would actually argue that we should either have `children` or allow for non-structured items, like this:

```
{% for item in items %}
  <li class="c-bolt-action-blocks__item">
    {{ item }}
  </li>
{% endfor %}
```

Does that make sense @danielamorse?

remydenton: Per Dev huddle conversation, the one change to make here is to change 'children' to 'content' to match our conventions.

- Update the twig file to deprecate the existing `children` prop
```
{% if children %}
  {% set content = children %}
{% endif %}
```
```
{% if content %}
  {{ content }}
{% endif %}
```
- Update the schema to add `content` and deprecate `children`

danielamorse:
@cjwhitedev the latest updates look good. Tested locally, works as expected. I'm good to merge.

@remydenton did you want to re-review?

PR #2065, opened 2021-01-08
DS-291 Popover custom width and docs rewrite

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-291

## Summary

Adds a new CSS var for customizing a popover's bubble width.

## Details

1. Component specific CSS var is created for the bubble width.
2. All docs are rewritten with the new template.
3. `boundary` is documented along with `placement`.
4. Updated no-js CSS so that the popover has no width or height when it's not activated.

## How to test

Run the branch locally and go through all the popover docs. Make sure everything works as expected.


adamszalapski (approved):
This is working as expected, great work @mikemai2awesome 🥇

mikemai2awesome:
Waiting for https://github.com/boltdesignsystem/bolt/pull/2073 and then update this.

danielamorse (changes requested):
Awesome docs. Works as expected, except in Edge 44/43 (18/17) where `min()` [isn't supported](https://caniuse.com/css-math-functions):

image

Can we come up with a better fallback for these versions of Edge? It doesn't have to be perfect but maybe a better default width?

mikemai2awesome:
@danielamorse Good catch, let me create a fallback. Damn edge...

mikemai2awesome:
@danielamorse I updated the default, it will fall back to 250px in Edge 42. As for the custom width, it's up to the developer to use `min()` or not, I added a note to the doc specifically stating `min()` does not work in Edge 42 and lower.

danielamorse (commented):
@mikemai2awesome the only browser I have to test this in is Edge 18 (closest available version in Browserstack) and this doesn't work there because the mixin it for Edge 42 only.

Is 42 the only version we care about below 78 or wherever the modern Edge versions begin? Or, do we need to cover 43/44, too?

cc @colbytcook

danielamorse (approved):

PR #2066, opened 2021-01-11
DS-295 remove xsmall icon size

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-295

## Summary

Removes `xsmall` size from Icon component that was added by mistake.

## Details

1. Removed `xsmall` from icon schema.
2. Removed `size` prop value from 2 Academy pages where it's not needed at all.

## How to test

Run the branch locally and check the the 2 Academy pages: Dashboard and Dashboard (My Missions). Make sure the chevron down icon for the decision hub version dropdown is matching the text's font size.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/50-pages/60-academy/03-organisms/collections/dashboard/_dashboard-band--my-missions.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/50-pages/60-academy/03-organisms/collections/dashboard/_dashboard-band--my-missions.twig
@@ -8,7 +8,6 @@
     icon: {
       name: "chevron-down",
       position: "after",
-      size: "xsmall"
     }
   } only %}
 {% endset %}
--- a/docs-site/src/pages/pattern-lab/_patterns/50-pages/60-academy/03-organisms/collections/search/_search-results--results-list.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/50-pages/60-academy/03-organisms/collections/search/_search-results--results-list.twig
@@ -6,7 +6,6 @@
     icon: {
       name: "chevron-down",
       position: "after",
-      size: "xsmall"
     }
   } only %}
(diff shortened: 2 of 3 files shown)


danielamorse (approved):
Works as expected. Good to know the icon will scale proportionately automatically.

I'm in the process of removing `xsmall` from the Academy codebase.

remydenton (changes requested):
Luckily, it looks like `xsmall` isn't being used in the most likely place where it would cause problems: https://gitlab.com/pegadigital/pegapdn/pdn/-/blob/release/release-49/config/sync/field.storage.paragraph.field_icon_size.yml.  It could theoretically be used elsewhere, and the "right" thing to do is deprecate it, but it does feel a lot cleaner to just remove it.

The problem that's still outstanding though (and that led to all this confusion in the first place) is that the icon `size` prop doesn't have a default, and if you don't pass any `size` value, you end up with what looks like an xmall icon.  For example:

```
{% include "@bolt-components-icon/icon.twig" with {
  name: "add-open",
  background: "circle",
  size: "medium"
}%}

{% include "@bolt-components-icon/icon.twig" with {
  name: "add-open",
  background: "circle", 
}%}
```
Gives this:
Screen Shot 2021-01-11 at 3 25 16 PM

Can we just set the default size value to `small`?  Otherwise, we should mark `size` as required, which it really isn't...

remydenton (approved):
As discussed on Slack, the default is actually to use the size of the surrounding text (1em).  Ideally, at least document that, but this is good to merge even if you can't get to that.  Thanks @mikemai2awesome .

danielamorse:
> ...the "right" thing to do is deprecate it, but it does feel a lot cleaner to just remove it.

@remydenton the icon schema change this PR reverts hasn't actually been released yet. So there shouldn't be any conflicts downstream. The Academy edits just proactively remove invalid usage.

PR #2071, opened 2021-01-12
Fix: secondary color variable

mikemai2awesome:
## Jira

N/A

## Summary

Fixes a bug where the secondary color variable is not rendering white in all themes.

## How to test

Run the branch locally and check secondary buttons on Button docs or other demo pages. Make sure the background color of the secondary buttons is white.


danielamorse (approved):
Tested locally, works as expected 🙏

PR #2074, opened 2021-01-14
Fix: remove bolt-button[type] browser defaults

mikemai2awesome:
## Summary

Fixes a bug when `bolt-button[type]` is rendering browser default button styles.

## Details

1. Changed the CSS to specifically target `bolt-button[type]`.
2. We need to avoid using existing HTML attribute name as prop name. This is a prime example of unexpected behaviors.

## How to test

Run the branch locally and find a bolt-button, add `type="submit"` and see if it produces browser default button styles in Safari. Also check the typeahead docs, make sure the search icon button does not look like a browser default button.

### DIFF
--- a/packages/components/bolt-button/src/button.scss
+++ b/packages/components/bolt-button/src/button.scss
@@ -5,6 +5,7 @@
 // Dev Notes
 // 1. [Morse] This mixin outputs a separate ruleset for each selector to prevent IE from failing on unrecognized selectors like `:host`. It is not a substitute for comma-separated selectors.
 // 2. [Morse] Bolt-link delegates focus to inner element.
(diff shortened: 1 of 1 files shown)


danielamorse (approved):

danielamorse:
> We need to avoid using existing HTML attribute name as prop name. This is a prime example of unexpected behaviors.

This issue here is that `[type="submit"]` conflicting with our own global styles ([normalize](https://github.com/boltdesignsystem/bolt/blob/master/packages/global/styles/03-generic/_generic-normalize.scss)) not default browser styles. It might still be a good idea to not have a prop "type". I just want to be clear on where the unexpected side-effects are coming from.

PR #2077, opened 2021-01-15
DS-302: Band twig update

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-302](https://pegadigitalit.atlassian.net/browse/DS-302)

## Summary

update band twig conventions

## Details

band still used the old validation so this updates background validation to use `{% set this = init(schema) %}`. also used this time to remove `base_class` variable and update as many other twig standards as I could manage.

## How to test

- test unexpected props and classes.
- visually test demos against master.


packages/components/bolt-band/band.schema.js
```diff
@@ -79,7 +79,7 @@ module.exports = {
       description:
         "Controls the internal vertical alignment of the band's content.",
       default: 'middle',
-      enum: ['start', 'center', 'end'],
+      enum: ['start', 'center', 'end', 'top', 'bottom', '25%'],
```
danielamorse, line 82: If "middle" is default, it should be included in `enum`.

packages/components/bolt-band/src/band.twig
```diff
@@ -57,17 +39,19 @@
   {% endif %}
 {% endfor %}
 
+{# Deprecated #}
+{% if contentTag %}
```
danielamorse, line 43: I'd move this deprecation block above `init()` with the other deprecations as it was before and simply output `this.data.tag.value` below:
```
<{{ this.data.tag.value }} ...>
```

Unless there's another reason to move it here?

packages/components/bolt-band/src/band.twig
```diff
@@ -57,17 +39,19 @@
...
 {# band component's custom element wrapper #}
 <bolt-band
-  {% if tag %} tag="{{ tag }}" {% endif %}
-  {% if size %} size="{{ size }}" {% endif %}
-  {% if theme %} theme="{{ theme }}" {% endif %}
-  {% if valign %} valign="{{ valign }}" {% endif %}
-  {% if full_bleed %} full-bleed {% endif %}
   {% if outer_classes %} class="{{ outer_classes|join(' ') }}" {% endif %}
-  {{ attributes | without("class") }}
+  {{ this.props|without("content")|without("class") }}
```
danielamorse, line 52: Was going to say add "items" here, but actually `init()` ignores deprecated props. So, just right as is 👍

packages/components/bolt-band/src/band.twig
```diff
@@ -1,44 +1,26 @@
 {% set schema = bolt.data.components["@bolt-components-band"].schema %}
 
-{# {% if enable_json_schema_validation %}
+{% if enable_json_schema_validation %}
   {{ validate_data_schema(schema, _self) | raw }}
-{% endif %} #}
-
-{# Variables #}
-{% set base_class = "c-bolt-band" %}
-{% set attributes = create_attribute(attributes|default({})) %}
```
danielamorse, line 9: Ah ha! We have found one of the exceptions where `attributes` (not just `inner_attributes`) is required. Look at the original Twig file. `attributes` is output twice, once on the outer element and once inside. This is not the way we do it today (we don't allow arbitrary attributes anywhere except the outer element), but for backwards compatibility we must output the attributes inside the component as they were before (see line 115 below).

danielamorse (changes requested):

packages/components/bolt-band/src/band.twig
cjwhitedev: @danielamorse i didnt do a normal depreciation because i cant do {% set this.data.tag.value %} so i had to deprecate this a different way and i mimicked how we did it on background. that said, i can still move this up.

cjwhitedev: actually, moving it before inti() breaks everything.

packages/components/bolt-band/band.schema.js
```diff
@@ -79,7 +79,7 @@ module.exports = {
       description:
         "Controls the internal vertical alignment of the band's content.",
       default: 'middle',
-      enum: ['start', 'center', 'end'],
+      enum: ['middle', 'start', 'center', 'end', 'top', 'bottom', '25%'],
```
mikemai2awesome, line 82: where did all these options come from??? @danielamorse @cjwhitedev

mikemai2awesome (changes requested):

packages/components/bolt-band/band.schema.js
cjwhitedev: @mikemai2awesome when the validator started working I was getting barked at. So those are being passed in somewhere.

cjwhitedev: except middle, we added middle because it was the default

mikemai2awesome: Go to those places and remove or fix them. Look at the CSS and the classes produced in twig, there's only start, center and end. Someone was making mistakes.

mikemai2awesome: Also this prop should not exist, it's completely ignored because of how content is build on Drupal, there's always a grid in a band, the grid handles this. This would just conflict with the grid and produce undesired results.

cjwhitedev: @mikemai2awesome so should we just deprecate it?

cjwhitedev: @mikemai2awesome there was an issue with variables in a background demo. this should be working as intended now.

cjwhitedev: https://github.com/boltdesignsystem/bolt/pull/2077/files#diff-7c7d18d446f59a4185ca5b588e4261b5392871aedf92c77c6d305e418731dc8cR81

mikemai2awesome: Ok, so it was confused with the background component's `valign` prop. Thanks for fixing it.

docs-site/src/pages/pattern-lab/_patterns/40-components/background/35-background-valign-variations.twig
```diff
@@ -16,7 +16,7 @@
         {
           pattern: "image",
           src: "/images/content/backgrounds/background-tall-2.jpg",
-          valign: valign
+          valign: image_valign
```
mikemai2awesome, line 19: The reason that the `valign` was passed incorrectly because this include does not have the `only` keyword. Let's make sure to add `only` in the closing tag. @cjwhitedev

mikemai2awesome (changes requested):

docs-site/src/pages/pattern-lab/_patterns/40-components/background/35-background-valign-variations.twig
cjwhitedev: thanks mike, i added that so im resolving.

cjwhitedev:
@danielamorse @mikemai2awesome I think all thats left is to confirm that the failed tests can be overwritten. They look good to me but I wanted to confirm with someone else.

mikemai2awesome (approved):
@cjwhitedev Since the markup changed because you fixed a few things. You just need to update the tests to make it pass. This is good otherwise.

danielamorse (approved):

PR #2091, opened 2021-02-01
DS-320 H5P-related component updates

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-320

## Summary

Update Band and Stack to no longer be web components. Add Academy-specific component for embedding H5P content.

## Details

H5Ps do not work properly when nested inside an element that's a web component. When the element renders it causes H5P iframe content to disappear. Since Band and Stack don't actually need to be web components, I'm removing the JS. Long-term these will be moved to the Elements/Layouts section. For now, simply removing the JS resolves the H5P issues.

## How to test

- Review files changes.
- Check component demos locally for regressions.

## Release notes

- The `bolt-band.is-ready` selector in [calculator.scss on .COM](https://gitlab.com/pegadigital/pegawww/pegawwwd8/-/blob/release/release-84/docroot/themes/custom/pegawww_theme/packages/components/components-calculator/src/calculator.scss#L33-36) is no longer necessary now that Band is not a web component. Once `2.31.0` is pulled in, you can safely move the animation rules into the parent `span`'s ruleset.


mikemai2awesome (approved):

packages/components/bolt-band/src/band.twig
```diff
@@ -45,10 +45,13 @@
   {% endif %}
 {% endfor %}
 
+{# @todo: remove `.is-ready` selector from calculator.css on .COM. Then, this line can be removed. #}
+{% set outer_classes = outer_classes|merge(['is-ready']) %}
```
remydenton, line 49: Is this worth a release note?  As I understand it, we can't remove [`is-ready` on pega.com](https://gitlab.com/pegadigital/pegawww/pegawwwd8/-/blob/release/release-84/docroot/themes/custom/pegawww_theme/packages/components/components-calculator/src/calculator.scss#L33) before we pull this in, but as soon as we do, we can just remove the `is-ready` class in both places.

docs-site/src/pages/pattern-lab/_patterns/50-pages/60-academy/index.scss
```diff
@@ -242,3 +242,12 @@
...
 }
+
+.c-acd-h5p-embed {
+  @include bolt-shadow('level-20', false, var(--m-bolt-neutral));
+  display: block;
+  padding: var(--bolt-spacing-y-medium) var(--bolt-spacing-x-medium);
+  color: var(--m-bolt-text);
+  border-radius: bolt-border-radius(large);
+  background-color: var(--m-bolt-bg);
+}
```
remydenton, line 253: This file is only used for demo purposes in Pattern Lab, right?  Academy has their own version on their end?

remydenton (approved):
Looks great.  I had a few inline nitpicks/questions, but nothing worth holding this up if you can't get to them

cjwhitedev (approved):

packages/components/bolt-stack/src/stack.twig
```diff
@@ -8,22 +8,15 @@
 {% set this = init(schema) %}
 {% set inner_attributes = create_attribute({}) %}
 
-{% set classes = [
-  'c-bolt-stack',
-] %}
-
 {#
```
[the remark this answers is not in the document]
danielamorse: It is still the convention for web components. Since Stack is no longer a WC I'm removing the unnecessary layer of HTML. I'm also removing it so we have parity between what Twig outputs and what `<bolt-stack></bolt-stack>` outputs, which is basically nothing but a custom element and the inner content. Stack will probably be moved to "Layouts" where similar conventions are in use.

packages/components/bolt-band/src/band.twig
danielamorse: 💯 , updated

docs-site/src/pages/pattern-lab/_patterns/50-pages/60-academy/index.scss
danielamorse: Yep, used PL to workout the Academy styles. I will copy these into the Academy repo as a theme-specific component.

PR #2092, opened 2021-02-02
DS-321 Update Slotify to handle only whitespace as content

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-321

## Summary

If a component initially contains only whitespace, content added later never appears. This PR updates Slotify to handle this case, which happens on [Collaboration center](https://pegadigitalit.atlassian.net/browse/CC-416).

## Details

Add some conditional logic to handle the case where there _are_ child nodes, but they contain only whitespace. Ensures a `<slot>` is added when that is true, as is the case when you use the Modal Twig template and pass `content: ''` - Twig adds line breaks.

## How to test

- To reproduce the bug, go to [modal demo on `2.30.2` site](https://v2-30-2.boltdesignsystem.com/pattern-lab/patterns/40-components-modal-05-modal/40-components-modal-05-modal.html) or do the same on `master` locally.
- Via Chrome dev tools delete the `<bolt-modal>` elements on the page.
- Enter the following via console:
```
const modal = document.createElement('bolt-modal');
modal.innerHTML = ' ';
document.body.append(modal);
```
- Then (important you do this is two steps not one):
```
modal.innerHTML = 'test';
modal.show();
```
- Verify the modal opens up, is empty and does not contain "test" text content.
- Then, checkout this feature branch and repeat locally to verify the content is properly added.
- Spot check other components to be sure there are no regressions to other web components.

### DIFF
--- a/packages/base-element/src/Slotify.js
+++ b/packages/base-element/src/Slotify.js
@@ -25,6 +25,7 @@ export class Slotify extends LitElement {
   }
 
   // Save a reference to the pseudoSlot content before lit-element renders
+  // 1. Always add at least a default slot to the slotMap. Otherwise, content added after the initial render will not appear.
   saveSlots() {
     const childNodes = Array.from(this.childNodes);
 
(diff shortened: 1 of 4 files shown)


cjwhitedev (approved):
works well!

PR #2098, opened 2021-02-04
DS-313: Fix button's use of the style prop

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-313](https://pegadigitalit.atlassian.net/browse/DS-313)

## Summary

Deprecate button style prop

## Details

Button uses a prop that should actually be twig-only. Deprecate the style prop and replace it with hierarchy.

## How to test

Do a visual regression test against master. 

## Release notes

button prop `style` is deprecated, please use `hierarchy` instead.


packages/components/bolt-button/src/button.js
```diff
@@ -23,7 +23,7 @@ class BoltButton extends BoltActionElement {
   static get properties() {
     return {
       ...this.props,
-      color: { type: String },
```
danielamorse, line 26: Unfortunately, I don't think we can remove `color` as the web component version of button with the `color` prop is used in the wild. [Academy homepage](https://academy.pega.com/) is one example. It may also be used with Get Smart Content (?). It's also used (at least in name) on [Microjourneys](https://github.com/boltdesignsystem/bolt/blob/master/packages/experimental/editor/src/setup-bolt.js#L363).

To be safe, we should support both `color` and `hierarchy` in the web component. Let's talk tomorrow in our meeting about how to do this.

danielamorse (commented):
@cjwhitedev this looks good for the most part, but see my comment about removing `color` in the web component. I may take this one if you want to focus on addressing the Card feedback.

packages/components/bolt-button/src/button.js
danielamorse: Update: added `color` prop back in with deprecation note.

danielamorse (approved):
Created follow-up task to remove `color` from the GrapeJS editor: https://pegadigitalit.atlassian.net/browse/DS-329

mikemai2awesome (approved):

PR #2099, opened 2021-02-06
Hotfix/DS-238 Babel loader update, merge in 2.x

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-238

## Summary

Update webpack config so that `@bolt` packages located in `node_modules` dir are run through `babel-loader`. Merge in latest code from `release/2.x` branch.

## Details

When individual Bolt packages are imported into theme components on Pega sites, they are downloaded to the local `node_modules` dir. This dir is currently excluded from the `babel-loader` in Bolt's [webpack config](https://github.com/boltdesignsystem/bolt/pull/2099/files#diff-c11127c8eba4870da6a15986f41a841277140509fcea414131e52f33ae0fbdbcL614) to avoid the expensive operation of running _all_ node modules through Babel. This PR makes an exception for `@bolt` packages so that those packages will be properly transpiled when included individually downstream. Before updating this rule, the with-without component was throwing an error when it tried to import `@bolt/element` or any other package that used ES6 syntax.

Also, I merged in the latest work from `release/2.x` so this branch should now include all of the latest work from both branches.

## How to test

- Build is passing
- Review [webpack update](https://github.com/boltdesignsystem/bolt/pull/2099/files#diff-c11127c8eba4870da6a15986f41a841277140509fcea414131e52f33ae0fbdbcR457)


remydenton (approved):
This looks good!  That last commit makes sense at least 😄 .

Since this is more or less a no going back situation, I'm holding off on merging just yet-- maybe we can do so after a double check in dev huddle tomorrow, then get pega.com set up for full regression testing Monday night.

danielamorse:
> Since this is more or less a no going back situation, I'm holding off on merging just yet-- maybe we can do so after a double check in dev huddle tomorrow, then get pega.com set up for full regression testing Monday night.

I just updated the target branch to `release/3.x`. I didn't mean for this to be merged into `master` yet. Let's get `release/3.x` QA'd on .COM, and then we can think about merging with `master`.

remydenton:
Ah, OK, that's more what I had been expecting.  Merging this now so we can pull into pega.com for full regression testing.

PR #2100, opened 2021-02-08
DS 327 Fix page header background color and static position

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-327

## Summary

1. Fixes a bug where the Page Header does not work with color themes.
2. Fixes a bug where the static prop of the Page Header is not working expected (other sticky elements overlaps the Page Header). 

## Details

1. Converted the CSS to use mode specific color vars.
2. Fixed the static version to use `position: relative` instead to maintain the correct stacking context.

## How to test

Run the branch locally and do the following us:

1. Add a theme class (eg. `.t-bolt-dark`) to the page header. Make sure it works with all themes.
2. Create a page that has a static page header and a sticky subnav. Make sure the sticky subnav does not overlap the page header's dropdown links.

### DIFF
--- a/packages/components/bolt-page-header/src/_page-header-desktop.scss
+++ b/packages/components/bolt-page-header/src/_page-header-desktop.scss
@@ -13,10 +13,10 @@
     padding-bottom: $bolt-page-header-desktop-padding-bottom;
     padding-left: var(--bolt-page-padding-x);
     background: linear-gradient(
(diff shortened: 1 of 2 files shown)


colbytcook (approved):

danielamorse (approved):
Works as expected. Theme switching is pretty cool 😎

PR #2101, opened 2021-02-09
DS-328 fix starter kit CSS order and misc visual bugs

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-328

## Summary

1. Fixes a bug where the form select input's text overlaps the chevron down icon
2. Fixes a bug where the underline on hover is not displaying for the "reversed underline" link variant
3. Fixes a bug where the page header's CTA text is off centered in mobile view
4. Fixes a bug where the Elements CSS is loaded after Components CSS

## Details

1. Added proper padding to the select input
2. Added missing text-decoration hover styles to link component and text link element
3. Re-ordered the elements and components within `boltrc.js` in starter kit (which also fixes the page header's CTA text alignment)
4. Removed layouts-list from `package.json` and `boltrc.js`, because it was causing the List component to break.

## How to test

Run the branch locally and check form, link, and page header docs. Make sure they are displaying properly now.

## Release Notes

**Button Group**
Removed `contentItems` prop. Use `buttons` instead.


mikemai2awesome:
@colbytcook I extracted some of the CSS fixes from https://github.com/boltdesignsystem/bolt/pull/2085, so we can get them merged while we still work on Best of Content.

danielamorse (approved):
Looks good. The order things are now in makes sense.

mikemai2awesome:
Waiting for https://github.com/boltdesignsystem/bolt/pull/2102 to take care of the failing deprecated card tests.

packages/components/bolt-form/src/form.scss
```diff
@@ -244,6 +244,12 @@ $bolt-input-transition: $bolt-transition;
   display: inline;
 }
 
+.c-bolt-input--select {
+  padding-right: calc(
+    var(--bolt-spacing-x-small) + 24px
```
colbytcook, line 249: Would it make sense (and be worth the effort) to change the "24px" here to a variable since it is also being used in the icon.scss? This is certainly not a blocker for this PR but might be a nice structural update moving forward.

We could create a system to update this according to the "small", "medium", "large" and "xlarge" icon sizes?

mikemai2awesome: Icon needs a refactor to be responsive, I'd rather do it then. The number won't change drastically even if it is converted to `rem` or `em`, so this isn't gonna break that easily.

colbytcook (approved):

danielamorse:
@remydenton FYI, we're removing the `contentItems` prop from Button Group in this PR. It throws an error due to the use of the ["pattern_template" function](https://github.com/boltdesignsystem/bolt/pull/2101/files#diff-eb8ffd383c5df0a96938d3ee0cf40fce067768034471bc22097c5e7e72e8ce47L35) and there being two Button components, element- and component-. Twig error is thrown when it can't resolve "button" via the "pattern_template" function.

I [searched GitLab](https://gitlab.com/search?group_id=5804310&project_id=&repository_ref=&scope=blobs&search=button-group%2A&snippets=false) and found no usage of `contentItems`. Every instance uses `buttons` prop.

remydenton:
Amen, every time someone uses `contentItems` or `pattern_template` a fairy dies.

PR #2105, opened 2021-02-10
DS-296 Image Element

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-296

## Summary

Adds an Image element that handles responsive images.

## Details

1. Image element is created with minimal CSS.
2. `fill` prop allows an image to always stretching to 100% width.
3. `background` prop transform the image into an absolute positioned background image. This removes the need to use the Background component.
4. Two CSS custom props are available to further adjust the background image fit and position.

## How to test

Run the branch locally and read through all the Image docs and demos under Elements. Make sure all props work as expected.


colbytcook:
This work is in a draft state until we start the Bolt Image sprint.

docs-site/src/pages/pattern-lab/_patterns/20-elements/image/05-image.twig
```diff
@@ -0,0 +1,51 @@
+{% set description %}
+  An image element embeds a responsive graphic on the page.
+{% endset %}
+
+{% set notes %}
+  <bolt-ol>
+    <bolt-li>To make an image accessible to screen readers and other assistive technology, make sure to include text description via the <bolt-code-snippet display="inline">alt</bolt-code-snippet> attribute.</bolt-li>
+    <bolt-li>To indicate an image as decoration, leave the <bolt-code-snippet display="inline">alt</bolt-code-snippet> attribute blank, do not remove it. For example: <bolt-code-snippet display="inline">alt=""</bolt-code-snippet>.</bolt-li>
+    <bolt-li>It is best practice to always define the <bolt-code-snippet display="inline">width</bolt-code-snippet> and <bolt-code-snippet display="inline">height</bolt-code-snippet> attributes, so the space that the image would take up is already calculated before the image finishes loading. This helps to reduce cumulative layout shifts.</bolt-li>
+    <bolt-li>If responsiveness is not needed, please use a plain <bolt-code-snippet display="inline">&lt;img&gt;</bolt-code-snippet> element instead.</bolt-li>
```
colbytcook, line 10: @mikemai2awesome is there a use case for a non-responsive image?

colbytcook (approved):

colbytcook:
Approved, just have a quick question

packages/elements/bolt-image/src/image.twig
```diff
@@ -0,0 +1,27 @@
...
+  {{ validate_data_schema(schema, _self)|raw }}
+{% endif %}
+
+{# Variables #}
+{% set this = init(schema) %}
+{% set attributes = create_attribute(attributes|default({})) %}
+
+{% set classes = [
+  'e-bolt-image',
+  this.data.fill.value ? 'e-bolt-image--fill',
```
danielamorse, line 22: @mikemai2awesome changed to `this.data.foo.value`. Must use `.value` when doing a conditional check with the data that comes back from `init()`. Otherwise, the presence of any value will be truthy.

packages/elements/bolt-image/src/image.twig
```diff
@@ -0,0 +1,27 @@
...
+  this.data.background.value ? 'e-bolt-image--bg',
+] %}
+
+{# Template #}
+{% spaceless %}<img {{ attributes.addClass(classes) }} alt="{{ alt }}" src="{{ src }}" {% if srcset %}srcset="{{ srcset }}"{% endif %} {% if sizes %}sizes="{{ sizes }}"{% endif %}>{% endspaceless %}
```
danielamorse, line 27: @mikemai2awesome all of the attributes listed individually here are passed through `attributes`. Listing them out individually will output them twice (view source rather than inspecting with dev tools and you'll see). All that's needed is `attributes.addClass(classes)`.

danielamorse (changes requested):
@mikemai2awesome looking great. I love the simplicity. The docs are really well written, even if the topic of `srcset` and `sizes` is hard to grasp.

I made a couple minor fixes and made one inline change request.

docs-site/src/pages/pattern-lab/_patterns/20-elements/image/05-image.twig
mikemai2awesome: yes, something like a avatar photo that stays at 48px by 48px.

packages/elements/bolt-image/src/image.twig
remydenton: +1, this should probably just look like this:

```suggestion
{% spaceless %}<img {{ attributes.addClass(classes) }}>{% endspaceless %}
```
I'm also confused about the use of `{% spaceless %}` here.  Per the docs:
> Use the spaceless tag to remove whitespace between HTML tags, not whitespace within HTML tags or whitespace in plain text

There's only one self-closing HTML tag here, so I'm not sure how we'd ever end up with whitespace between tags 🤔 .  Maybe this all boils down to the following?

```suggestion
<img {{ attributes.addClass(classes) }}>
```

remydenton (changes requested):
See inline comment about attributes.  Otherwise, this looks great, such an improvement over the image component 😂

packages/elements/bolt-image/src/image.twig
mikemai2awesome: thanks for the change

packages/elements/bolt-image/src/image.twig
mikemai2awesome: Updated, removed all that.

danielamorse (approved):

remydenton (approved):

PR #2106, opened 2021-02-11
DS-325: Deprecate Teaser in Pattern Lab

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-325](https://pegadigitalit.atlassian.net/browse/DS-325)

## Summary

remove and deprecate teaser in pattern lab.

## Details

bolt-teaser doesn't do anything we can't do with other components, and we'd like the name to make our best-of cards. this deprecates teaser in pattern lab and provides a point of reference for the drupal devs to replace any instances of teaser.

## How to test

check out pattern lab. i saw examples in card-replacement, tabs, and obviously, teaser.

## Release notes

deprecating bolt-teaser, to be replaced


danielamorse (commented):
@cjwhitedev I removed the vrt-urls test since it's just getting in the way.

I see we're replacing the name "Teaser" with "Text group". @mikemai2awesome are you good with "Text group" as a name for a text pattern? Should it just be "Text" as it is the Text component?

As soon as the new Teaser component is ready to go the "Teaser (Deprecated)" section should be deleted because you won't be able to use the old Teaser, but just in case we need to do another release before the best of content work is ready to go, I agree this is good to have. 🙏

danielamorse (approved):

mikemai2awesome (changes requested):
@danielamorse @cjwhitedev Can we nix 05-teaser-repalcement-example.twig? I don't want to show anything other than the deprecated warning. This is so trivial to put together, I don't see a need for it.

mikemai2awesome (approved):

PR #2108, opened 2021-02-16
DS-323: deprecate auto slides per view. set max slides to 4.

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-323](https://pegadigitalit.atlassian.net/browse/DS-323)

## Summary

Allow users to have 4 slides per view.

## Details

Change schema to allow 4 slides per view and a max sliders per view of 4. Deprecate auto from slides per view to avoid further confusion.

## How to test

Check out carousel demos

## Release notes

`auto` has been deprecated from the carousel `slides_per_group` prop.  Use an integer (1, 2, 3, 4) for `slides_per_group` instead.  Note that the number of items shown will be always automatically reduced as necessary at smaller breakpoints.

### DIFF
--- a/packages/components/bolt-carousel/carousel.schema.js
+++ b/packages/components/bolt-carousel/carousel.schema.js
@@ -19,9 +19,10 @@ module.exports = {
     },
     slides_per_view: {
       type: 'string',
-      description: 'Controls the number of slides to be shown at once.',
+      description:
+        'Controls the number of slides to be shown at once. Note: the value auto is deprecated, please be explicit.',
       default: '1',
-      enum: ['auto', '1', '2', '3'],
+      enum: ['1', '2', '3', '4'],
     },
     slides_per_group: {
       type: 'string',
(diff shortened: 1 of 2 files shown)


packages/components/bolt-carousel/carousel.schema.js
```diff
@@ -33,9 +34,9 @@ module.exports = {
       type: 'integer',
       description:
         'Limits the maximum number of slides that can display at any screen size. Combine this with the <code>slidesPerView</code> option to get the benefits of automatically adjusting the number of slides displayed at different screen sizes while putting a max limit of how many slides display at once.',
-      default: 3,
+      default: 4,
```
mikemai2awesome, line 37: We are changing a default here, is this a breaking change?

mikemai2awesome (changes requested):

packages/components/bolt-carousel/carousel.schema.js
cjwhitedev: It is not. The only case where the default mattered as far as I can tell is when folks used `auto` sliders per view. We're removing that option so changing the max will not have an effect.

remydenton: We had a giant convo in Slack about this on Friday.  It will only affect carousels that have `slides_per_view` set to `auto`, which is not the default.  The rationale for this change is a) not many carousels will have `auto` selected since it's functionally the same as `3` and b) if `auto` _is_ selected, the intention is to have the number determined automatically, which it still is, it will now just be 4 if 4 items will fit.

remydenton (approved):
Looks great.  I love how we don't even have to update the PL demo since it pulls automatically from schema!

danielamorse (approved):
Nicely work, works perfectly 👍

mikemai2awesome (approved):

PR #2109, opened 2021-02-16
DS-303 Fix init() bugs

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-303

## Summary

Fix an `init()` bug where `this.props` and `this.data` values do not match when a prop like `target` is passed via `attributes` object. Also, omit deprecated props from the object returned by `init()`.

## Details

**Include matching attributes in `this.data`**
On the Button component, when you pass `attributes: { target: "_blank" }` (something I assume we support because there is a Jest test for it), that `target` value is added to `<bolt-button>`, but the default value, `_self`, is added to the inner `<a>`. This is because attributes that match a component prop are added to `this.props` but not `this.data`, and we use `this.data` to determine what value the `<a>` gets. This PR updates `init()` to include matching attributes in `this.data`.

**Omit deprecated props**
Make sure deprecated props are not returned in `init()` object. This ensures they are not added to the custom-element. All deprecated props should be converted before `init()` is called. This came up on Button where both `hierarchy` and `style` were being output on the `<bolt-button>` element. Rather than filter out each deprecated prop in the component, let's filter them out in `init()`.

I'm making a separate PR for these fixes since they affect all components. I wanted it to be reviewed and tested separately.

## How to test

- Tests are passing
- No regressions to existing components. Pay especially close attention to `BoltActionElements` like Button, Link, Trigger.

### DIFF
--- a/packages/twig-integration/twig-extensions-shared/src/Utils.php
+++ b/packages/twig-integration/twig-extensions-shared/src/Utils.php
@@ -251,9 +251,17 @@ class Utils {
             $convertedPropName = $isData ? self::convertToSnakeCase($propName, $caseType) : self::convertToKebabCase($propName, $caseType);
             $props[$convertedPropName] = $items[$propName];
           }
+          elseif (isset($items["attributes"][$propName])) {
+
+            // No value for this prop was found in $items, but one was defined via attributes.
+            if ($isData) {
+              $convertedPropName = self::convertToSnakeCase($propName, $caseType);
+              $props[$convertedPropName] = $items["attributes"][$propName];
+            }
+          }
           elseif (isset($propSchema["default"])) {
 
-            // No value for this prop was found in $items, but a default is defined in the schema.
+            // No value for this prop was found in $items or attributes, but a default is defined in the schema.
             if ($isData) {
               $convertedPropName = self::convertToSnakeCase($propName, $caseType);
               $props[$convertedPropName] = $propSchema["default"];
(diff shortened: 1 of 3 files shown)


cjwhitedev (approved):

packages/twig-integration/twig-extensions-shared/src/Utils.php
```diff
@@ -251,9 +254,17 @@ public static function buildPropsArray($items, $schema, $isData = FALSE) {
             $convertedPropName = $isData ? self::convertToSnakeCase($propName, $caseType) : self::convertToKebabCase($propName, $caseType);
             $props[$convertedPropName] = $items[$propName];
           }
+          elseif (isset($items["attributes"][$propName])) {
```
remydenton, line 257: I was a little wary of this because of the differences between attributes in Bolt (where it's an array) and Drupal (where it's a special attributes object).  But [the docs](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Template%21Attribute.php/class/Attribute/8.2.x) for the object confirm that you can get properties using the same syntax as you would for an array, so I think this will work fine.

remydenton (approved):
Looks great, thanks @danielamorse.  Attributes still scare me a little because, unlike props, _anything_ can be passed in.  But maybe that concern is unfounded, and certainly these changes are needed if we want to support them.  Also, +1 for the deprecated fix.

packages/twig-integration/twig-extensions-shared/src/Utils.php
danielamorse: Thanks for following up on this one. Good to get a second look, especially with PHP.

PR #2110, opened 2021-02-23
Feature/ds 360 card action icon override

andrewmriv:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-360

## Summary

Allows users to override action icons in card components.

## Details

When adding an action to a card component, a user can specify which icon they would like shown. If no icon is specified, default logic is applied (`right-chevron` is shown for internal links and `external-link` is shown for external links.)

## How to test

- Review files changed
- Specify one of the icons from [here](http://localhost:3000/pattern-lab/?p=viewall-components-icon) in the actions an ensure that that icon is shown in place of the `chevron-right` or `external-link` icons
- Ensure that not specifying an icon falls back on the default behavior


packages/components/bolt-card-replacement/card-replacement.schema.js
```diff
@@ -121,6 +121,9 @@ module.exports = {
         external: {
           type: 'boolean',
         },
+        icon: {
+          type: 'string',
+        },
```
remydenton, line 126: This is correct the way you did it.  A few additional things I would add:
* Add a brief description so that users know they should be passing the name of a Bolt icon
* Optional, but a "none" option could be useful to match the way headline is doing it.  Do this last since you'll have to add some extra logic in the twig and js and it's not strictly required.

docs-site/src/pages/pattern-lab/_patterns/40-components/card/60-card-replacement-icon.twig
```diff
@@ -0,0 +1,21 @@
...
+  },
+  actions: [
+    {
+      text: "This is a button",
+      url: "https://pega.com",
+      icon: "download",
+    },
+  ],
+} only %}
+
```
remydenton, line 21: Instead of creating a new example, let's just add a third variant to the existing "card button variations" page.

https://feature-ds-360-card-action-icon-override.boltdesignsystem.com/pattern-lab/?p=components-card-button-variations

remydenton (changes requested):
Nice work on this with minimal guidance @andrewmriv.  I left a few inline comments, and there's one additional change which is adding support for the new prop to the web component javascript as well as twig (see the web component examples shown at https://boltdesignsystem.com/pattern-lab/?p=components-card-replacement-with-web-component).  I believe in this case, you would just need to update these two places in the js:

https://feature-ds-360-card-action-icon-override.boltdesignsystem.com/pattern-lab/?p=components-card-button-variations
https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js#L54

In retrospect, this task is a great introduction to some of the main oddities of Bolt 😄 .

packages/components/bolt-card-replacement/card-replacement.schema.js
remydenton: For reference, this schema is using the [JSON schema](https://json-schema.org/understanding-json-schema/) standard.

mikemai2awesome: What @remydenton said, `none` can completely remove the icon.

docs-site/src/pages/pattern-lab/_patterns/40-components/card/60-card-replacement-icon.twig
mikemai2awesome: Yes I agree.

mikemai2awesome (changes requested):
Good first pass. Just couple inline comments.

docs-site/src/pages/pattern-lab/_patterns/40-components/card/10-card-button-variations.twig
```diff
@@ -1,5 +1,5 @@
 {% grid "o-bolt-grid--flex o-bolt-grid--matrix o-bolt-grid--center" %}
-  {% cell "u-bolt-width-6/12 u-bolt-width-4/12@small" %}
+  {% cell "u-bolt-width-6/12" %}
```
andrewmriv, line 2: Since there are now 4 examples, I set the width to 50% so there could be 2 rows of 2 examples.

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig
```diff
@@ -8,12 +8,17 @@
           width: "full",
           align: "start",
           target: external == true ? "_blank" : "_self",
-          icon: {
-            name: external == false or external == null ? "chevron-right" : "external-link",
-          },
           attributes: attributes|default({})
         } %}
 
+        {% if icon is not empty and icon != 'none' %}
```
andrewmriv, line 14: Let me know if there is a better way to write out this if statement. I essentially wanted:
- If icon exists
-- If icon != 'none'
--- Set icon to specified icon
- else
-- if external == false or external == null
--- Set icon to chevron-right
-- else
--- Set icon to external-link

mikemai2awesome (approved):

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig
remydenton: It's pretty good as-is-- I think it was the right choice to do away with the ternary syntax.  How about the following though to clarify the "none" option and reduce the redundancy of the `icon is empty` checks?

Note that in twig, unlike PHP, you don't have to check if a variable is empty before doing other checks like if `icon != 'none'`.  If it's empty or undefined, that won't throw an error.
```
{% if icon != 'none' %}
  {% if icon %}
    {% set button_props = button_props|merge({ icon: { name: icon } }) %}
  {% elseif external == false or external == null %}
    {% set button_props = button_props|merge({ icon: { name: "chevron-right" } }) %}
  {% else %}
    {% set button_props = button_props|merge({ icon: { name: "external-link" } }) %}
  {% endif %}
{% endif %}
```

remydenton:
As discussed with @andrewmriv, we still need to update the logic in the web component.  My inline comment about icon logic is optional.

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig
andrewmriv: Thanks for this! Have never had to do a nested twig conditional so I didn't even realize that was possible. I updated it since this looks a lot better

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
```diff
@@ -49,10 +52,16 @@ class BoltCardReplacementAction extends withContext(BoltElement) {
                 size="${ifDefined(this.spacing ? this.spacing : undefined)}"
               >
                 ${this.slotify('default')}
```
andrewmriv, line 54: Apologies in advance if what I did below is not best practices. Happy to learn how to do this better if there is a better way.

Question - What does `this.slotify('default')` do?

danielamorse: @andrewmriv, nicely done! You did it right except for the `this.slotify()` part (which I wouldn't expect you to know about yet). I just replaced the second `this.slotify()` method with an empty string.

Do you have Prettier enabled? The `html` syntax is a pain to format by hand. I rely 100% on Prettier to fix it for me.

`this.slotify()` is a Bolt-ism. It outputs a `<slot>` element. The argument sets the `name` of the slot. So `this.slotify('icon')` outputs `<slot name="icon">`.

Slotify will output a "fake" slot if your browser doesn't support the `<slot>` element and/or shadow DOM, or if we intentionally want to render markup to the light DOM via the `no-shadow` prop.

It's a good topic for a future dev huddle!

danielamorse (approved):

remydenton (approved):

PR #2111, opened 2021-02-26
Upgrade @babel/core, @babel/preset-env

danielamorse:
## Jira

n/a

## Summary

Upgrade `@babel/core`, `@babel/preset-env` to fix "No 'exports' main defined" error.

## Details

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in node_modules/@babel/helper-compilation-targets/package.json
```
From [Github issue](https://github.com/babel/babel/issues/11216), the solution is to `update all @babel deps to v7.8.7 or above`.

I found it was only necessary to update `@babel/core` and `@babel/preset-env`. Both of which have a dependency on `@babel/helper-compilation-targets`, older versions of which seem to cause the bug.

I'm targeting Node v12 for this fix because that version has [long-term support](https://nodejs.org/en/about/releases/) and it is the version currently used by many of us and Pega properties. Later this year we might consider upgrading to Node v14.

## How to test

- Check out this feature branch.
- Get on Node 12:
  - `nvm install 12`
  - `nvm use 12`
- Run:
  - `yarn clean`
  - `yarn setup`
  - `yarn build`
- Verify build finishes without the "export" error shown above.

### DIFF
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -22,7 +22,7 @@
     "extends @bolt/browserslist-config"
   ],
   "dependencies": {
-    "@babel/core": "^7.4.5",
+    "@babel/core": "^7.8.7",
     "@bolt/api": "^3.1.0",
     "@bolt/babel-preset-bolt": "^3.1.0",
     "@bolt/browserslist-config": "^3.1.0",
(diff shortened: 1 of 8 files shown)


mikemai2awesome (approved):
Works as expected.

colbytcook (approved):

PR #2112, opened 2021-02-27
DS-388 Allow `download` attribute to be added to inner `<a>`

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-388

## Summary

Allow `download` attribute to be added to the inner `<a>` of `<bolt-button>` and download a resource without opening in a new tab.

## Details

Initially I thought the solution would be to allow arbitrary attributes to be added to to the inner element, but this is too risky as many attributes are already used on `<bolt-button>` in the wild, and suddenly applying them all to the inner element is a potential breaking change.  Instead, simply check for the `download` attribute and add it to the `<a>` like we do for `target`.

Note: only applies if Button has `url`.

## How to test
- Tests are passing
- Review files changed
- Run locally and paste this example into PL:
```
{% include "@bolt-components-button/button.twig" with {
  text: "This is a button",
  url: "http://pega.com",
  attributes: {
    download: "file.pdf"
  }
} only %}
```

### DIFF
--- a/packages/components/bolt-button/__tests__/__snapshots__/button.js.snap
+++ b/packages/components/bolt-button/__tests__/__snapshots__/button.js.snap
@@ -135,6 +135,24 @@ exports[`button Button type: submit 1`] = `
 </bolt-button>
 `;
 
+exports[`button Button will add \`download\` attribute to inner anchor element 1`] = `
+<bolt-button download="download.pdf"
+             url="download.pdf"
+>
+  <a href="download.pdf"
+     target="_self"
+     download="download.pdf"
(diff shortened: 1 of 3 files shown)


andrewmriv (approved):

mikemai2awesome (approved):
This is good, but fyi, the attribute should be used like this instead:

```twig
{% include "@bolt-components-button/button.twig" with {
  text: "Download",
  url: "file.pdf",
  attributes: {
    download: true
  }
} only %}
```

Only if you want to change the "save as" file name (and you don't put extension in the name), you'd put value into the attribute. Otherwise you just set it to true. This attribute also only works for files from the same origin.

@danielamorse @andrewmriv @remydenton

remydenton (approved):

PR #2115, opened 2021-03-02
DS-399 Remove example-integrations

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-399

## Summary

`example-integrations` is not actively maintained and it contains dependencies that require security updates. Remove it from the monorepo and consider moving into separate project on GitHub as an archive.

## Details

Removing `example-integrations` rather than continue to upgrade dependencies there to fix security vulnerabilities.

I suggest we archive this in a separate Bolt repo. Thoughts?

## How to test

- Review files changed.
- Comment, should we move this to a separate repo?


remydenton (approved):
I agree, rather than sinking time into maintaining stuff that isn't really being used, let's delete it.  The best place to see Drupal integration is... on our Drupal sites!

remydenton:
Re: archiving, I'm not even sure it's necessary.  All the code is stored in git if we ever want to restore it, but that'll get less likely as time goes on and it gets more and more outdated.

adamszalapski (approved):

danielamorse:
@adamszalapski thanks for your good work on these integrations. They will live on in our git history. 🙏

PR #2120, opened 2021-03-04
DS-399 Dependency cleanup

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-399

## Summary

Upgrade `yaml-loader`, remove `stylelint-selector-bem-pattern`.

## Details

This is the last of this round of dependency-cleanup. It upgrades `yaml-loader` and removes `stylelint-selector-bem-pattern` to fix security vulnerabilities. The latter stylint package does not appear to do anything (cc: @mikemai2awesome).

The remaining security vulnerabilities are so ingrained in Pattern Lab and Jest that they cannot be addressed unless we upgrade both of them. As far as I can tell, the vulnerabilities are not in the code we ship, but in the libraries we use in the monorepo. So, they could only be exploited by a contributor. @adamszalapski does that sound right?

## How to test
- Review code
- Tests pass

### DIFF
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -115,7 +115,7 @@
     "webpack-manifest-plugin": "^2.2.0",
     "webpack-merge": "^4.2.2",
     "write-file-webpack-plugin": "^4.5.1",
-    "yaml-loader": "^0.5.0",
+    "yaml-loader": "^0.6.0",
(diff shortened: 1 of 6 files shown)


adamszalapski (approved):
Looks awesome 😎

mikemai2awesome (approved):

PR #2121, opened 2021-03-04
DS-405: Upgrade babel to latest minor release to fix Edge 18 JS error

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-405

## Summary

Upgrade Babel to latest minor release to fix Edge 18 JS error.

## Details

After the `@babel` upgrade from `7.4.x` to `7.8.7` to fix a Node JS error (#2111), there is a new JS error in Edge (18/44) that breaks JS execution completely. I looked into the error but never pin pointed the exact cause, just that it appears after the upgrade. Upgrading all `@babel` dependencies from `7.8.7` to `7.13x` finally fixed the error and everything appears to work as expected. I attempted to find the exact release which "fixed" the issue but the whole cycle of switching versions, cleaning, installing and rebuilding is extremely time-consuming. In the end, since the latest version works, I just went with that.

## How to test

- Tests are passing
- Runs locally
- No errors in Edge 18/44 or any other supported browsers

### DIFF
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -22,7 +22,7 @@
     "extends @bolt/browserslist-config"
   ],
   "dependencies": {
-    "@babel/core": "^7.8.7",
+    "@babel/core": "^7.13.8",
     "@bolt/api": "^3.1.0",
     "@bolt/babel-preset-bolt": "^3.1.0",
     "@bolt/browserslist-config": "^3.1.0",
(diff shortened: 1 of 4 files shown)


colbytcook (approved):
I was able to confirm that I did not see an errors on Edge v.18

adamszalapski (approved):

PR #2127, opened 2021-03-11
Fix: quick filters overflow style

mikemai2awesome:
## Summary

Fixes an issue where the Best of Content quick filters' overflow fading effect is displaying incorrectly.

## Details

1. Added back JS to determine the different states of overflow.
2. Changed CSS to do a shadow effect instead of fading. This will work regardless of background color or image.

## How to test

Run the branch locally and check out the Best of Content gallery page in phase 1 and phase 2 where the quick filters are introduced. Adjust your browser width wide enough so the quick filters are not overflowing, make sure there are no undesired gradient or shadows on the edges of quick filters. Then adjust your browser width narrow enough so the quick filters are overflowing, horizontally scroll them to make sure there's a hint of shadow to indicate scrolling.


mikemai2awesome:
Waiting for this to merge into master first: https://github.com/boltdesignsystem/bolt/pull/2123

docs-site/src/pages/pattern-lab/_patterns/50-pages/65-best-of-content/00-boc-quick-filters-docs.twig
```diff
@@ -0,0 +1,270 @@
...
+      // } only %}
+    </div>
+  </div>
+</form>
+{% endverbatim %}
+{% endset %}
+
+{% set css %}
+{% verbatim %}
+@import '@bolt/core-v3.x';
```
danielamorse, line 60: Not a blocker, but I'd rather point devs to the live code. This means we now need to update code in two places should anything change.

danielamorse (approved):

PR #2132, opened 2021-03-17
DS-333: replace onClick and onClickTarget with snake case

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-333](https://pegadigitalit.atlassian.net/browse/DS-333)

## Summary

deprecate onClick and onClickTarget in BoltActionElement in favor of snake case

## Details

Replace BoltActionElement properties onClick and onClickTarget with snake case counterparts, on_click and on_click_target

## How to test

Make sure none of the demos are broken, and confirm that any precautions put in place would prevent disaster. Also confirm I didn't miss any uses of the camel case props.

## Release notes

deprecate onClick and onClickTarget for snake case.


packages/components/bolt-link/src/link.twig
```diff
@@ -28,12 +38,12 @@
...
 
-{% if onClick or attributes["on-click"] %}
-  {% set onClick = onClick | default(attributes["on-click"]) %}
+{% if on_click or attributes["on-click"] %}
+  {% set on_click = on_click | default(attributes["on-click"]) %}
 {% endif %}
 
-{% if onClickTarget or attributes["on-click-target"] %}
-  {% set onClickTarget = onClickTarget | default(attributes["on-click-target"]) %}
+{% if on_click_target or attributes["on-click-target"] %}
```
danielamorse, line 45: We haven't converted Link to use `init()` yet? Totally missed this. Is there a ticket to convert this component?

packages/base-element/src/BoltActionElement.js
```diff
@@ -15,11 +15,11 @@ class BoltActionElement extends BoltElement {
       disabled: {
         type: Boolean,
       },
-      onClick: {
+      on_click: {
         type: String,
         attribute: 'on-click',
       },
-      onClickTarget: {
+      on_click_target: {
```
danielamorse, line 22: Here in the JS these will always be camelCase, only snake case in the schema and Twig.

If you import the schema and convert this base Class to use the `...this.props` pattern they will be automatically converted for you. I suggest doing that, just like you did on a bunch of other components.

packages/base-element/src/BoltActionElement.schema.js
```diff
@@ -24,15 +24,23 @@ module.exports = {
...
+        'When used with on_click_target, an event to fire on the targeted elements when this element is clicked.  When used without on_click_target, arbitrary javascript to execute when this element is clicked.',
     },
-    onClickTarget: {
+    on_click_target: {
       type: 'string',
       description:
-        'Requires onClick.  A CSS selector for elements that the onClick event will fire on when this element is clicked.',
+        'Requires on_click.  A CSS selector for elements that the on_click event will fire on when this element is clicked.',
+    },
+    onClick: {
```
danielamorse, line 37: Nice, deprecate it here and import this everywhere 👌

danielamorse (changes requested):
@cjwhitedev, looks good! See my comment about converting `packages/base-element/src/BoltActionElement.js`. Then PR this and send for review 👍

packages/components/bolt-link/src/link.twig
cjwhitedev: [https://pegadigitalit.atlassian.net/browse/DS-372](https://pegadigitalit.atlassian.net/browse/DS-372)

packages/base-element/src/BoltActionElement.js
cjwhitedev: should be resolved with my latest commit, but im not sure how to test it properly. the console log seemed correct at least.

danielamorse (approved):
Build is now passing 😅 

I increased the Jest timeout globally, but what I really think fixed it was that I updated the Button, Link, and Trigger snapshots which removed an old test result from the file. I'm not sure how it got in that state, but updating it seemed to fix everything.

PR #2135, opened 2021-03-19
DS-421: passing custom attributes to the band

ckjackslack:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-421](https://pegadigitalit.atlassian.net/browse/DS-421)

## Summary

Creating custom attribute list and passing them to the band include.

## Details

As in summary. The explanation why this change is needed is in the ticket in Job Summary.

## How to test

As far as Drupal projects are concerned
- add Hero component to the content page
- try to pass something as an attribute e.g. from within the hook `pega_bolt_hero_paragraph_view`, it should be added to the band markup

### Visual changes

There should be no visual changes as long as nothing is passed.

### DIFF
--- a/packages/components/bolt-hero/hero.twig
+++ b/packages/components/bolt-hero/hero.twig
@@ -6,6 +6,11 @@
 
 {% set this = init(schema) %}
 
+{% set band_attributes = create_attribute({}) %}
+{% if attributes %}
+  {% set band_attributes = band_attributes|merge(attributes) %}
+{% endif %}
+
 {% set imageValign = this.data["image_valign"].value %}
 {% set imageAlign = this.data["image_align"].value %}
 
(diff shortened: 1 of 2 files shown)


packages/components/bolt-hero/hero.twig
```diff
@@ -6,6 +6,11 @@
 
 {% set this = init(schema) %}
 
+{% set band_attributes = create_attribute({}) %}
```
mikemai2awesome, line 9: I think you can just do this one line instead:

```
{% set attributes = create_attribute(attributes|default({})) %}
```

mikemai2awesome (changes requested):
@ckjackslack Awesome job getting your first PR in. We also need to add `attributes` to the schema. Something like this will work:

```js
attributes: {
  type: 'object',
  description:
    'A Drupal attributes object. Applies extra HTML attributes to the parent element.',
},
```

remydenton (approved):
Looks great, thanks for the PR @ckjackslack.

mikemai2awesome (approved):

PR #2143, opened 2021-03-24
DS-425 Fix Tabs "ready" event

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-425

## Summary

Waiting until Tabs component is fully rendered to dispatch the "ready" event.

## Details

Tabs is not really ready on initial render. It takes one more cycle of rendering before the Tabs markup is all there. The base class, however, dispatches the "ready" event after first render in [the `firstUpdated()` function](https://github.com/boltdesignsystem/bolt/blob/master/packages/base-element/src/lib/decorators/render-and-rendered-events.js#L15-L23). This PR updates Tabs to wait until after the second render to call the base class `firstUpdated()` function.

## How to test
- Review the code.
- On `master` add this code to `docs-site/src/index.js`:
```
const tabs = document.querySelector('bolt-tabs');
tabs.addEventListener('ready', function(e) {
  if (tabs === e.target) {
    console.log(
      tabs.renderRoot.querySelectorAll('.c-bolt-tabs__label-text'),
    );
  }
});
```
- Then go to a page with Tabs on it and open the console. It should return an empty array.
- Then check out this feature branch and repeat. It should return the tab label elements.

### DIFF
--- a/packages/components/bolt-tabs/src/tabs.js
+++ b/packages/components/bolt-tabs/src/tabs.js
@@ -479,9 +479,7 @@ class BoltTabs extends withContext(BoltElement) {
     });
   }
 
-  firstUpdated() {
-    super.firstUpdated && super.firstUpdated();
-
+  async firstUpdated() {
     // Use `slotMap` not `querySelectorAll` so that we don't get nested panels
     this.panels = this.slotMap
       .get('default')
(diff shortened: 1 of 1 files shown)


remydenton (approved):
Looks great and works based on the testing instructions.  Thanks for the fix @danielamorse!

PR #2144, opened 2021-03-26
DS-239 Popover rework

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-239

## Summary

Convert Popover and Tooltip to use Tippy.js with `focus-trap`.

## Details
- Replace Popper with Tippy.js
- Configure Tippy to append content to end of page to solve z-index issues
- Add focus management plugin that works with `focus-trap`
- Add `trigger_event` prop to Popover to show on "hover"
- Set Popover `theme` default to `xlight`, as popover content is no longer automatically inherited from trigger
- Remove no-JS tooltip/popper solution, content is on the page but hidden without JS
- Move `focus-trap` to core so that it can be shared across components
- Delete some old comments, unused dependencies in Modal

Known issues/open questions for @mikemai2awesome :
- Should Tooltip use the "bubble-tip" style or always round?
- If you pass an unfocusable element to Popover as trigger, you can't use enter key to show popover, only click. Is this a use case we support?
- When `<focus-trap>` gets focus, e.g. when you click into it, there is an outline.
- Where is the best place for the shared Tippy CSS?
- When a popover opens on "hover" and you tab into the popover, should it stay open until you mouseout or hit escape? Currently closes.

## How to test
- Checkout feature branch, run `yarn setup`
- Review code
- Review Tooltip and Popover demos, including new Popover [Trigger Event demo](https://feature-ds-239-popover-rework.boltdesignsystem.com/pattern-lab/?p=components-popover-trigger-event-variations)
- Review [Best of Content Gallery (Phase 1) demos](https://feature-ds-239-popover-rework.boltdesignsystem.com/pattern-lab/patterns/50-pages-65-best-of-content--20-boc-gallery-phase1/50-pages-65-best-of-content--20-boc-gallery-phase1.html), "share" button
- Review tab management solution, i.e. what happens when you use the keyboard to navigate popover content


packages/components/bolt-popover/src/popover.js
```diff
@@ -22,186 +15,33 @@ class BoltPopover extends BoltElement {
...
-        this.popper.destroy();
-      }
-      this.setupPlacement();
-    }
-  }
-
-  setupPlacement() {
-    this.popover = this.renderRoot.querySelector('.c-bolt-popover');
-    this.content = this.renderRoot.querySelector('.c-bolt-popover__content');
-
```
colbytcook, line 185: @danielamorse everything above that was removed I am assuming was old rendering stuff or popperjs stuff?

packages/components/bolt-popover/src/popover.js
```diff
@@ -212,69 +52,84 @@ class BoltPopover extends BoltElement {
...
 
-  render() {
-    if (!this._wasInitiallyRendered) {
-      this.sortChildren();
-      this.getTextContentLength();
+    this.trigger = firstFocusableElement || this;
+    this.content = this.querySelector('[slot="content"]');
+    // this.trigger = this.slotMap
+    //   .get('default')
+    //   .find(item => item instanceof HTMLElement);
```
colbytcook, line 72: @danielamorse do we need this?

packages/components/bolt-popover/src/popover.js
danielamorse: Yes, old state and class management that is now handled by Tippy 😄

packages/components/bolt-popover/src/popover.js
danielamorse: No, nice catch. Removed it, will push in a few mins once the current build finishes.

packages/core-v3.x/utils/tippy/index.js
```diff
@@ -0,0 +1,142 @@
+import tippy from 'tippy.js';
+import { queryShadowRoot } from '../query-shadow-root';
+import { isFocusable, isHidden } from '../is-focusable';
+
+export { tippy };
+
+// Note: if `hideOnEsc` or `handleFocus` fail, they fail silently
+export const hideOnEsc = {
```
cjwhitedev, line 8: Is using the escape key to close the popover better than just tabbing out of it?

cjwhitedev (commented):
If the escape key is the accessibly correct way to handle this, consider this approved. Everything looked great to me.

packages/core-v3.x/utils/tippy/index.js
mikemai2awesome: yes, esc capability is needed for accessibility.

mikemai2awesome (changes requested):
@danielamorse I found the following issues:

1. Auto and bottom placement options don't have transitions.
2. Can't close popover by clicking on trigger.
3. Popover placement options don't have the correct tip.

danielamorse:
@colbytcook addressed [your comments](https://pegadigitalit.atlassian.net/browse/DS-239?focusedCommentId=25569), please take another look. You can comment right here if there's any additional feedback.

@mikemai2awesome nice catch. `#2` and `#3` I broke on Friday while making some other "fixes". Note: I ended up moving the `tippy-box` rules to [the vendor CSS file](https://github.com/boltdesignsystem/bolt/pull/2144/commits/c1c9d66228d8abcadde9d596d1baf1c7306331c8) now that we have designated "theme" names for tooltip and popover. I also had to move border-radius and margin rules to fix specificity issues breaking the tips and positioning. I couldn't reproduce `#1`. Can you retest all of the above [here](https://github.com/boltdesignsystem/bolt/pull/2144/commits/c1c9d66228d8abcadde9d596d1baf1c7306331c8)?

colbytcook (approved):

mikemai2awesome:
@danielamorse As discussed in huddle, we need to take care of focusable elements within the popover, keyboard interactions when using the hover option, and the missing transitions/animations for ones that are coming down.

danielamorse:
@mikemai2awesome fixed the following:
- Animations in Safari (https://github.com/boltdesignsystem/bolt/pull/2144/commits/5389a42c24aae7da3ccdf11256c76b09461a93b4)
- Added CSS comments referencing vendor styles (https://github.com/boltdesignsystem/bolt/pull/2144/commits/0ed2471a2d6085c73768f59ff2174a7426e99c2b)
- Updated "hover" option to open on enter/space not focus (https://github.com/boltdesignsystem/bolt/pull/2144/commits/82b60cb9eaca435542d65c188b28dda6295de04f)
- Fixed focus management in Safari (https://github.com/boltdesignsystem/bolt/pull/2144/commits/9afc215639ce321653ba45a16e10c5071dd1b2b8)

You can pull the latest and retest (not deployed yet).

mikemai2awesome:
@danielamorse I updated the the typos in the doc page and rewrote the copy a little bit to be less techy. Also all other docs are already showing the default click behavior, it's a bit redundant to show again. The page can just focus on talking about hover.

mikemai2awesome (approved):

PR #2145, opened 2021-03-29
DS-372: Standardize Twig and Schema (Link)

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-372](https://pegadigitalit.atlassian.net/browse/DS-372)

## Summary

Standardize Twig and Schema for Link

## Details

Standardize Twig and Schema for Link. `shadow-root` is removed intentionally, as it is no longer needed. The snapshots have been updated because the default values of properties no longer need to be printed. For more detailed information, please see the initial ticket, as well as the associate TWIG and Schema tickets. 

## How to test

Check out the tests and demos and make sure nothing is a breaking change in the code.

## Release notes

Deprecate `isHeadline` prop in favor or `is_headline`.


colbytcook (approved):

danielamorse (changes requested):
Nicely done! Almost there. I left some inline comments and requested a few changes.

I think it'd be good to note a few details that might not be obvious to the reviewer in the 
**Details** section. For instance, I would mention that `shadow-root` is intentionally removed as it's no longer necessary, and that certain attributes no longer show up in the snapshots because they are default values that aren't output once we switch to use `init()`. I know these things because we've been looking at it together, but another reviewer might not. You don't have to go overboard with every detail, just the things you think a newcomer might not get right away.

packages/components/bolt-link/src/link.twig
```diff
@@ -1,11 +1,21 @@
...
 
-{% set schema = bolt.data.components["@bolt-components-link"].schema %}
+{% set schema = bolt.data.components['@bolt-components-link'].schema %}
 
 {% if enable_json_schema_validation %}
   {{ validate_data_schema(schema, _self) | raw }}
 {% endif %}
 
+{# DEPRECATED.  Use the property `url` instead of `href`. #}
+{% if href %}
```
danielamorse, line 10: @cjwhitedev I added this simple deprecation statement instead of [the old one](https://github.com/boltdesignsystem/bolt/pull/2145/files#diff-a570967cefe213c5d718a8efcdb554f61d28351886ef81a79a1b6b480a99de1fL31), which checked for both the prop and the attributes. `init()` will consolidate the props and the attributes _and_ add them to the `this.data` object, so it's safe to just check for the prop here.

packages/components/bolt-link/src/link.twig
```diff
@@ -16,101 +26,70 @@
...
+1. Those that should be applied to the inner tag (namely, 'is-' and 'has-' classes)
 2. Those that should be applied to the outer custom element (everything else EXCEPT c-bolt-* classes, which should never be passed in via atttributes)
 #}
 {% set outer_classes = [] %}
 {% set inner_classes = classes %}
 
-{% for class in attributes["class"] %}
-  {% if class starts with "is-" or class starts with "has-" %}
+{% for class in this.props.class %}
+  {% if class starts with 'is-' == true %}
```
danielamorse, line 73: @cjwhitedev is the `== true` necessary? Looks like we're doing it both ways throughout the codebase. It would be nice to figure out the best way and normalize it.

packages/components/bolt-link/src/link.twig
```diff
@@ -16,101 +26,70 @@
...
 {# link component's custom element wrapper #}
 {% spaceless %}<bolt-link
-  {% if display %} display="{{ display }}" {% endif %}
-  {% if valign %} valign="{{ valign }}" {% endif %}
-  {% if url %} url="{{ url }}" {% endif %}
-  {% if target %} target="{{ target }}" {% endif %}
-  {% if isHeadline %} is-headline="true" {% endif %}
+  {{ this.props|without('href')|without('class') }}
   {% if outer_classes %} class="{{ outer_classes|join(' ') }}" {% endif %}
-  {% if on_click %} on-click="{{ on_click }}" {% endif %}
```
danielamorse, line 93: @cjwhitedev we can remove these now that you've converted them to snake case and we're using `this.props` 🎉

danielamorse (approved):
Thanks for addressing the feedback @cjwhitedev . Approved, just waiting on https://github.com/boltdesignsystem/bolt/pull/2156 to be merged to fix failing Carousel tests.

packages/components/bolt-link/src/link.twig
danielamorse: I checked and the `==` is not necessary. I updated this block to look like the example in the Wiki Twig example: https://github.com/boltdesignsystem/bolt/wiki/Twig%3A-Example-Component

colbytcook (approved):

PR #2147, opened 2021-03-29
DS-298 Tech doc layout design

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-298

## Summary

Adds Side Nav component, Content Pagination component, Holy Grail layout, and Article element, and updates TOC component with a transparent variant.

## Details

1. New components: Side Nav (docs commented out) and Content Pagination
2. New layout: Holy Grail (Layout section in PL is now open, but docs commented out)
3. New element: Article
4. Update [Jest config](https://github.com/boltdesignsystem/bolt/pull/2147/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.scss&file-filters%5B%5D=.snap&file-filters%5B%5D=.twig&hide-deleted-files=true#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1) to fix broken tests

## How to test

Run the branch locally and check the docs for TOC component, ~~Side Nav component~~ (commented out, WIP), Content Pagination component, ~~Holy Grail layout~~ (commented out, WIP), and Article element. Make sure all props work as stated in the docs.

### Visual changes

TOC component's active indicator's (the teal bar) width is reduced to better reflect information architecture. Side Nav carries more importance than TOC, so TOC's style need to be reduced a little bit.


packages/components/bolt-content-pagination/content-pagination.schema.js
```diff
@@ -0,0 +1,69 @@
...
+    },
+    parent: {
+      type: 'object',
+      description: 'Set the parent page link.',
+      properties: {
+        content: {
+          type: 'string',
+          description: 'Text content of the link.',
+        },
+        tooltip_content: {
```
colbytcook, line 19: Should we set defaults to these "tooltip_content", something like "[X] topic" depending on the direction of the functionality?

packages/components/bolt-content-pagination/content-pagination.schema.js
```diff
@@ -0,0 +1,69 @@
...
+    },
+    previous: {
+      type: 'object',
+      description: 'Set the previous page link.',
+      properties: {
+        content: {
+          type: 'string',
+          description: 'Text content of the link.',
+        },
+        tooltip_content: {
```
colbytcook, line 38: same as above

packages/components/bolt-content-pagination/content-pagination.schema.js
```diff
@@ -0,0 +1,69 @@
...
+    },
+    next: {
+      type: 'object',
+      description: 'Set the next page link.',
+      properties: {
+        content: {
+          type: 'string',
+          description: 'Text content of the link.',
+        },
+        tooltip_content: {
```
colbytcook, line 57: same as above

packages/components/bolt-side-nav/src/side-nav.js
```diff
@@ -0,0 +1,13 @@
+var sideNavLinkToggleTrigger = document.querySelectorAll(
+  '.js-bolt-side-nav-toggle-trigger',
+);
+
+sideNavLinkToggleTrigger.forEach(function(el) {
+  el.addEventListener('click', function(event) {
+    if (el.getAttribute('aria-expanded') === 'true') {
```
colbytcook, line 7: Are we certain this booleans will always be a strings? Would "==" be safer here?
@danielamorse this might be a question for you.

packages/global/styles/04-elements/_elements-misc.scss
```diff
@@ -1,5 +1,84 @@
+/* ------------------------------------ *\
+   ELEMENTS - Misc.
```
colbytcook, line 2: Should we change this title to match the pattern in "heading" and "code"

packages/layouts/bolt-holy-grail/src/holy-grail.js
```diff
@@ -0,0 +1,15 @@
+var holyGrailSidebarToggleTrigger = document.querySelectorAll(
+  '.js-bolt-holy-grail-sidebar-toggle-trigger',
+);
+
+holyGrailSidebarToggleTrigger.forEach(function(el) {
+  el.addEventListener('click', function(event) {
+    if (el.getAttribute('aria-expanded') === 'true') {
```
colbytcook, line 7: @danielamorse same question as before

packages/components/bolt-content-pagination/content-pagination.schema.js
mikemai2awesome: Topic is too specific, maybe i can add `previous`, `next`, and `one level up`.

packages/components/bolt-content-pagination/src/content-pagination.twig
```diff
@@ -0,0 +1,77 @@
...
+  {{ validate_data_schema(schema, _self)|raw }}
+{% endif %}
+
+{# Variables #}
+{% set this = init(schema) %}
+{% set attributes = create_attribute(attributes|default({})) %}
+
+{# Template #}
+<ul {{ attributes.addClass('c-bolt-content-pagination') }}>
+  {% if parent %}
```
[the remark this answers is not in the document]
mikemai2awesome: as discussed, to be done later.

packages/global/styles/04-elements/_elements-misc.scss
mikemai2awesome: Updated to be consistent across the board.

packages/components/bolt-content-pagination/content-pagination.schema.js
mikemai2awesome: Added default texts.

danielamorse (approved):
Love all the new components! I think they're all important additions to the design system.

Approved, with a couple comments:
- Holy Grail layout doesn't adapt when you collapse the side nav here: `/pattern-lab/patterns/30-layouts-holy-grail-10-holy-grail-use-case-documentation/30-layouts-holy-grail-10-holy-grail-use-case-documentation.html`. It _does_ here: `/pattern-lab/patterns/50-pages-70-community-00-community-tech-doc/50-pages-70-community-00-community-tech-doc.html`. Is this a Pattern Lab thing?
- On small-screen, Secondary menu modal doesn't close after you click on a TOC item. What's the expected UX here? If it is for the modal to close, should we add an example of some glue code JS to be added on the Drupal side?

Side Nav JS looks good, well done! I just tweaked syntax to use ES6. I'm fine using `js-bolt-side-nav-toggle-trigger` as a selector here, but this is a case where we could also use data-attributes. I think we're still establishing conventions for non-web-component JS. I was thinking of `js-` classes as the convention for classes added by the user to control JS, not classes added by us. Though, I think page-header uses the `js-` convention already. Anyway, since it's 100% in our control let's roll with this for now and if we need to adjust it later it's no problem.

Worth noting that this JS will be loaded everywhere all the time. It's slim so not a big concern. Down the road, we might look into loading just what we need via Drupal libraries or other Webpack tricks.

packages/components/bolt-side-nav/src/side-nav-li.twig
```diff
@@ -0,0 +1,18 @@
+{# Variables #}
+{% set attributes = create_attribute(attributes|default({})) %}
+{% set _link_attributes = create_attribute(link.attributes|default({})) %}
+
+{# Template #}
+<li {{ attributes.addClass('c-bolt-side-nav__list-item') }}>
+  <a {{ _link_attributes.addClass('c-bolt-side-nav__link') }} {% if current %}aria-current="page"{% endif %}>
```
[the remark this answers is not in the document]
mikemai2awesome: I added the `without`.

packages/layouts/bolt-holy-grail/src/holy-grail.twig
```diff
@@ -0,0 +1,73 @@
...
+{# Array of classes based on the defined + default props #}
+{% set classes = [
+  'l-bolt-holy-grail',
+] %}
+
+<div {{ attributes.addClass(classes) }}>
+  {% if sidebar %}
+    {% set sidebar_classes = [
+      'l-bolt-holy-grail__sidebar',
+      sidebar.auto_width ? 'l-bolt-holy-grail__sidebar--auto-width',
```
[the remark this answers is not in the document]
mikemai2awesome: Ok, I see. @remydenton what do you think? Make sub-template or not?

packages/components/bolt-content-pagination/src/content-pagination.twig
```diff
@@ -0,0 +1,76 @@
+{# Schema Validation #}
+{% set schema = bolt.data.components['@bolt-components-content-pagination'].schema %}
+{% if enable_json_schema_validation %}
+  {{ validate_data_schema(schema, _self)|raw }}
+{% endif %}
+
+{# Variables #}
+{% set this = init(schema) %}
```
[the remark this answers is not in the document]
mikemai2awesome: Ok, let's keep for now.

packages/global/styles/04-elements/_elements-table.scss
```diff
@@ -0,0 +1,41 @@
+/* ------------------------------------ *\
+   Table resets
+\* ------------------------------------ */
+
+table {
```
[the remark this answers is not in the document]
mikemai2awesome: Nope, this actually takes care of layout shifts. You won't see any when loading a pure web component `<bolt-table>` now.

mikemai2awesome:
@danielamorse Thanks, I fixed the first bullet point, good catch. For the second one, we should figure out something to close it when user clicks on TOC, we'll create a ticket to enhance.

colbytcook:
@mikemai2awesome on the "/pattern-lab/?p=pages-community-tech-doc" page, the "Scroll to Top" button is not working with keyboard controls. I am not sure if this is expected or not but I wanted to call it out

mikemai2awesome:
I'll resume after I am back. More work to be done on holy grail layout and side nav.

packages/components/bolt-side-nav/src/side-nav.js
```diff
@@ -0,0 +1,13 @@
+const sideNavLinkToggleTrigger = document.querySelectorAll(
+  '[data-bolt-side-nav-toggle-trigger]',
```
remydenton, line 2: This has two problems.  We don't have to solve this immediately to merge this PR, but it is a building tech debt each time we do it.
1) javascript won't re-bind to any content that gets added with AJAX
2) It'll also be loaded on every page of every website, most of which won't have any `data-bolt-side-nav-toggle-trigger` attributes.

The answer in Drupal to `1` is to use Drupal behaviors so it can potentially be reattached whenever an AJX command runs, and the answer to `2` is to make it a library so that it can be conditionally loaded using back end logic.  Since we're not in Drupal, the answers are trickier 😄 .  

One brainstorm idea, what if we set up Bolt to support Drupal behaviors?  That would mean we rewrite this code (any similar functionality) to look like a standard Drupal behavior, like this:
```
Drupal.behaviors.boltSideNav = {
  attach: function (context, settings) {
    const sideNavLinkToggleTrigger = context.querySelectorAll('[data-bolt-side-nav-toggle-trigger]');
    sideNavLinkToggleTrigger.forEach(el => {
      el.addEventListener('click', event => {
        if (el.getAttribute('aria-expanded') === 'true') {
          el.setAttribute('aria-expanded', 'false');
        } else {
          el.setAttribute('aria-expanded', 'true');
        }
      });
    });
  },
}
```
Then we just need one place to attach behaviors globally in pattern lab.  That could look like this:
```
// A global function to attach behaviors, based on this snippet in Drupal core: https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/misc/drupal.es6.js#L157-172.
Drupal.attachBehaviors = function(context, settings) {
    context = context || document;
    settings = settings || {};
    const behaviors = Drupal.behaviors;
    // Execute all of them.
    Object.keys(behaviors || {}).forEach(i => {
      if (typeof behaviors[i].attach === 'function') {
        try {
          behaviors[i].attach(context, settings);
        } catch (e) {
        }
      }
    });
  };

// Attach behaviors once.
Drupal.attachBehaviors(document, {});
```

This idea is mostly for @danielamorse to consider.  It's not immediately needed for this PR.

packages/layouts/bolt-holy-grail/src/holy-grail.twig
```diff
@@ -0,0 +1,68 @@
...
+  {{ attributes|without('gutter')|without('padding-top')|without('padding-bottom') }} 
+  {% if this.data.gutter.value %}gutter="{{ this.data.gutter.value }}"{% endif %}
+  {% if this.data.padding_top.value %}padding-top="{{ this.data.padding_top.value }}"{% endif %} 
+  {% if this.data.padding_bottom.value %}padding-bottom="{{ this.data.padding_bottom.value }}"{% endif %}
+  {% if this.data.full_bleed.value %}full-bleed{% endif %}
+>
+  {% if sidebar %}
+    <bolt-holy-grail-sidebar {% if sidebar.auto_width %}auto-width{% endif %}>
+      <button type="button" aria-expanded="false" data-bolt-holy-grail-toggle-trigger>
+        <bolt-holy-grail-sidebar-trigger-text>
```
remydenton, line 22: What's the rationale for having all these custom elements like `<bolt-holy-grail-sidebar-trigger-text>`?  Wouldn't it be more consistent with previous work to just use BEM classes?

packages/layouts/bolt-holy-grail/src/holy-grail.twig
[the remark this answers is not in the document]
remydenton: I'm not that opinionated for or against, but I do think we should have a policy about this and stick to it.

`init()` standardizes the way we set defaults and validate props, but if we use it inconsistently, then the standardization argument goes out the window.  I was looking for precedents and came across [the following in tab panel](https://github.com/boltdesignsystem/bolt/blob/release/3.x/packages/components/bolt-tabs/src/TabPanel/tab-panel.twig#L15-L16):

```
{% set classes = [
  "c-bolt-tab-panel",
  this.data.align.value ? "c-bolt-tabs--align-" ~ this.data.align.value : "",
  panel_spacing ? "c-bolt-tab-panel--spacing-" ~ panel_spacing : "",
%}
```

It looks like `this` is used for `align`, but not for `panel_spacing`.  Am I missing the rationale for that @danielamorse?  Obviously a tangent from the current PR, the broader question is what our rubrik is for using `this()`.

packages/layouts/bolt-holy-grail/src/holy-grail.twig
mikemai2awesome: In the Layout POC, we used all custom elements, so I was trying to stick with that convention. If we decide to not do that for Layouts, then we can change it all back. The initial idea is for all Layouts to be `<bolt-layout-*>`.

packages/layouts/bolt-holy-grail/src/holy-grail.twig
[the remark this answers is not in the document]
danielamorse: @remydenton Tabs is wrong in a couple ways. We're using `panel_spacing` because it's [passed in as data](https://github.com/boltdesignsystem/bolt/blob/release/3.x/packages/components/bolt-tabs/src/tabs.twig#L47-L52) from the parent. It's [not actually a prop](https://github.com/boltdesignsystem/bolt/blob/release/3.x/packages/components/bolt-tabs/tabs.schema.js#L12-L34) of Tab Panel. We don't want to pass data like this, and we don't want to mix `this.data` with regular data.

`init()` doesn't work well with nested props so in order to use it consistently, every time we hit a nested prop we'd have to call `init()` again (in a sub-template or assigned to a new var). Organizing our templates around the limitations of `init()` doesn't necessarily sound like a good idea.

`init()` works best with Web Components and their props, which are never nested. We might be trying to use the wrong tool to validate our props on non Web Components. 

I don't have a good answer atm. It's something I'm thinking about too as we build more non-WC components. Let talk more next week.

mikemai2awesome:
@remydenton @danielamorse @colbytcook I converted holy grail layout to BEM classes and commented out docs for Side Nav and Holy Grail as I am still working on it. We can merge this as is. Side Nav and Holy Grail will be uncommented once they are ready.

packages/layouts/bolt-holy-grail/src/holy-grail.twig
mikemai2awesome: Converted to BEM classes.

packages/components/bolt-side-nav/src/side-nav.js
mikemai2awesome: This was discussed in huddle.

packages/layouts/bolt-holy-grail/src/holy-grail.twig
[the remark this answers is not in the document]
mikemai2awesome: Still working on Holy Grail, so I'll revisit later.

danielamorse (approved):
The recent HTML and CSS changes look good at first glance. Everything is hidden so I will wait until later to do a final review. This is ok to merge as is for now.

remydenton (approved):
The basics also look good to me, I say go ahead and merge to avoid further conflicts and work out the TODOs in follow up PRs.

colbytcook (approved):

PR #2149, opened 2021-03-30
WWW-467: Remove separate hard-coded classes on desktop heading links

remydenton:
## Jira

Related to https://pegadigitalit.atlassian.net/browse/WWW-467, though that doesn't describe the bug

## Summary

Fixes attributes in page header links.

## Details

The bug this fixes is that custom classes cannot be passed in `link.attributes` without breaking the two existing hard-coded CSS classes, `c-bolt-page-header__nav-link` and `c-bolt-page-header__nav-link--heading`.

Whenever they're used, `Attributes` should be expected to potentially contain all HTML attributes.  Use `addClass()` to add additional classes, or `|without()` if you wish to print some attributes separately.

### Confirm the bug

- On master, create a page header link that adds a custom class using link.attributes 
```
{% include '@bolt-components-page-header/page-header-nav-li.twig' with {
  link: {
    content: 'Foo',
    desktop_heading: true,
    attributes: {
      class: 'bar',
      href: 'http://pega.com',
    },
    children: 'foo'
  }
} only %}
```
- Confirm that the `c-bolt-page-header__nav-link` and `c-bolt-page-header__nav-link--heading` classes do not appear on the link.

### Confirm the fix
- Checkout this branch and repeat the above.  The `c-bolt-page-header__nav-link` and `c-bolt-page-header__nav-link--heading` classes should not be removed.

### DIFF
--- a/packages/components/bolt-page-header/src/page-header-nav-li.twig
+++ b/packages/components/bolt-page-header/src/page-header-nav-li.twig
@@ -23,6 +23,11 @@
   current ? 'is-current',
 ] %}
 
+{% set link_classes = [
+  'c-bolt-page-header__nav-link',
+  link.desktop_heading ? 'c-bolt-page-header__nav-link--heading',
+] %}
+
 {# Template #}
 <li {{ attributes.addClass(classes) }}>
   {% if content and not link %}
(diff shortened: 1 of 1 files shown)


mikemai2awesome (changes requested):
This breaks the mobile view. Let me investigate.

mikemai2awesome:
@danielamorse @remydenton This was solved in our call.

remydenton:
Looks great, thanks @mikemai2awesome.

Hold off on merging this one pending a potential tooltip hotfix from @danielamorse.

PR #2150, opened 2021-03-31
DS-439 Add `inverted` option to Tooltip

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-439

## Summary

Add `inverted` option to Tooltip.

## Details

Adds `inverted` option to Tooltip so that user can invert colors when tooltip renders over a dark background.

## How to test

- Checkout feature branch and view "Inverted" Tooltip demo.
- Verify other non-inverted Tooltip demos continue to work as expected.

## Release notes

If you are using Tooltip over a dark background (for example a Band with theme `xdark`) you must manually set `inverted: true` on the Tooltip for the Tooltip to have enough contrast with the background color. This used to happen automatically, but now it is a manual step because Tooltip is rendered at the end of the page where the theme variables cannot be inherited.


mikemai2awesome (approved):
@danielamorse I updated the docs. Approved.

PR #2152, opened 2021-03-31
DS-427 Enhance Teaser description ellipsis display

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-427

## Summary

Improves the ellipsis display for Teasers that have short description text.

## Details

1. Updated CSS to display ellipsis based on a class
2. Updated Twig to include logic that checks the description content length
3. This is not perfect but should cover 99% of use cases (the other alternative is to use complex JS to count the characters and lines, I don't think it's worth it)

## How to test

Run the branch locally and check the Teaser description on hover doc. Play around with the description content, try different length above and below 80 characters. Make sure it is working properly.


cjwhitedev (approved):

danielamorse (changes requested):
@colbytcook it looks like Clamp JS just adds inline styles that we could add ourselves without JS:
image

You can find the same rules in [this CSS-Tricks article](https://css-tricks.com/almanac/properties/l/line-clamp/):
```
  display: -webkit-box;
  -webkit-line-clamp: 3;
  -webkit-box-orient: vertical;  
  overflow: hidden;
```

Clamp JS is helpful if you support IE11 but for all the browsers we support (including Edge 18) it just [inlines CSS](https://github.com/josephschmitt/Clamp.js/blob/master/clamp.js#L236-L246) if you support clamp natively.

Let me know if I'm missing something. AFAICT, you should be able to add those line-clamp rules to [teaser.scss](https://github.com/boltdesignsystem/bolt/blob/891023a4af8dcbeeb298605e1f5c496fc98ef0da/packages/components/bolt-teaser/src/teaser.scss#L199-L216) and call it a day.

danielamorse (approved):
Made [a tweak](https://github.com/boltdesignsystem/bolt/pull/2152/commits/1331698843675b7c4ca12d98d116217228eb2cfb) to bring back the centering on shorter descriptions.

This is looking good to me and working in all major browsers.

@mikemai2awesome is there a reason you didn't go with line-clamp initially?

PR #2156, opened 2021-04-06
DS-447 Fix broken `renderWC` promise

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-447

## Summary

Fix broken promise in renderWC to fix intermittent Jest snapshot failures, especially on the Carousel component.

## Details

[This promise](https://github.com/boltdesignsystem/bolt/blob/master/packages/testing/testing-helpers/index.js#L53) in renderWC does not work because `customElements` and `whenDefined` objects do not exist within the Jest page context.

Because this promise does not actually wait, the HTML returned may vary, resulting in intermittent snapshot failures. If we fix this broken promise, Jest test snapshots that use `renderWC` should return consistent results, or at least fail less often and for different reasons.

## How to test

- Review files changed
- Tests are passing


colbytcook (approved):
The code was reviewed in the PR and the dev huddle and all the tests pass on my local

Approved.

PR #2158, opened 2021-04-09
DS-418: 2.30 regression: Image modal scroll dragging overlay up

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-418

## Summary

Modal scroll fix

## Details

The fix prevents a scroll but from appearing when a image is taller than the viewport

## How to test

Navigate to "/pattern-lab/?p=components-modal-usage-image-and-caption" and open up the "Enlarge image and show caption" modal and scroll down. The modal overlay should not scroll up with the page.


danielamorse (approved):
Works as expected cross-browser. Approved 👍

PR #2185, opened 2021-04-21
DS-411: Component Test Refactor

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-411

## Summary

Restructured the tabs jest test to organize the jest test into a more scalable pattern

## How to test

On your local run "npx jest packages/components/bolt-tabs/__tests__/tabs.js " 
Review the tabs snapshot


danielamorse (approved):
Looks great, approved.

I had to fix one unrelated test locally that was failing due to timeout (https://github.com/boltdesignsystem/bolt/pull/2185/commits/d727e7b0dbc3ae22f9a11440e08c13f5bfb9ec02). Removing the timeout here lets it use the default, which is longer.

PR #2186, opened 2021-04-21
Bugfix/ds 458 listing component fix auto escaping issues

adamszalapski:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-458

## Summary

Fix auto escaping after integration to Drupal

## Details

We handle concatenation in a different way that is better handled in Drupal.

## How to test

It should be tested on the Drupal website.

### DIFF
--- a/packages/components/bolt-listing-teaser/src/listing-teaser.twig
+++ b/packages/components/bolt-listing-teaser/src/listing-teaser.twig
@@ -49,11 +49,11 @@
       {% endif %}
       {% if this.data.more_info.value %}
         <div class="c-bolt-listing-teaser__flag-content__item">
-          {% set _icon_chevron_right %}
+          {% set _icon_chevron_right %}{% spaceless %}
             {% include '@bolt-components-icon/icon.twig' with {
               name: 'chevron-right',
             } only %}
-          {% endset %}
+          {% endspaceless %}{% endset %}
           {% set _more_info_content %}
             {{ this.data.more_info.value }} &hellip; {{ 'Read More'|t }}{{ _icon_chevron_right }}
           {%  endset %}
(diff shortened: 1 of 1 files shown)


mikemai2awesome (approved):

packages/components/bolt-listing-teaser/src/listing-teaser.twig
```diff
@@ -54,11 +54,11 @@
               name: 'chevron-right',
             } only %}
           {% endset %}
-          {% set _read_more_link %}
-            {{ 'Read More'|t }}{{ _icon_chevron_right|spaceless }}
-          {% endset %}
+          {% set _more_info_content %}
+            {{ this.data.more_info.value }} &hellip; {{ 'Read More'|t }}{{ _icon_chevron_right }}
```
remydenton, line 58: Any reason you didn't include `|spaceless` after `_icon_chevron_right`?  I think that shouldn't cause autoescaping issues, though I'm not entirely sure why it's necessary.  @mikemai2awesome?

remydenton (approved):

packages/components/bolt-listing-teaser/src/listing-teaser.twig
adamszalapski: In my opinion, spaceless removes spaces, and I don't see any point to use spaceless for icon and read more if we rendered them without any space, but if you guys thinks that I should add it I can do that.

mikemai2awesome: I double checked this and tried to put spaceless in different places, didn't seem to get it to work. I didn't want to hold this up, but keep in mind the chevron can be pushed to the next line if we don't use spaceless due to a white space between text and icon. Where can we put spaceless to make this work? @remydenton @adamszalapski

danielamorse (approved):

packages/components/bolt-listing-teaser/src/listing-teaser.twig
remydenton: As discussed with @mikemai2awesome on Slack, we still have some widowing to take care of, but it's a problem on master right now anyway, so let's merge this and @mikemai2awesome will figure it out later 😄 .

PR #2192, opened 2021-04-23
CC-472 Add reply prop and rename more_info prop to description in Listing Teaser

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/CC-472

## Summary

Adds a `reply` prop, and rename `more_info` to `description` in the Listing Teaser component

## Details

1. Added `reply` prop to showcase a relevant reply.
2. Updated CSS to style the reply content.
3. Added docs to demonstrate the `reply` prop.
4. `more_info` has been renamed to `description`, its type has been changed to `any`.
5. `warning` prop's type has been changed to `any`.

## How to test

Run the branch locally and read the Listing Teaser props, especially the one with reply prop and Collaboration Center use case. Make sure the prop works as expected.


packages/components/bolt-listing-teaser/src/listing-teaser.twig
```diff
@@ -82,6 +82,26 @@
           } only %}
         </div>
       {% endif %}
+      {% if this.data.reply %}
```
adamszalapski, line 85: We shouldn't use `this.data` for checking content different than `string`. @danielamorse, @remydenton please correct me if I'm wrong. In this case, we should use `{% if reply %}`

packages/components/bolt-listing-teaser/src/listing-teaser.twig
```diff
@@ -82,6 +82,26 @@
...
+          {% set _icon_chevron_right %}{% spaceless %}
+            {% include '@bolt-components-icon/icon.twig' with {
+              name: 'chevron-right',
+            } only %}
+          {% endspaceless %}{% endset %}
+          {% set _reply_link %}
+            <span class="e-bolt-text-link e-bolt-text-link--reversed-underline">{{ 'Go to reply'|t }}<span class="e-bolt-text-link__icon-after">{{ _icon_chevron_right }}</span></span>
+          {%  endset %}
+          {% include '@bolt-components-stack/stack.twig' with {
+            content: this.data.reply,
```
adamszalapski, line 96: The same issue here we should pass `content: reply`

packages/components/bolt-listing-teaser/src/listing-teaser.twig
```diff
@@ -82,6 +82,26 @@
...
         </div>
       {% endif %}
+      {% if this.data.reply %}
+        <div class="c-bolt-listing-teaser__flag-content__item c-bolt-listing-teaser__flag-content__item--reply">
+          {% set _icon_chevron_right %}{% spaceless %}
+            {% include '@bolt-components-icon/icon.twig' with {
+              name: 'chevron-right',
+            } only %}
+          {% endspaceless %}{% endset %}
+          {% set _reply_link %}
```
adamszalapski, line 92: I noticed that right now `Go to reply` will point to the same place as a whole row (whole row clickability) but what I heard was that header should link to the post where the answer was added and `Go to replay` should point to that reply directly. I don't know if you know it @mikemai2awesome or maybe you agreed that we will change the existing approach.

adamszalapski (changes requested):
@mikemai2awesome I add few comments, please check.

packages/components/bolt-listing-teaser/src/listing-teaser.twig
adamszalapski: Maybe we should add `reply_link` prop?

mikemai2awesome: thanks for the heads up. let me make some changes

packages/components/bolt-listing-teaser/src/listing-teaser.twig
mikemai2awesome: all set

packages/components/bolt-listing-teaser/src/listing-teaser.twig
mikemai2awesome: all set

packages/components/bolt-listing-teaser/src/listing-teaser.twig
mikemai2awesome: prop is updated

adamszalapski (approved):
Looks good, great job @mikemai2awesome

colbytcook (approved):

remydenton (approved):
Looks great, nice work.

adamszalapski (approved):
Everything ok, nice work @mikemai2awesome

remydenton (approved):

PR #2193, opened 2021-04-26
DS-407: Update modal demos to use [data-bolt-modal-target] convention

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-407

## Summary

Update Modal demos and docs to use the `[data-bolt-modal-target]` convention. Update copy in demos.

## Details

I updated the Modal component demos and instances of Modal in other component demos and "Pages". I also updated some of the copy in the Modal component docs to reflect the new recommended convention.

> Note: fixes a [Listing Teaser dependency bug](https://github.com/boltdesignsystem/bolt/pull/2193/files#diff-dc47f6b3774425192c0192d558ad761b8a5bc1689442d96a3533ffbf643ca20eL20) which was required to get e2e tests to pass.

## How to test

- Review [Modal docs](https://feature-ds-407-update-modal-docs-demos.boltdesignsystem.com/pattern-lab/?p=viewall-components-modal). Verify modals still toggle as expected. Review copy changes.
- Review other modified component docs and pages:
    - [Device Viewer](https://feature-ds-407-update-modal-docs-demos.boltdesignsystem.com/pattern-lab/?p=components-device-viewer-content-magnifier-modal)
    - [Trigger Component](https://feature-ds-407-update-modal-docs-demos.boltdesignsystem.com/pattern-lab/?p=components-trigger-advanced-usage)
    - [Calculator Results](https://feature-ds-407-update-modal-docs-demos.boltdesignsystem.com/pattern-lab/?p=pages-d8-results) (click "email my results")
    - [Academy Topic Detail](https://feature-ds-407-update-modal-docs-demos.boltdesignsystem.com/pattern-lab/?p=pages-acd-topic-detail-page) (click first article image)
    - [BOC Gallery](https://feature-ds-407-update-modal-docs-demos.boltdesignsystem.com/pattern-lab/?p=pages-boc-gallery-phase1) (click "more filters")
    - [BOC Listing](https://feature-ds-407-update-modal-docs-demos.boltdesignsystem.com/pattern-lab/?p=pages-boc-listing-pegaworld-phase1) (click "filters")
    
## Release Notes
- The `[on-click]`/`[on-click-target]` method of toggling a modal is deprecated. Instead use `[data-bolt-modal-target]`. See [the docs](https://feature-ds-407-update-modal-docs-demos.boltdesignsystem.com/pattern-lab/?p=components-modal-usage-javascript) for more info (TODO: update link after `v3.6.0` release).

PR #2195, opened 2021-04-26
WWW-739: Video Load Update

colbytcook:
## JIRA

https://pegadigitalit.atlassian.net/browse/WWW-739

## Summary

Reworked the modal video behavior to have the videos load when a modal is opened

## Details

- Created a conditional that will load a video immediately if it does not have a 'data-video-modal' attribute on the video (so inline videos will immediately initialize)
- Created a function to handle all the Pega specific initialization (initialize plugins...)
- Added functionality that will initialize and autoplay the video when the "modal:show" event fires
- Added functionality that will pause the video on the "modal:hide" event

## How to test

- Pull down the branch
- Make sure the video are being initialized on the AFTER modal open on "/pattern-lab/?p=pages-boc-gallery-phase2" (Observe that you see the video ID listed in the Network/XHR panel after the modal is opened)
- Do a cursory check on other video modal pages


remydenton (changes requested):
This looks good, and basically matches what we have in Drupal.

However, since our new policy and future plan are to make all modal videos work like this, I think we need to update the documentation at https://feature-video-modal-load.boltdesignsystem.com/pattern-lab/?p=components-modal-usage-video.  It is actually already out-of-date [on master](https://boltdesignsystem.com/pattern-lab/?p=components-modal-usage-video), in that the second example doesn't do anything that it says it does -- the button launches the modal, and the javascript that is supposed to play and pause does nothing at all.  And with the new updates in this PR, I don't the first example isn't actually using the javascript it claims it is either.

At the very least, we should remove anything on that page that is no longer accurate.  Ideally, also take the opportunity to document the new best practices for modal videos using `data-video-delayed`.

danielamorse:
@remydenton @colbytcook just a heads up that I made some edits to the Modal Video demo page as part of PR #2193, see: https://github.com/boltdesignsystem/bolt/pull/2193/commits/d157f19b6642f546c63673ebbd52a7232c0bfbe9. I updated the demo to use the new `[data-bolt-modal-target]` convention. If this is not appropriate for video modals feel free to overwrite what I did.

Also, to get these demos working I had to move references to `videojs` inside the click handler because Brightcove is loaded after this script is run, so `videojs` is undefined otherwise. We should probably move all video player JS into [the pattern lab JS file](https://github.com/boltdesignsystem/bolt/blob/master/docs-site/src/index.js) that gets included after the Brightcove script, and then copy code snippets into our demos.

remydenton:
Thanks @danielamorse.  I believe switching to `data-bolt-modal-target` shouldn't affect this-- basically, modal videos are agnostic to _how_ their parent modal is opened, just _that_ it is opened (i.e. the video listens for `modal:show` and `modal:hide` events and doesn't care about the distinction between `data-bolt-modal-target` and `on-click`/`on-click-target`).  @colbytcook, please confirm.

+1 to moving this into the pattern lab js file.

PR #2198, opened 2021-05-06
DS-248 Rebuild Navbar

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-248

## Summary

Consolidate "navbar", "nav-indicator", "nav-priority", "navlink" all into a single component, "navbar". Fix bugs related to scroll offset.

## Details

PR includes the following:
- Remove components: nav-indicator, nav-priority, navlink (https://github.com/boltdesignsystem/bolt/commit/ac0950c688e9f6075e3dab41ae5b82c5d8dc8c48)
- Update Navbar component (https://github.com/boltdesignsystem/bolt/commit/8e942342afb74ea3484bb656f66cae2f19408160)
- Update Page Header to add flag for sticky-offset (https://github.com/boltdesignsystem/bolt/commit/a0f62eb42a1ad0e741f3cf02aade89b421cd7ebe)
- Update smoothscroll to include flag for opting out of default behavior (https://github.com/boltdesignsystem/bolt/commit/2e772f41dab1c038a18fa3a51cfe64e523416ee9)
- Update Waypoint to only fire on position change to improve performance (https://github.com/boltdesignsystem/bolt/commit/2ba15a74c1dbe709f0b02b5336f9f5afe0050da0)
- Update selector in autotrack analytics component (https://github.com/boltdesignsystem/bolt/commit/1b2bd9f13f7270e75550d718dfce1d3f7d2351c7)
- Update Navbar docs (https://github.com/boltdesignsystem/bolt/commit/99089838e9d2aa2942e11bc3fb2288f0d949256f)
- Update Navbar on page demos and homepage (https://github.com/boltdesignsystem/bolt/commit/7dfc4812a65e82e0d21d57d9ed2be197e390e6c4)
- Remove unnecessary Sticky Jest test (https://github.com/boltdesignsystem/bolt/commit/e262a0e161833d7d20dbff055ad38ab74cf2addb)

Additional notes:
- Removed "short" demo
- Removed "transparent" demo, covered in "theme" demo
- Add `title[secondary_content]` prop to accommodate Navbar on PL homepage without use of a block

## How to test

- Review all Navbar demos and docs
- Review changes to schema

Caveats:
- Navbar tests are currently being updated.
- More comprehensive 4.x release notes will be prepared when this is actually released.

### Navbar
- Nav Priority, Nav Indicator, and Navlink components have been consolidated into Navbar (with 3 sub templates, see docs for details)
- Navbar is "sticky" by default. Do not put it inside a Bolt Sticky component. Use `static` prop to disable "sticky".
- `links` is now type `any`. Pass it renderable content, not an array of data. See docs for usage.
- Removed `theme: false`. Use `theme: 'none'` to disable theme.
- Remove `moreText` prop. Hardcoded to "More". Though, still translated.
- Removed `title['text']`. Use `title['content']` instead.
- Renamed `title[hide_title]` to `title[hide_text]`
- Added `content` prop for optional custom content that will replace `title` and `links` templates.
- Added `spacing` prop.

### Visual changes

- Navbar active item indicator no longer animates laterally. It simply fades in-out.
- Navbar overflow ("More") menu no longer slides down. It simply fades in-out.


mikemai2awesome:
@danielamorse I reverted the `width` prop because it was an unnecessary breaking change, and default to `true` is against our conventions. Also fixed some spacing in the CSS.

I'm leaving the following for you to fix:

1. When closing the dropdown, the text is jumping.
    * jump
 
2. We need deprecations for all the prop changes.

packages/components/bolt-navbar/src/navbar.twig
```diff
@@ -1,38 +1,64 @@
...
+{% if this.data.scroll_offset_selector.value %}
+  {% set attributes = attributes.setAttribute('data-bolt-scroll-offset-selector', this.data.scroll_offset_selector.value) %}
+{% endif %}
+
+{% if this.data.scroll_offset.value %}
+  {% set attributes = attributes.setAttribute('data-bolt-scroll-offset', this.data.scroll_offset.value) %}
+{% endif %}
+
+{% if this.data.more_text.value %}
+  {% set attributes = attributes.setAttribute('data-bolt-more-text', this.data.more_text.value|t) %}
```
remydenton, line 43: This won't work-- German translators, for example, will be asked to provide a translation of "this.data.more_text.value".

You don't need to use `|t` with variables because whatever string is being passed in will have already been translated.  This isn't the case with the default of "More" though, because that originates in [a bolt yml file](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-navbar/navbar.schema.js#L74).

I actually don't think there's even a use case to change this text-- it can always be "More".  It just needs to be translatable.  We can just hard-code it in twig, like this:

{{ 'More'|t }}

When it comes to javascript, you can either a) grab the string from twig (where it's already translated) or b) use `Drupal.t('More')` (which is the js version of the t function).  I think we previously had this as a prop on the web component so that it could come from twig, and we wouldn't have to use `Drupal.t()` so that we could be "CMS agnostic".  But since we've long since decided that's a dumb idea, maybe `Drupal.t()` is OK to use in Bolt itself...

packages/components/bolt-navbar/src/overflow-menu.js
```diff
@@ -0,0 +1,206 @@
...
+export class BoltOverflowMenu {
+  constructor(el, options = {}) {
+    if (!el) return;
+
+    this.el = el;
+    this.options = {
+      items: [],
+      container: null,
+      baseClass: 'js-bolt-overflow-menu',
+      moreText: 'More',
```
remydenton, line 13: This 'More' won't be translatable here.  Either grab it from twig or use `Drupal.t('More')`.

danielamorse:
@mikemai2awesome,

> I reverted the `width` prop

Makes sense 🙏 

> 1. When closing the dropdown, the text is jumping.

Fixed

> 2. We need deprecations for all the prop changes.

Discussed with @remydenton and we're going to skip deprecations in this case. Not much of a point given the timeline.

> ## EDIT: dropping another bomb, lol.

Updated the schema and template names to be consisten with Page Header. Good call.

packages/components/bolt-navbar/src/overflow-menu.js
danielamorse: This is just a default set in the JS. It is overridden by the data-attribute from Twig [here](https://github.com/boltdesignsystem/bolt/pull/2198/files#diff-3fb4dcc466a19eb8b1f18db88d7c436c7fc9ed8dda9150dbb71d8bfaad928a59R188).

packages/components/bolt-navbar/src/navbar.twig
danielamorse: Ah, ok. Thanks for explaining. I removed the `more_text` prop and hardcoded it (with translation) [here](https://github.com/boltdesignsystem/bolt/pull/2198/files#diff-152935d43dcc3684234eadbf3baf5a4baacbb2b8624cc796db4ab99cd7f42560R45).

I'm open to `Drupal.t()` but I'd like to try it somewhere else first, when there are fewer moving parts. There's already so much new work here, just trying to contain it somewhat.

packages/components/bolt-navbar/navbar.schema.js
```diff
@@ -3,25 +3,21 @@ module.exports = {
...
-          default: 'h2',
-          enum: ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'],
+        content: {
+          type: 'any',
+          default: 'Go to',
+          description:
+            'Title text. Should be plain-text but may contain some HTML. Defaults to "Go to" on small screens if empty.',
         },
-        text: {
+        tag: {
```
colbytcook, line 17: referring to this PR (https://github.com/boltdesignsystem/bolt/pull/2226) we are updating the "tag" to be "type", is that something we should be doing here?

colbytcook: Disregard, these are different use cases

colbytcook (approved):

PR #2204, opened 2021-05-12
WWW-476: Share URL Update

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-476

## Summary

Deprecated the URL prop on the Share component to allow for more flexible Drupal integration

## Details

- Added 'attributes' to the share and menu item items schema to allow for data attributes to the added to the bolt-share on both the "display="inline" and display="menu" types.
- Rewrote the share component jest test to the latest standards (added a deprecated test and added a 'display' prop test)
- Updated the Teaser jest test to the new snapshot
- Updated the doc site with the new share schema structure

## How to test

Pull down this branch and confirm that the share functionality is still work

## Release notes

Able to pass in data attributes to the share component for custom functionality


cjwhitedev (approved):

mikemai2awesome (approved):

danielamorse (approved):
@colbytcook I had to update twig to fix the `url` prop breaking when the `menu` display was used. It worked for `attributes.href` but not for deprecated `url`. I added a test for that case. I also move some of the test data out of the docs into the Tests section.

The new test file is looking good!

Please retest just to verify it all still works as expected.

PR #2206, opened 2021-05-13
DS-435 Standardize Twig and Schema (Card)

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-435](https://pegadigitalit.atlassian.net/browse/DS-435)

## Summary

Standardize Twig and Schema for Card

## Details

Standardize Twig and Schema for Card. The snapshots have been updated because the default values of properties no longer need to be printed. For more detailed information, please see the initial ticket, as well as the associate TWIG and Schema tickets.

## How to test

Check out the tests and demos and make sure nothing is a breaking change in the code.

## Release notes

Card `borderRadius` prop deprecated for `border_radius`.


packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig
```diff
@@ -1,36 +1,42 @@
-<bolt-card-replacement-action {% if external %} external {% endif %} {% if url %} url="{{url}}" {% endif %}>
+{% set schema = bolt.data.components['@bolt-components-card-replacement'].schema.properties.actions.items %}
+{% set this = init(schema) %}
+
+<bolt-card-replacement-action
+  {% if this.data.external.value %} external {% endif %}
+  {% if this.data.url.value %} url="{{this.data.url.value}}" {% endif %}
+  >
```
danielamorse, line 7: We should be able to use `this.props` here. Are you intentionally not doing that because `attributes` are passed below into the Button template and then they'd be applied to both elements? 🤯 💩

danielamorse (approved):
Nice work. Everything is working as expected, which is no small feat for such a complicated component.

I made the following changes:

- Formatting (https://github.com/boltdesignsystem/bolt/pull/2206/commits/2e77094eab7921776d7d8bb08326f935cec6d1bd)
- Schema fixes (https://github.com/boltdesignsystem/bolt/pull/2206/commits/65aa9dd4d06b9264f5412c3c91047dc5354e303f):
  - Added missing `attributes` props to schema (was unofficially supported)
  - Added missing `items` key (arrays [require `items`](https://json-schema.org/understanding-json-schema/reference/array.html#id4) for validation)
  - Added missing descriptions
- Reverted the card-replacement-body conversion to use `init()`, more below (https://github.com/boltdesignsystem/bolt/pull/2206/commits/61e78f411294a3df3b3e294292487fbb240bf5d1)
- Added a "test" page for stress testing (https://github.com/boltdesignsystem/bolt/pull/2206/commits/ccaf57294f69aea6a53f1f9006c94564084a34e8)

The [card-replacement-body template](https://github.com/boltdesignsystem/bolt/pull/2206/files#diff-1c4fba138d8724fa82178805fc1c1761d607125c89693f4167418974b207c63b) is not a good candidate for `init()` because it also expects props from the parent and hence requires more than one schema. For `init()` to work properly it needs just the context of that template and its schema. This should be reworked, but not as part of this task. I appreciate the effort getting that setup, but I rolled it back. 

I'd like to get one more review and then it's good to merge.

colbytcook:
@remydenton @danielamorse @remydenton the Jest test has been updated, feel free to check it out

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig
cjwhitedev: well, i initially did it because there's a bunch of props that don't need to be printed like `text` and some of the `on-click` stuff. so i was getting failing tests, and it made sense to do it this way, but it could be fixed by using `without`, so ill have another commit up today fixing that.

danielamorse: Sorry, my comment was confusing - we _should_ be able to use `this.props` here but we can't because we're applying attributes [to the inner button element](https://github.com/boltdesignsystem/bolt/blob/724ac462f676f1a4620b0f8b0ef0f4ec0eded4f4/packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig#L18). `this.props` would put attributes on two elements introducing a potential breaking change. `without` filter can't fix it because the attributes could be any arbitrary attributes.

remydenton:
Sorry for sitting on this for so long!

Overall, it looks really good.  I did have one thing while we're refactoring: would it make sense to make the subcomponent twig templates _not_ private?  I think the main actionable item would be removing the leading underscore from these twig templates (and leaving the old ones in there as adaptors for backwards compatibility).  It's kind of a pain in the butt and ultimately just semantic, but if we're doing cleanup, it'd be nice to get it right.

The main use case is using these components independently to assemble your own card.  Piotr Muszynski wanted to build his own modified horizontal card and the most expedient way was to include these files directly:  https://gitlab.com/pegadigital/bolt/pega_bolt/-/blob/feature/WWW-329-new-brightcove-component/features/pega_bolt_card/templates/horizontal-image-on-right.twig.  I don't see any obvious reason why this _shouldn't_ be possible.  @danielamorse , any thoughts?

danielamorse:
> would it make sense to make the subcomponent twig templates _not_ private?

It would make sense if this was built to our current standards. The problem I see is that the subcomponents  are not completely self-contained. We don't use `only` when we include them from the main card template, and they expect some props from the parent. The only example I know for sure is [`spacing`](https://github.com/boltdesignsystem/bolt/pull/2206/files?file-filters%5B%5D=.js&file-filters%5B%5D=.twig#diff-1c4fba138d8724fa82178805fc1c1761d607125c89693f4167418974b207c63bR1-R14). It may in fact be the only example.

If we make these templates public we should also make inherited props like `spacing` explicit in the schema and convert to use `only`. @remydenton do you think that could be done safely in this PR? I think it probably could be, but I'd rather not disturb this component any further 😆 . I'm not opposed if you want to go for it - now or later.

remydenton:
Per discussion in dev huddle, making sub-components like this externally usable is a practice we want to do going forward, but the value of adding it to card (which may soon be replaced by teaser) is limited.   We're going to merge this PR as-is since it adds value and focus on keeping sub templates public for new components.

remydenton (commented):
Approved!

PR #2210, opened 2021-05-17
DS-435: Card Test Update

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-435

## Summary

Updated the card test with the new standard

## How to test

* Pull down the branch
* Run npx jest packages/components/bolt-card-replacement/__tests__/card-replacement.js


danielamorse (approved):
Beautiful tests! I can read them and understand exactly what is going on 🙏 🙏 🙏 

Also, checked the snapshots and the markup looks correct.

We can merge once it passes on Travis. This will go with `4.0`.

PR #2211, opened 2021-05-18
Update the z-index for tooltip and popover

adamszalapski:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-479

## Summary

Update default `z-index` to match earlier default for tooltip and popover.

## Details

`zIndex` prop was added to tippy call.

## How to test

Open popover and tooltip demo and confirm that element created by tippy on the bottom of the page use:
- `z-index: 120` for popovers

image
- `z-index: 140` for tooltips

image


danielamorse (commented):
@adamszalapski, thanks for raising this issue and making the PR.

As I mentioned on Slack, I have a few concerns. If we set z-index to 120/140 we'll have to account for the total height of sticky content. Otherwise, it will display "top" when there's not actually room for it to be visible (covered by sticky elements).

This can be handled via [popperOptions](https://atomiks.github.io/tippyjs/v6/all-props/#popperoptions) and the [detectOverflow/padding](https://popper.js.org/docs/v2/utils/detect-overflow//#padding). Additionally, we'll need a system for knowing what sticky elements are actually "pinned".

We might want to coordinate this with the release of Navbar and Page Header. Let's talk more in grooming/dev huddle.

mikemai2awesome (changes requested):
what @danielamorse said and also this isn't using our z-index system. we do not want to manually write random z-index values, we need to find a way to use the system.

packages/components/bolt-popover/src/popover.js
```diff
@@ -15,12 +15,30 @@ class BoltPopover extends BoltElement {
...
+      paddingTop: { type: Number },
     };
   }
 
   constructor() {
     super();
     this.uuid = this.uuid || Math.floor(10000 + Math.random() * 90000);
+
+    this.zIndex = parseInt(
+      getComputedStyle(this).getPropertyValue('--c-bolt-popover-zindex'),
```
mikemai2awesome, line 28: Let's use `--c-bolt-popover-z-index` instead to be consistent with how z-index is spelled.

packages/components/bolt-popover/src/popover.scss
```diff
@@ -12,6 +12,7 @@
 
 //  Note: this does not actually render to the Shadow DOM so :host rule not required
 bolt-popover {
+  --c-bolt-popover-zindex: #{bolt-z-index('popover')};
```
mikemai2awesome, line 15: Same comment on the naming.

packages/components/bolt-tooltip/src/tooltip.scss
```diff
@@ -10,6 +10,7 @@
  */
 //  Note: this does not actually render to the Shadow DOM so :host rule not required
 bolt-tooltip {
+  --c-bolt-tooltip-zindex: #{bolt-z-index('tooltip')};
```
mikemai2awesome, line 13: Same comment on the naming.

mikemai2awesome (changes requested):

packages/components/bolt-popover/src/popover.js
```diff
@@ -15,12 +15,30 @@ class BoltPopover extends BoltElement {
...
+
+    this.zIndex = parseInt(
+      getComputedStyle(this).getPropertyValue('--c-bolt-popover-z-index'),
+      10,
+    );
+
+    let offsetHeight = 0;
+    // if header is sticky
+    if (
+      document.querySelector('header.is-fixed') !== null ||
```
mikemai2awesome, line 35: Question for @danielamorse, are you using `is-*` classes? I thought you said you stopped using them.

packages/components/bolt-tooltip/src/tooltip.js
```diff
@@ -13,12 +13,30 @@ class BoltTooltip extends BoltElement {
...
+
+    this.zIndex = parseInt(
+      getComputedStyle(this).getPropertyValue('--c-bolt-tooltip-z-index'),
+      10,
+    );
+
+    let offsetHeight = 0;
+    // if header is sticky
+    if (
+      document.querySelector('header.is-fixed') !== null ||
```
mikemai2awesome, line 33: Same comment about `is-*` classes.

mikemai2awesome (approved):
Approved with a question for @danielamorse, see inline comments.

packages/components/bolt-tooltip/src/tooltip.js
danielamorse: @mikemai2awesome `header.is-fixed` is for the old page header. It's the only way we can target it.

We need a universal "sticky" selector, and we need the ability to differentiate between the sticky page header and other sticky elements. How about:
- `.js-bolt-sticky.c-bolt-page-header` // sticky header
- `.js-bolt-sticky:not('c-bolt-page-header')` // everything else

Alternatively, we do:
- `.js-bolt-sticky-page-header` // sticky header
- `.js-bolt-sticky` // everything else

I think I prefer the latter.

On the navbar branch I'm using `[data-bolt-sticky-header]`, but I think a `js-*` class more consistent with our conventions because we're using is as a JS selector. Plus, I see "sticky" as a property not a state, and `data-*` is for state, like `[data-bolt-ready]`.

danielamorse:
@cjwhitedev, todo:
- In CSS use the modal z-index value instead, comment that this needs to be refactored
- See Popover JS and apply my changes to Tooltip

packages/components/bolt-tooltip/src/tooltip.js
```diff
@@ -21,7 +23,38 @@ class BoltTooltip extends BoltElement {
     this.uuid = this.uuid || Math.floor(10000 + Math.random() * 90000);
   }
 
+  getPaddingTop() {
```
danielamorse, line 26: @cjwhitedev in retrospect this is be a good function to move to core and let both components share. Not necessary now, but next time we have to touch this code I'd say it's a good idea.

danielamorse (approved):
Looks good, ship it!

PR #2216, opened 2021-05-28
DS-481: Fix popover/tooltip when shadow dom disabled

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-481

## Summary

Fix popover and tooltip when rendered to light dom, for example when inside a `<form>`.

## Details

Popover and tooltip were breaking when inside a `<form>` where shadow dom is disabled (for legacy reasons). It was breaking because the "content" slot was not actually being rendered in the light dom version of these components. It is automatically rendered when using the shadow dom version, but not rendered at all when shadow dom is disabled. To fix this, I'm now explicitly rendering the "content" slot and hiding it via CSS. This now works when shadow dom is on and off.

This wasn't discovered earlier because rendering a popover or tooltip to the light dom is an edge case, but this is a good reminder to include light dom testing when we're working with our web components.

## How to test

- Review code
- See tooltip/popover docs, verify continue to work as expected
- See two new test pages where components are rendered to light dom, verify they work as expected:
  - http://localhost:3000/pattern-lab/?p=tests-popover-no-shadow
  - http://localhost:3000/pattern-lab/?p=tests-tooltip-no-shadow

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/999-tests/popover/05-popover-no-shadow.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/999-tests/popover/05-popover-no-shadow.twig
@@ -1,0 +1,34 @@
+{% set link %}
+  {% include '@bolt-components-link/link.twig' with {
+    text: 'call to action',
+    url: 'https://pega.com',
+  } only %}
+{% endset %}
+
+{% set trigger %}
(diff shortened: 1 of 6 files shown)


adamszalapski (approved):
Working as expected. 👍

cjwhitedev (approved):

PR #2229, opened 2021-06-11
DS-467 remove card deprecations

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-467

## Summary

Removes deprecated borderRadius prop from card component

## Details

borderRadius has been replaced with border_radius and will be removed in 4.0

## Release notes

The `borderRadius` prop in the card component has been renamed to `border_radius`.  Usage of the `borderRadius` prop will no longer work starting in Bolt 4.0.


cjwhitedev (approved):
looks right to me.

PR #2230, opened 2021-06-14
DS-457: Table Width Bug Fix

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-457](https://pegadigitalit.atlassian.net/browse/DS-457)

## Summary

Address bug for tables that aren't as wide as their container.

## Details

Change default styles of table to not include a background (because there was an improperly styled flash when undoing the inverse). Compare table body width to table width and apply background and styles if the former is equal to or wider than the latter.

## How to test

Pull the branch and check out the temporary demo page I made as well as the real demo pages.

### DIFF
--- a/packages/components/bolt-table/src/table.scss
+++ b/packages/components/bolt-table/src/table.scss
@@ -21,8 +21,6 @@ $bolt-table-row-hover-background-color: rgba(bolt-color(gray), 0.1);
 
 .c-bolt-table {
   @include bolt-horizontal-scroll;
-  @include bolt-table-border(right);
-  @include bolt-table-border(bottom);
   @include bolt-table-border(left);
 
   display: inline-block;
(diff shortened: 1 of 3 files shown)


packages/components/bolt-table/src/table.scss
```diff
@@ -31,27 +29,41 @@ $bolt-table-row-hover-background-color: rgba(bolt-color(gray), 0.1);
...
-  background-attachment: scroll, scroll;
-  background-color: var(--m-bolt-bg);
-  background-repeat: no-repeat;
-  background-size: 0.5em 100%, 0.5em 100%;
   overflow-x: auto;
   border-collapse: collapse;
   border-spacing: 0;
   -ms-overflow-style: -ms-autohiding-scrollbar;
+
+  &--background {
```
mikemai2awesome, line 37: This modifier should be named `--responsive` since that's what it's doing.

packages/components/bolt-table/src/table.scss
```diff
@@ -66,6 +78,10 @@ $bolt-table-row-hover-background-color: rgba(bolt-color(gray), 0.1);
   );
 }
 
+.c-bolt-table__body {
+  @include bolt-table-border(bottom);
```
mikemai2awesome, line 82: This is what i meant by moving around the border, going from table to table body and cells. This breaks existing tables, especially the web components (and this is the majority use case). See page: https://boltdesignsystem-2kddt3vvb-boltdesignsystem.vercel.app/pattern-lab/?p=components-table-with-web-component

mikemai2awesome (changes requested):

danielamorse:
Let's address this when we rebuild the Table component: https://pegadigitalit.atlassian.net/browse/DS-457?focusedCommentId=32787

PR #2220, opened 2021-06-04
DS-432 link deprecations

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-432

## Summary

Removes deprecated props from link component

## Details

- Removes four previously deprecated props -- `href`, `isHeadline`, `onClick`, and `onClickTarget` -- from the link component
- Removes all remaining usages of these props in patter lab
- Fixes a bug in microjourneys editing caused by usage of the deprecated `isHeadline` prop.

## How to test

- Confirm that none of the deprecated props -- `href`, `isHeadline`, `onClick`, and `onClickTarget` -- work in the link component.

- Go to /pattern-lab/?p=experiments-micro-journeys
- Confirm that there are no errors when clicking the `edit` button (currently, there are errors on boltdesignsystem.com)

## Release notes

- Four previously deprecated props -- `href`, `isHeadline`, `onClick`, and `onClickTarget` -- have been removed from the link component.

PR #2231, opened 2021-06-14
Revert "DS-432: Remove deprecated props from link component"

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-432

## Summary

Adds back in support for deprecated link props removed in #2220

## Details

As decided in dev huddle, we want to deprecate the link component altogether in favor of the link element.   It's not that spending time on removing deprecated props is wrong, but that the time is better spent elsewhere.

## How to test
- Checkout the branch locally
- Set `schemaErrorReporting` to `none` in [pattern lab's boltrc](https://github.com/boltdesignsystem/bolt/blob/master/docs-site/.boltrc.js#L24) to allow it to build when deprecated props are passed.
- Try adding any of the four deprecated props that this PR adds back and confirm they work

(testing probably isn't required since this just reverts previously existing code)

### DIFF
--- a/packages/components/bolt-link/link.schema.js
+++ b/packages/components/bolt-link/link.schema.js
@@ -27,6 +27,22 @@ module.exports = {
   title: 'Link',
   description: 'Text link.',
   type: 'object',
+  not: {
+    anyOf: [
+      {
+        required: ['href'],
+      },
+      {
+        required: ['isHeadline'],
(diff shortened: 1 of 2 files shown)


danielamorse (approved):
Works as expected. Neat - I didn't know about `schemaErrorReporting`.

PR #2232, opened 2021-06-16
DS-335 Set up site layout

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-335

## Summary

Introduces a Layout called Site. The classic pancake stack site layout that enables the footer to be always positioned on the bottom.

## Details

1. Site added under Layouts.
2. This replaces `.c-bolt-site`.
3. Once this is implemented into the base theme and all sites convert to use the base theme template without overrides, we can remove everything related to `.c-bolt-site`.

## How to test

Run the branch locally and check all the docs under Layouts > Site. Make sure all props and blocks work as expected.

### DIFF
--- a/packages/layouts/bolt-site/src/site.twig
+++ b/packages/layouts/bolt-site/src/site.twig
@@ -1,0 +1,57 @@
+{# Schema Validation #}
+{% set schema = bolt.data.components['@bolt-layouts-site'].schema %}
+{% if enable_json_schema_validation %}
+  {{ validate_data_schema(schema, _self)|raw }}
+{% endif %}
+
+{# Variables #}
+{% set this = init(schema) %}
+{% set attributes = create_attribute(attributes|default({})) %}
+
+{# Array of classes based on the defined + default props #}
+{% set classes = [
+  'l-bolt-site',
+] %}
+
+{# Template #}
+<div {{ attributes.addClass(classes) }}>
+  {% if header %}
+    {{ header }}
+  {% else %}
+    {% block site_header %}
+      {#
+      /**
+      *  This must be a single element, for example:
+      *  <header>
+      *    ...
+      *  </header>
+      */
+      #}
+    {% endblock %}
+  {% endif %}
+
+  <main id="main-content" class="c-bolt-site__content">
+    {% if content %}
+      {{ content }}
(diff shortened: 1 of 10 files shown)


andrewmriv (approved):

packages/layouts/bolt-site/src/site.twig
```diff
@@ -0,0 +1,57 @@
...
+      {#
+      /**
+      *  This must be a single element, for example:
+      *  <header>
+      *    ...
+      *  </header>
+      */
+      #}
+    {% endblock %}
+  {% endif %}
```
remydenton, line 31: A simpler way to do this is:

```
  {% block site_header %}
    {#
      /**
      *  The contents of this block must be a single element, for example:
      *  <header>
      *    ...
      *  </header>
      */
      #}
    {{ header }}
  {% endblock %}
```

That way, if there's a `header` variable, it just gets printed and you're done.  But if this is being extended and there's a `site_header` block, then whatever is in this block overrides this and `{{ header }}` is ignored.

That said, I take back what I said earlier on the call; I don't think this should be a block at all (not least to keep the "no twig blocks" rule, I didn't realize we were going to put this in Bolt).  There's no reason to do this:

```
{% embed '@bolt-layouts-site/site.twig' %}
  {% block site_header %}
    This is the header.
  {% endblock %}
  {% block site_content %}
    This is the main content.
  {% endblock %}
  {% block site_footer %}
    This is the footer.
  {% endblock %}
{% endembed %}
```

When you can just do this:
```
{% set header %}
  This is the header.
{% endset %}
{% set content %}
  This is the main content.
{% endset %}
{% set footer %}
  This is the footer.
{% endset %}
{% include '@bolt-layouts-site/site.twig' with {
  header: header,
  content: content,
  footer: footer,
} only %}
```

So, this whole line can just be:
```suggestion
  {#
    /**
    *  The contents of header must be a single element, for example:
    *  <header>
    *    ...
    *  </header>
    */
    #}
  {{ header }}
```

mikemai2awesome:
@remydenton I'm merging this since we just need the class for Andrew to start working, if there are other twig conventions we wanna massage, we can do it separately.

PR #2233, opened 2021-06-16
DS-499 Update Navbar offset props

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-499

## Summary

Update Navbar offset props to include: `sticky_offset`, `sticky_offset_selector`, and `scroll_offset`.

## Details

We're renaming and reworking some of the offset props for clarity, and so the naming convention will scale. This PR differentiates "sticky" offset which relates to the position of the navbar itself and "scroll" offset which relates to the scroll position. Internally, the JS takes both into account when calculating sticky and scroll offsets.

> Note: we opted to leave out `scroll_selector_offset` as there is no real use case for this prop yet. If one comes up we can deal with it then.

## How to test
- Review the code changes to the schema, Twig, and JS
- Review the offset docs pages:
  - [Sticky offset](https://feature-ds-499-navbar-offset-props.boltdesignsystem.com/pattern-lab/?p=components-navbar-sticky-offset)
  - [Scroll offset](https://feature-ds-499-navbar-offset-props.boltdesignsystem.com/pattern-lab/?p=components-navbar-scroll-offset)


mikemai2awesome (approved):
Updated the docs. All look good.

PR #2245, opened 2021-06-25
DS-526: Replace Button in Carousel Component

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-526](https://pegadigitalit.atlassian.net/browse/DS-526)

## Summary

Replace component button with element button in carousel component.

## Details

we're deprecating and removingt the button component. this replaces instance of the button component in carousel with the button element.

## How to test

look at the demos in pattern lab as well as the code and the updated jest snap.

### Visual changes

The button element is slightly larger than the button component for accessibility purposes, this is expected.

### DIFF
--- a/packages/components/bolt-carousel/__tests__/carousel.js
+++ b/packages/components/bolt-carousel/__tests__/carousel.js
@@ -49,8 +49,12 @@ const carouselSlideImage = `
 `;
 
 const carouselButtonControls = `
-  <bolt-button slot="previous-btn" color="secondary" border-radius="full" icon-only>Previous <bolt-icon slot="before" name="chevron-left"></bolt-icon></bolt-button>
-  <bolt-button slot="next-btn" color="secondary" border-radius="full" icon-only>Next <bolt-icon slot="after" name="chevron-right"></bolt-icon></bolt-button>
+<button class="e-bolt-button e-bolt-button--border-radius-full e-bolt-button--secondary e-bolt-button--icon-only"  aria-label="Previous" slot="previous-btn">
+<span class="e-bolt-button__icon-center"><bolt-icon name="chevron-left"size="large"></bolt-icon></span>
+</button>
+<button class="e-bolt-button e-bolt-button--border-radius-full e-bolt-button--secondary e-bolt-button--icon-only"  aria-label="Next" slot="next-btn">
+<span class="e-bolt-button__icon-center"><bolt-icon name="chevron-right"size="large"></bolt-icon></span>
+</button>
 `;
 
 describe('carousel', () => {
--- a/packages/components/bolt-carousel/__tests__/carousel.js
+++ b/packages/components/bolt-carousel/__tests__/carousel.js
@@ -49,12 +49,12 @@ const carouselSlideImage = `
 `;
 
 const carouselButtonControls = `
-<button class="e-bolt-button e-bolt-button--border-radius-full e-bolt-button--secondary e-bolt-button--icon-only"  aria-label="Previous" slot="previous-btn">
+<span class="e-bolt-button e-bolt-button--border-radius-full e-bolt-button--secondary e-bolt-button--icon-only" aria-hidden="true" slot="previous-btn">
 <span class="e-bolt-button__icon-center"><bolt-icon name="chevron-left"size="large"></bolt-icon></span>
-</button>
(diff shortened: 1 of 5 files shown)


mikemai2awesome (changes requested):
As discussed in Slack, remove the button inside another button. Use a span with the button classes can be a quick fix:

```
<div class="c-bolt-carousel__button c-bolt-carousel__button--next c-bolt-carousel__button--inside" tabindex="0" role="button" aria-label="Next slide">
  <span class="e-bolt-button e-bolt-button--border-radius-full e-bolt-button--secondary e-bolt-button--icon-only" slot="next-btn">...</span>
</div>
```

packages/components/bolt-carousel/__tests__/carousel.js
```diff
@@ -49,8 +49,12 @@ const carouselSlideImage = `
 `;
 
 const carouselButtonControls = `
-  <bolt-button slot="previous-btn" color="secondary" border-radius="full" icon-only>Previous <bolt-icon slot="before" name="chevron-left"></bolt-icon></bolt-button>
-  <bolt-button slot="next-btn" color="secondary" border-radius="full" icon-only>Next <bolt-icon slot="after" name="chevron-right"></bolt-icon></bolt-button>
+<button class="e-bolt-button e-bolt-button--border-radius-full e-bolt-button--secondary e-bolt-button--icon-only"  aria-label="Previous" slot="previous-btn">
```
mikemai2awesome, line 52: Should this be the same as the others and use `span`? If so, remove `aria-label`. Add `aria-hidden="true"` instead since the label is already rendered on the `role="button"` element.

mikemai2awesome (changes requested):

packages/components/bolt-carousel/src/carousel.js
```diff
@@ -503,16 +503,20 @@ class BoltCarousel extends BoltElement {
...
 
-    const buttonTemplate = (text, iconName) => html`
-      <bolt-button
-        size="medium"
-        border-radius="full"
-        color="secondary"
-        icon-only
+    const buttonTemplate = iconName => html`
+      <span
+        type="button"
```
mikemai2awesome, line 508: Missed one here, remove `type`.

mikemai2awesome (changes requested):

packages/components/bolt-carousel/src/carousel.js
```diff
@@ -504,15 +504,19 @@ class BoltCarousel extends BoltElement {
...
-        color="secondary"
-        icon-only
+      <span
+        type="button"
+        class="e-bolt-button
+        e-bolt-button--medium
+        e-bolt-button--border-radius-full
+        e-bolt-button--secondary
+        e-bolt-button--icon-only"
+        aria-label="${text}"
```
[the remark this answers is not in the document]
mikemai2awesome: Still need to remove `type`.

mikemai2awesome (approved):

packages/components/bolt-carousel/src/carousel.js
```diff
@@ -503,16 +503,19 @@ class BoltCarousel extends BoltElement {
...
+    const buttonTemplate = iconName => html`
+      <span
+        class="e-bolt-button
+        e-bolt-button--medium
+        e-bolt-button--border-radius-full
+        e-bolt-button--secondary
+        e-bolt-button--icon-only"
+        aria-hidden="true"
       >
-        ${text}
```
danielamorse, line 513: Are we intentionally removing the text here? Isn't "Next/Previous" good for accessibility?

danielamorse (changes requested):
Looks great, two requests:
- See comment about removing icon text
- Add Release Notes to this PR describing visual changes to button for QA

packages/components/bolt-carousel/src/carousel.js
mikemai2awesome: We are only using this span for styles, and we have `aria-hidden="true"` on the span. The label is on the element with `role="button" already.

danielamorse (approved):
Nice, thanks for explaining and for updating the PR description.

I updated the heading structure around "Visual changes" in the PR description. I think it needs to be like this for the auto-release script to compile the release notes:
```
## Release notes
### Visual changes
...
```
Here's a good example: https://github.com/boltdesignsystem/bolt/pull/2207

:shipit:

PR #2250, opened 2021-06-30
DS-535: Replace Button in Typeahead Component

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-535](https://pegadigitalit.atlassian.net/browse/DS-535)

## Summary

Replace component button with element button in typeahead component.

## Details

we're deprecating and removing the button component. this replaces instance of the button component in typeahead with the button element.

## How to test

look at the demos in pattern lab as well as the code and the updated jest snap.

### Visual changes

The button element is slightly larger than the button component for accessibility purposes, this is expected.


packages/components/bolt-typeahead/typeahead.js
```diff
@@ -99,51 +99,51 @@ class BoltTypeahead extends withLitEvents {
...
           {
             [`is-visible`]: this.inputValue !== '',
           },
         )}
+        aria-label="${this.clearInputText}"
+        type="reset"
+        @click=${this.clearSearch}
+        disabled=${ifDefined(this.disabled === true ? true : undefined)}
       >
-        ${this.clearInputText}
```
danielamorse, line 139: @mikemai2awesome I see we're replacing text with aria-labels. Is this the way forward for icon-only buttons?

danielamorse (changes requested):
Nice work on the JS. Looks like a pretty clean replacement.

Take a look at focus styles though. When you focus on the submit or close buttons they pop out of place.

https://user-images.githubusercontent.com/36336/124037832-2d36f980-d9ce-11eb-82bd-09f671dbbd35.mov

mikemai2awesome (changes requested):
Same comment as @danielamorse

packages/components/bolt-typeahead/typeahead.js
mikemai2awesome: yes, it's redundant to have both. either use aria-label or use text. in icon only buttons, use aria-label.

cjwhitedev:
@danielamorse i had seen this before and thought i fixed it. i don't see it on my local in firefox, safari, or chrome. what browser are you using?

packages/components/bolt-typeahead/typeahead.js
```diff
@@ -99,51 +99,51 @@ class BoltTypeahead extends withLitEvents {
...
-        type="reset"
-        @click=${this.clearSearch}
-        no-shadow
+        aria-label="${this.submitButtonText}"
+        type="submit"
+        @click=${this.submit}
         disabled=${ifDefined(this.disabled === true ? true : undefined)}
+      >
+        <span class="e-bolt-button__icon-center">
+          <bolt-icon
```
mikemai2awesome, line 116: oh, since we are writing web component markup here, you'll need to manually add `aria-hidden="true"`. that was handled in twig but doesn't happen automatically in web component.

packages/components/bolt-typeahead/typeahead.js
```diff
@@ -99,51 +99,51 @@ class BoltTypeahead extends withLitEvents {
...
-        ${this.clearInputText}
-        <bolt-icon
-          name="close-solid"
-          slot="before"
-          class=${cx('c-bolt-typeahead__icon')}
-          title=${this.clearInputText}
-        />
-      </bolt-button>
+        <span class="e-bolt-button__icon-center">
+          <bolt-icon
```
mikemai2awesome, line 140: Same here, `aria-hidden` needed. And please remove `title`. That attribute should never be used. We already have `aria-label` on the `<button>`.

mikemai2awesome (changes requested):
@cjwhitedev couple more fixes then it's all set.

packages/components/bolt-typeahead/typeahead.js
mikemai2awesome: @cjwhitedev Remove the `title` please.

mikemai2awesome (approved):

PR #2251, opened 2021-07-01
DS-381: Remove click-handler JS module

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-381

## Summary

Remove unofficial `data-on-click` behavior intended for use on Button/Link elements.

## Details

I added this JS module when I thought we would support the same `on-click` and `on-click-target` behavior on elements that we used on components. Since then, we've decided to not add support for this type of JS helper on elements. Instead, just use the native `onclick` attribute or write your own custom JS.

This JS was never officially released or documented so it is safe to remove. It. does not affect existing `on-click` behavior on components.

## How to test

- See files changed


adamszalapski (approved):
I'm ok with that. 👍

colbytcook (approved):

PR #2256, opened 2021-07-02
DS-534: Replace Button in Card Component

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-534](https://pegadigitalit.atlassian.net/browse/DS-534)

## Summary

Replace component button with element button in card component.

## Details

We're deprecating and removing the button component. this replaces instance of the button component in card with the button element.

## How to test

check out the demos and the code for any possible breaking changes.

### Visual changes

**Card component**
- Card actions (buttons/links in card footer) have slightly larger font-size for accessibility purposes.
- Card actions on hover now have underline instead of just changing color.


mikemai2awesome (changes requested):
I feel like we should do things the right way once and for all:
```
<bolt-card-replacement-action role="link" href="https://yahoo.com" target="_blank">
  ...
</bolt-card-replacement-action>
<bolt-card-replacement-action role="button" type="submit">
  ...
</bolt-card-replacement-action>
```
There are so many existing problems with the button/link inside the shadow dom, this will get it back to proper HTML.

Additionally, there is a lot that goes into removing white space in links and buttons, which you need to do the same here. Otherwise you will end up with underline like this:

Screen Shot 2021-07-02 at 11 41 18 AM

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig
```diff
@@ -8,35 +8,47 @@
...
-        {# @todo we should not be passing the `item` attributes into the button template. Try to fix with next breaking change. #}
         {% set _button_props = {
-          text: this.data.text.value,
-          hierarchy: 'text',
-          width: 'full',
-          align: 'start',
+          content: this.data.text.value,
+          hierarchy: 'transparent',
+          display: 'block',
           target: this.data.external.value == true ? '_blank' : '_self',
```
cjwhitedev, line 15: future @cjwhitedev this is incorrect

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
```diff
@@ -35,36 +41,42 @@ class BoltCardReplacementAction extends withContext(BoltElement) {
...
         this.slotMap.get('default')[0].tagName === undefined
           ? html`
-              <bolt-button
-                color="text"
-                width="full"
-                align="start"
-                url="${ifDefined(this.url ? this.url : undefined)}"
+              <a
+                class="${buttonClasses}"
+                href="${ifDefined(this.url ? this.url : undefined)}"
```
danielamorse, line 61: @mikemai2awesome if `this.url` is undefined should this be a semantic `button[type="button"]`?

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
```diff
@@ -35,36 +41,42 @@ class BoltCardReplacementAction extends withContext(BoltElement) {
...
                   this.url ? (this.external ? '_blank' : '_self') : undefined,
                 )}"
-                size="${ifDefined(this.spacing ? this.spacing : undefined)}"
               >
                 ${this.slotify('default')}
                 ${this.icon !== 'none'
-                  ? html`
-                      <bolt-icon
-                        slot="after"
+                  ? // prettier-ignore
```
danielamorse, line 68: What whitespace problem does this solve? When I remove it I don't notice a difference on the front-end. Can you show me an example?

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig
```diff
@@ -8,35 +8,47 @@
...
     {% block action_item %}
       {% if this.data.text.value %}
-        {# @todo we should not be passing the `item` attributes into the button template. Try to fix with next breaking change. #}
         {% set _button_props = {
-          text: this.data.text.value,
-          hierarchy: 'text',
-          width: 'full',
-          align: 'start',
-          target: this.data.external.value == true ? '_blank' : '_self',
+          content: this.data.text.value,
```
danielamorse, line 12: @cjwhitedev what do you think of this refactor? This way you build up the button attributes and only add it once to the button props.
```
        {% set _button_attributes = attributes|default({}) %}
        {% set _button_attributes = _button_attributes|merge({ target: this.data.external.value == true ? '_blank' : '_self' }) %}

        {% if this.data.url.value %}
          {% set _button_attributes = _button_attributes|merge({ href: this.data.url.value }) %}
        {% else %}
          {% set _button_attributes = _button_attributes|merge({ type: 'button' }) %}
        {% endif %}

        {% set _button_props = {
          content: this.data.text.value,
          hierarchy: 'transparent',
          display: 'block',
          attributes: _button_attributes
        } %}

        {% if spacing is not empty %}
          {% set _button_props = _button_props|merge({ size: spacing }) %}
        {% endif %}

        {% if this.data.icon.value != 'none' %}
        ...
```

danielamorse (changes requested):
@cjwhitedev nice work, almost there. This one was super tricky!

I made some inline comments. Take a look and lmk if you have any questions.

Thanks for adding the note for visual changes note for QA. Reminder that it has to be under the `## Release Notes` heading for it to be automatically compiled when we release. I added it 👍 

Also, if we no longer have a dependency on Bolt Button it can be removed from [Card's package.json file](https://github.com/boltdesignsystem/bolt/blob/feature/DS-534-card-button-replacement/packages/components/bolt-card-replacement/package.json#L22).

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
mikemai2awesome: yes. but that locks it down to just `type="button"` without a way to do `reset` and `submit`.

mikemai2awesome:
@cjwhitedev @danielamorse There's still this white space issue on hover, see the underline is longer than it should be. It's not that jarring so if it's hard to fix then we can leave it.
Screen Shot 2021-07-13 at 12 50 02 PM

danielamorse:
> @cjwhitedev @danielamorse There's still this white space issue on hover

Oh yeah, meant to mention this - it's the whitespace in the light dom. So if you remove whitespace around the text it fixes it. Like so:
```
<bolt-card-replacement-action>Button text</bolt-card-replacement-action>
```

This is different than other whitespace issues we've fixed with JS in the past. I recommend letting the user fix it by simply not including line breaks.

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.scss
```diff
@@ -5,6 +5,8 @@
 @import '@bolt/core-v3.x';
 @import '../card-replacement-settings-and-tools.scss';
 
+// @TODO: can we unify these? why are the styles different?
```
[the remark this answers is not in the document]
cjwhitedev: Yes - I guess I was initially confused why they were two different styles. It seems like one of them only applies to the WC but the other applies to both the twig and the WC. Why style them differently? But now I'm wondering if I just misunderstood the nesting.

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
cjwhitedev: @mikemai2awesome @danielamorse so do we just leave this alone or is there work to be done?

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
cjwhitedev: This resolves the underline @mikemai2awesome called out in the web component uses.

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig
cjwhitedev: I think its a bit long winded, but neither @colbytcook or I could figure out another way to merge attributes without overwriting them. If you have a shorter way then I'd love to use it, but as far as I'm concerned, this works, and it's fine.

packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.scss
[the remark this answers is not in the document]
danielamorse: `bolt-card-replacement-action` is the WC element. `.c-bolt-card_replacement__action` is the inner div in the shadow dom. `.e-bolt-button` is repeated in both because it can appear either in the shadow or the light dom. Does that clear things up @cjwhitedev?

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
danielamorse: I vote to leave this alone. It only affects the web component. If you use Twig you get the proper semantic element.

danielamorse (approved):
Approved 🎉 

@cjwhitedev @mikemai2awesome,
- Added a demo of card opening a modal (`/pattern-lab/?p=tests-card-links-and-buttons`)
- Fixed the focus style
- Fixed the Twig logic so that semantic buttons don't get `target` attribute

What's up with the Listing Teaser snapshot in this PR? Is that something that's on `master` we can pull in rather than commit here?

mikemai2awesome (approved):

PR #2257, opened 2021-07-02
DS-204: Print Styles

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-204](https://pegadigitalit.atlassian.net/browse/DS-204)

## Summary

Add a print style utility for academy to a demo. 

## Details

- `e-bolt-printable` class on the `<body>` of a page will hide everything in print mode.
- `e-bolt-printable__article` class on the article container will display the container in print mode.
- Added `all` to the style lint config. It has to come first, otherwise it will override everything if you do `all: initial` after all the rules you’ve written

## How to test

Open the [demo page](https://feature-ds-204-print-styles.boltdesignsystem.com/pattern-lab/patterns/50-pages-60-academy-05-pages-t2-inner-pages-topic-details-acd-topic-detail-page/50-pages-60-academy-05-pages-t2-inner-pages-topic-details-acd-topic-detail-page.html) we added the element classes to and use the browser's print utility to see a preview of the print styles. Either print or save as PDF, only the article would display in print mode.


danielamorse (approved):

mikemai2awesome (approved):

mikemai2awesome (commented):
@cjwhitedev @danielamorse As a follow up, can we move the CSS to the proper element folder?

colbytcook (approved):

mikemai2awesome (changes requested):
@cjwhitedev  I just realized the CSS is way too generic, it will apply to every page, that means you'll be printing blanks everywhere.

I fixed the CSS and docs to make sure the usage is more specific and you can target certain pages to use the print styles, but I don't know what to do with the twig and schema. The usage involves adding a class to the `<body>` and then another class to the targeted container. Twig doesn't make sense at all.

mikemai2awesome (approved):
I removed the twig and schema, then made sure the doc still loads. Keeping it as element right now.

PR #2259, opened 2021-07-07
DS-539 Fix toolbar buttons alignment

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-539

## Summary

The Toolbar component ws updated, so that when there is a long breadcrumb the items on the right (download, share) do not wrap and align at the top of the container.

## Details

Above “medium” breakpoint Toolbar items do not wrap and are aligned to the top of the toolbar.

https://user-images.githubusercontent.com/85670248/124731768-d3c15400-df12-11eb-8ef6-b757978b73b8.mov

## How to test

Check if items on the right (download, share) behave properly above and below `medium` breakpoint.

### DIFF
--- a/packages/components/bolt-toolbar/toolbar.scss
+++ b/packages/components/bolt-toolbar/toolbar.scss
@@ -43,6 +43,7 @@
 
   @include bolt-mq(xsmall) {
     grid-template-columns: auto 1fr auto;
+    align-items: center;
 
     > * + * {
       margin-top: 0;
--- a/packages/components/bolt-toolbar/toolbar.scss
+++ b/packages/components/bolt-toolbar/toolbar.scss
@@ -59,6 +59,11 @@
   position: relative;
   width: 100%;
   margin-bottom: calc(var(--bolt-spacing-y-small) * -1);
+
+  @include bolt-mq(medium) {
+    flex-wrap: nowrap;
+    align-items: flex-start;
+  }
 }
 
 .c-bolt-toolbar__item {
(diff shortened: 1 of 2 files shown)


mikemai2awesome (changes requested):
This does not work, it is wrapping too early. Set it to wrap at `small` breakpoint.

Secondly, the ones with shorter titles are broken now. The title is not center aligned vertically:

Screen Shot 2021-07-07 at 11 15 47 AM

MarcinMr:
@mikemai2awesome 

I was suggested `medium` break point from the ticket description.

There is a case with rocket icon alignment:

1.`align-items: center`

1.a. Above `small` bp, neutral text length. Looks good

above-small-center-v-alignment-normal-text


1.b. Above `small` bp, longer text length. Looks good

above-small-centered-v-long-text


1.c Below `small` bp, longer text length. Rocket is vertically centered to the breadcrumb text

below-small-centered-v-long text

2.`align-items: flex-start`

2a. Below `small` bp, long text length. Looks good

below-small-top-v-longer-text


2b. Below `small` bp, short text length. Looks bad and it will also break above `small` bp if text is short like in your screenshot.

below-small-top-v-short-text


**I think we should stay with `align-items: center`**

mikemai2awesome:
@MarcinMr Actually, you can get it to behave as expected with css grid. Reference the changes I made to understand the method.

toolbar

mikemai2awesome (approved):

PR #2260, opened 2021-07-07
DS-548 Listing teaser custom status numbers

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-548

## Summary

Adds a `status.number_items` prop to render custom status numbers in Listing Teaser.

## Details

1. Twig and schema added for status number.
2. New prop added to render custom status numbers instead of replies and views.
3. Replaced all instances of `status.replies` and `status.views` with `status.number_items`.
4. Added deprecation text for `status.replies` and `status.views` in schema.

## How to test

Run the branch locally and check the custom status number doc under Listing Teaser. Make sure the new prop works as expected.


colbytcook (approved):

docs-site/src/pages/pattern-lab/_patterns/40-components/listing-teaser/20-listing-teaser-status-and-action.twig
```diff
@@ -90,6 +88,16 @@
       },
     } only %}
   {% endset %}
+  {% set number_items %}
```
cjwhitedev, line 91: if we're changing the demo here, is the visible markup still correct as is?

cjwhitedev (commented):
Everything else looks right to me, assuming that you can get the Jest Tests to pass. Looks like they're failing in Travis right now.

docs-site/src/pages/pattern-lab/_patterns/40-components/listing-teaser/20-listing-teaser-status-and-action.twig
mikemai2awesome: docs updated

cjwhitedev (approved):

PR #2261, opened 2021-07-08
DS-489 twig deprecations

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-489

## Summary

Replaces deprecated twig syntax and usages

## Details

A subset of commits from https://github.com/boltdesignsystem/bolt/pull/2258, this PR is all the work that should be 100% backward compatible and can be merged while still using twig 1.

## How to test

Confirm no regressions in pattern lab


colbytcook:
@remydenton this all looks good, I didn't review the PHP files because I really can't tell if they are correct (maybe @andrewmriv can peek at them?). But I reviewed all the twig file updates and is seems like there was four types of changes, let me know if I got these correct.

## Twig 1 to Twig 2 changes
* `sameas` -> `same as`
* `extends [PATH]` -> `import [PATH] as [VARIABLE]` 
* checking "true" requires ` is defined`
* accessing array (objects?) values, `this.props.[KEY]` -> `this.props['[KEY]']`

packages/twig-integration/twig-extensions-shared/src/Layout/GridCellNode.php
```diff
@@ -96,25 +96,12 @@ static public function cellFunctionToCall(){
...
-    ";
-
-    // Pre-render the inline Twig template + the data we've merged and normalized
-    $rendered = $stringLoader->render(array("string" => $string, "data" => $data));
+    $env = new \Twig_Environment(new \Twig_Loader_Array([]), [
+      'debug' => true,
+      'autoescape' => false,
+    ]);
+    $template = $env->createTemplate('<div {{ attributes | raw }}> {{ contents }} </div>');
+    $rendered = $env->render($template, array('contents' => $contents, 'attributes' => $attributes ));
```
andrewmriv, line 104: Smallest nitpick (though to be fair this is more of a Drupal coding standard):
- Consider replacing `array()` with `[]`

andrewmriv (approved):

packages/twig-integration/twig-extensions-shared/src/Layout/GridCellNode.php
remydenton: Good call, I think I was referencing some examples that had the old syntax.   I will update it!

remydenton: OK, updated!

PR #2270, opened 2021-07-15
DS-563: Fix boolean attribute syntax

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-533 (bug discovered during implementation)

## Summary

Fix boolean attribute syntax in navbar

## Details

When trying to set the `current` attribute to `true` in Drupal, I got an error that `setAttribute()` was missing an argument (only had one instead of two).  This PR fixes the syntax by passing `TRUE` as the second argument to boolean attributes.

The Drupal docs don't explicitly address boolean attributes, but this seems correct to me based on the other docs at https://www.drupal.org/docs/8/theming-drupal-8/using-attributes-in-templates#s-attributessetattributeattribute-value

## How to test

This may only be a problem in Drupal, but it seems like you should be able to test by creating a navbar item with `current` set to `true`.  If attributes behave the same in pattern lab as they do in Drupal, it will throw an error, or perhaps just not respect the prop.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/999-tests/navbar/05-navbar-with-active-link.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/999-tests/navbar/05-navbar-with-active-link.twig
@@ -1,0 +1,63 @@
+{% set navbar_data = [
+  {
+    link: {
+      content: 'Section 1',
+      attributes: {
+        href: '#section-1'
+      }
+    }
+  },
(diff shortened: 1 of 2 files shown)


danielamorse (approved):
Nice catch! I verified using the `current` prop at all breaks the component, which says that we didn't actually test it 😬 

I added a test page so now we at least have on instance of this prop in use.

PR #2271, opened 2021-07-15
DS-562 Remove unexpected margin bottom on Navbar title

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-562

## Summary

Removes unexpected margin-bottom when you use a heading tag (`h1`, `h2`, etc.) with Navbar title.

## Details

If Navbar title uses a heading tag, it inherits global heading styles. This PR removes the margin-bottom that global styles add.

## How to test

- Review changed files
- View [test page](http://localhost:3001/pattern-lab/?p=tests-navbar-title-tags) to verify there is no margin-bottom on Navbar titles.

### DIFF
--- a/packages/components/bolt-navbar/src/navbar.scss
+++ b/packages/components/bolt-navbar/src/navbar.scss
@@ -119,9 +119,8 @@
 
 // Title
 .c-bolt-navbar__title {
+  all: unset; // 'unset' all styles when title uses heading tag
   display: flex;
-  margin-bottom: 0; // Override global heading styles when title uses heading tag
-  font-weight: initial;
   color: var(--m-bolt-headline);
   white-space: nowrap;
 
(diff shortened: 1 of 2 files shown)


packages/components/bolt-navbar/src/navbar.scss
```diff
@@ -120,6 +120,7 @@
 // Title
 .c-bolt-navbar__title {
   display: flex;
+  margin-bottom: 0; // Override global heading styles when title uses heading tag
```
mikemai2awesome, line 123: Here's a case where you can use `all` to reset everything and start anew. https://developer.mozilla.org/en-US/docs/Web/CSS/all

mikemai2awesome (changes requested):

packages/components/bolt-navbar/src/navbar.scss
danielamorse: Updated to use `all: unset`, ready for review. Note: `initial` reset the font-family, `unset` changes exactly what we want to change and nothing else.

mikemai2awesome (approved):

remydenton (approved):
Looks great, thanks for the fix @danielamorse.

PR #2273, opened 2021-07-16
DS-556 Convert $bolt-transition to CSS var

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-556

## Summary

Delete all SCSS variables `$bolt-transition` and `bolt-transition-*` and apply relevant CSS var.

## Details

All SCSS transition vars were replace with CSS vars.
Here is reference to the CSS vars: https://github.com/boltdesignsystem/bolt/blob/master/packages/global/styles/00-vars/_vars-transition.scss

## How to test

Make sure there is no any of SCSS var: `$bolt-transition`, `$bolt-transition-*` and are respectively switched.
Compare transitions on this branch with the master branch.

### Deprecations

3 scss variables are deprecated: `$bolt-transition-ease`, `$bolt-transition-timing`, `$bolt-transition`. Find and replace with `var(--bolt-transition-ease)`, `var(--bolt-transition-timing)`, `var(--bolt-transition)`.


packages/components/bolt-text/src/text.scss
```diff
@@ -186,7 +186,7 @@
 
   a {
     opacity: bolt-opacity(100);
-    transition: all $bolt-transition;
+    transition: all vr(--bolt-transition);
```
mikemai2awesome, line 189: Fix the typo here.

mikemai2awesome (changes requested):

packages/components/bolt-text/src/text.scss
MarcinMr: @mikemai2awesome typo fixed

packages/core-v3.x/styles/01-settings/settings-global/_settings-global.scss
```diff
@@ -29,14 +29,6 @@ $bolt-border-color: var(--m-bolt-border) !default;
 /// Bolt border radius
 $bolt-border-radius: 3px !default;
 
-// Transition
```
mikemai2awesome, line 32: We cannot remove. We are only deprecating.

mikemai2awesome (changes requested):
Let's deprecate here as well, rather than remove. There are still many uses in the wild, removing will cause a breaking change: https://gitlab.com/search?group_id=5804310&repository_ref=master&scope=blobs&search=%24bolt-transition*&snippets=false

packages/core-v3.x/styles/01-settings/settings-global/_settings-global.scss
MarcinMr: @mikemai2awesome changes undone

mikemai2awesome (approved):

PR #2274, opened 2021-07-16
DS-408: Convert yeoman to plop alternative

adamszalapski:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-408

## Summary

Rewriting the component generator from yeoman to plop and adding the element generator.

## Details

The previous yeoman generator is removed. In his place, we added a plop generator that is easier to use and more developer-friendly. Except we rewrite the component generator we add elements generator.

## How to test

1. Pull the branch locally and run `yarn setup`.
2. Run `yarn cc` to generate component and check if all the files are generated correctly.
3. Run `yarn jest packages/generators/bolt-generator/__tests__/component/bolt-generator.js` and check if all tests passed.
4. Run `yarn ce` to generate element and check if all the files are generated correctly.
5. Run `yarn jest packages/generators/bolt-generator/__tests__/element/bolt-generator.js` and check if all tests passed.

We need only two acceptations but I want you all to check this code.


colbytcook (approved):

packages/generators/bolt-generator/helpers/add-bolt-package.js
```diff
@@ -0,0 +1,38 @@
+const fs = require('fs');
+const prettier = require('prettier');
+const resolve = require('resolve');
+const sortPackageJson = require('sort-package-json');
+
+const boltPackageJsonPath = resolve.sync('@bolt/starter-kit/package.json')
+const boltPackageJson = require(boltPackageJsonPath);
+
+const updateBoltPackage = (packageName, writePath) => {
```
danielamorse, line 9: 👏

danielamorse (commented):
@adamszalapski this is great 🎉 I thought we'd never get around to it.

I especially like:
- `plop` seems much simpler than `yo`
- Well-organized "helpers"
- Testing coverage
- User input validation, e.g. "Don't include underscores"
- Uses the latest Jest test templates

One question: if you try to run `yarn start` immediately after creating a component/element you'll see an error:
```
Error: Cannot find module '@bolt/components-foo/package.json'
```

I think this is because `yarn` doesn't know about the new package yet, even though it has been added as a dependency. Running `yarn` fixes it. Should we automatically run `yarn` on create or just leave that up to the user? Do you know of another way to register the new package?

Approved once we decide on this question^

PR #2277, opened 2021-07-20
COM-780 keyboard input styles

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/COM-780

## Summary

Fixes `kbd` styles so that it doesn't overlap other content.

## Details

1. Updated the `kbd` styles to reduce the padding.
2. Updated `code` font-size to match `kbd`, allowing code text to relatively update its size along with neighboring text.

## How to test

Run the branch locally, create a dummy page with some placeholder text, add some `kbd` and `code` in the text content. Make sure these elements don't overlap each other when they are close.


danielamorse (approved):
Works as expected, looks much better.

PR #2278, opened 2021-07-22
DS-520 Refactor Page Footer component

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-520

## Summary

Refactored Page Footer component.

## Details

1. HTML was reduced by removing bands, wrappers, and grid.
2. Layout is responsive across all breakpoints.
3. All texts and links have at least AA color contrast.
4. Proper semantics and aria attributes are added.
5. Proper interactive states (hover, active, focus, etc.) are added for all links.
6. Mobile design has been enhanced.

## How to test

Run the branch locally and check if Page Footer component conforms to accessibility standards listed in Details.


mikemai2awesome (changes requested):
@MarcinMr Can you move the doc to the Page Footer folder first? Not the Footer folder. I'll have more feedback on the CSS later.

MarcinMr:
@mikemai2awesome I moved `footer-refactor.twig` from `footer` folder to the `page-footer`. I was misled that the current `footer` being used is in the `footer` folder.

mikemai2awesome (changes requested):
@MarcinMr Nice first draft. I love you applied what you learned, it's great to see so much accessibility and semantic HTML built in.

Next, I would like to challenge you to update the markup to something similar to the below:

```html
<footer class="c-page-footer">
  <div class="c-page-footer__main">
    <div class="c-page-footer__description">
      {% include '@bolt-components-headline/headline.twig' with {
        text: 'About Pegasystems',
        tag: 'h2',
      } only %}
      <p>Pegasystems is...</p>
    </div>
    <nav class="c-page-footer__nav" aria-label="Footer Navigation">
      <h3 class="c-page-footer__nav-heading">Join the conversation</h3>
      <ul class="c-page-footer__nav-list">
        <li class="c-page-footer__nav-list-item"></li>
        ...
      </ul>
      <h3 class="c-page-footer__nav-heading">Company</h3>
      <ul class="c-page-footer__nav-list">
        <li class="c-page-footer__nav-list-item">
          {% include '@bolt-elements-text-link/text-link.twig' with {
            content: 'Link',
            icon_after: icon_chevron_right,
            reversed_underline: true,
            attributes: {
              href: '',
            }
          } only %}
        </li>
        ...
      </ul>
      <h3 class="c-page-footer__nav-heading">Pega Sites</h3>
      <ul class="c-page-footer__nav-list">
        <li class="c-page-footer__nav-list-item">{{ text link include }}</li>
        ...
      </ul>
      <h3 class="c-page-footer__nav-heading">Resources</h3>
      <ul class="c-page-footer__nav-list">
        <li class="c-page-footer__nav-list-item">{{ text link include }}</li>
        ...
      </ul>
    </nav>
  </div>
  <aside class="c-page-footer__aside t-bolt-gray-xlight">
    <nav class="c-page-footer__nav c-page-footer__nav--inline" aria-label="Utility Navigation">
      <h3 class="c-page-footer__nav-heading">Language</h3>
      <ul class="c-page-footer__nav-list">
        <li class="c-page-footer__nav-list-item">{{ text link include }}</li>
        ...
      </ul>
      <h3 class="c-page-footer__nav-heading">Legal</h3>
      <ul class="c-page-footer__nav-list">
        <li class="c-page-footer__nav-list-item">{{ text link include }}</li>
        ...
      </ul>
    </nav>
    <span class="c-page-footer__copyrights">&copy;2021 Pegasystems Inc.</span>
  </div>
</footer>
```

Secondly, I would like you to turn the links into accordions in smaller screens. You can reference [IBM](https://www.ibm.com/) and [Apple](https://www.apple.com) for such design pattern.

MarcinMr:
@mikemai2awesome I updated the markup, styled it and turned links to accordion on smaller screens.

Don't you think that listing elements one by one like in the first screenshot is a little bit waste of space ? This approach is on the reference Apple website you've sent me.

What about wrapping elements when accordion expanded like in the second screen ?

---
---

**1. Accordion expanded `width: 100%` list elements**

footer list

**2. Accordion expanded `flex-wrap` approach**

footer-flex-wrap

MarcinMr:
@mikemai2awesome 

I've done the transitions for mobile footer. To make a fade effect I need to use
`opacity` but i also use `visibility` to not breaking the accesibility as we were talking before.
For the slide effect I toggle between `max-height: 0` and `max-height: 100rem` basing on `aria-expanded` attribute.
I set a value on `max-height` to something bigger than the `c-page-footer__nav-list` will ever get.

mikemai2awesome:
@MarcinMr I added prefer-reduced-motion media query to stop all moving transitions. Next, you can try to set up the twig templates.

MarcinMr:
@mikemai2awesome Could you please check the footer and its templates.

I also have a problem that I cannot figure out what I'm doing wrong.

in the `packages > ... > src > page-footer.twig` line 22. There is a string "Navigation heading" written manually.

In the `doc-site > ... > 40-components > 05-page-footer.twig` line 502 > `text: 'About Pegasystems'`

I cannot pass that value from line 502 `05-page-footer.twig` to line 22 in `page-footer.twig`. I receive an error in the terminal that there is a NULL value found. But when I dumped the values below the footer you can see that the value is not NULL. This happens only for `content.headline.text`; `content.headline.size` and `content.headline.tag` work correctly.

files

error

mikemai2awesome (approved):
@MarcinMr I could not figure out the error you were having, I changed the twig to use another method instead. I also cleaned up the rest. For future reference, please remember to use proper indentations for all code, it makes it really hard to read when indentations are all over the place. Thanks!

packages/components/bolt-page-footer/package.json
```diff
@@ -1,18 +1,24 @@
...
   "homepage": "https://boltdesignsystem.com",
   "bugs": "https://github.com/bolt-design-system/bolt/issues",
   "repository": "https://github.com/bolt-design-system/bolt/tree/master/packages/components/bolt-page-footer",
   "license": "MIT",
-  "style": "page-footer.scss",
+  "main": "index.js",
+  "style": "index.scss",
   "dependencies": {
-    "@bolt/components-band": "^4.2.0",
+    "@bolt/components-typeahead": "^4.2.0",
```
colbytcook, line 12: @MarcinMr Do we know why the typeahead component is listed as a dependent here? It doesn't seem to be used in the page footer package.

MarcinMr: @colbytcook Hey, I guess we shouldn't have this. We were using bands and typehead in the old footer version. Possible that when I was updating I didn't delete this or added it when I was testing something and forget to get rid of it?

packages/components/bolt-page-footer/page-footer-nav-ul.schema.js
```diff
@@ -0,0 +1,46 @@
...
+      description:
+        'Headline of the nav list. This must be set for each nav list.',
+      properties: {
+        attributes: {
+          type: 'object',
+          description:
+            'A Drupal attributes object. Applies extra HTML attributes to the parent element.',
+        },
+        content: {
+          type: 'string',
```
remydenton, line 34: Change type of `content` from type `string` to `any`

packages/components/bolt-page-footer/src/page-footer.twig
```diff
@@ -0,0 +1,29 @@
...
+    </div>
+    {{ primary_nav }}
+  </nav>
+  <aside>
+    <nav class="c-page-footer__nav c-page-footer__nav--aside" aria-label="{{ 'Utilities'|t }}">
+      {{ secondary_nav }}
+      <span class="c-page-footer__nav-item c-page-footer__nav-item--copyrights">&copy;{{ 'now'|date('Y') }} Pegasystems Inc.</span>
+    </nav>
+  </aside>
+  <span id="page-footer-new-window" hidden>{{ 'Open in new window'|t }}</span>
```
remydenton, line 28: 👍  Nice usage of `aria-describedby`.

docs-site/src/pages/pattern-lab/_patterns/40-components/page-footer/05-page-footer.twig
```diff
@@ -0,0 +1,487 @@
...
+    description: footer_description,
+    primary_nav: primary_nav,
+    secondary_nav: secondary_nav
+  } only %}
+{% endset %}
+
+{% set twig_markup %}
+{% verbatim %}
+// The main template
+{% include 'bolt-components-page-footer/page-footer.twig' with {
```
remydenton, line 452: `bolt-components-page-footer` is missing a leading `@` symbol.  Same for the other includes inside this verbatim section.

remydenton (changes requested):
This looks great @MarcinMr, I had only a few very minor inline comments.

One other question, how come we didn't have to add this component to the Base starter kit's .boltrc, [like this](https://github.com/boltdesignsystem/bolt/blob/feature/DS-520-refactor-site-footer-for-pega.com/packages/base-starter-kit/.boltrc.js#L53)?  I think that's required for it to be used in Drupal, though I thought it wouldn't work in pattern lab without that either.  @colbytcook, @mikemai2awesome, or @danielamorse, do you any of you guys know how this is working without that?

packages/components/bolt-page-footer/page-footer-nav-ul.schema.js
MarcinMr: It's done

docs-site/src/pages/pattern-lab/_patterns/40-components/page-footer/05-page-footer.twig
MarcinMr: I added missing '@' ;)

remydenton (approved):

remydenton (changes requested):
Wait, hold on, did you see this question @MarcinMr?  We can discuss in dev huddle tomorrow if it doesn't make sense.

> One other question, how come we didn't have to add this component to the Base starter kit's .boltrc, like this? I think that's required for it to be used in Drupal, though I thought it wouldn't work in pattern lab without that either.

MarcinMr:
> Wait, hold on, did you see this question @MarcinMr? We can discuss in dev huddle tomorrow if it doesn't make sense.
> 
> > One other question, how come we didn't have to add this component to the Base starter kit's .boltrc, like this? I think that's required for it to be used in Drupal, though I thought it wouldn't work in pattern lab without that either.

I'm sorry @remydenton I've missed that. Now I updated your request. I added `page-footer` component to the `base-starter-kit > .boltrc.js`

So my question is: every new component has to be added to the `> base-starter-kit > .boltrc.js` ? If so I'm curious like you how could this work without having it in `.boltrc.js` :)

remydenton:
Thanks for the update @MarcinMr.  I solved the mystery-- there was a separate entry for the footer component in the PL-only `.boltrc` file.  Here's the relevant [comment in that file](https://github.com/boltdesignsystem/bolt/blob/f637b83eb1b32be6c7c1ca26ced05702bab7e6ab/docs-site/.boltrc.js#L85-L88):

```
/* IMPORTANT: Do NOT list components here that are intended to be used
 * outside of the docs site-- those should instead be pulled in through
 * baseConfig.  Use this only for internal components.
 */
```

It makes sense that this was here previously, since the previous footer wasn't meant for consumption outside of Bolt.  But now that that's changed, we want it in base-starter-kit instead.  To answer this question:

```
So my question is: every new component has to be added to the > base-starter-kit > .boltrc.js ?
``` 
Correct, all new components that are meant to be used outside of Pattern Lab (which should be most of them) should get added to the base-starter-kit boltrc file.

remydenton (approved):

PR #2282, opened 2021-07-26
DS-574: Disable cssnano `calc` optimizations

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-574

## Summary

Disable `cssnano` "calc" optimizations.

## Details

`cssnano` is breaking certain "calc" rules when it tries to optimize them.

We use the ["default" config option](https://cssnano.co/docs/optimisations) with a handful of [additional options disabled](https://github.com/boltdesignsystem/bolt/blob/7aa85514e8309775f1bb6c25c1d1c53998af19cc/packages/build-tools/create-webpack-config.js#L344-L349). I'm adding "calc" to the list of disabled optimizations.

Down the road we will switch to the "lite" config. For now, just to be safe, disable only "calc".

## How to test

- Review files changed
- Smoke test of PL site


mikemai2awesome (approved):

remydenton (approved):
Thanks for this @danielamorse, good find!  It's frustrating that cssnano seems to be too aggressive out of the box, but I think your update is the right approach (as opposed to just removing it).

PR #2285, opened 2021-07-29
DS 557 Deprecate status blue and typo fix

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-557

## Summary

Color name `blue` was changed to `info`. Subheading was changed on the [color system](https://boltdesignsystem.com/pattern-lab/?p=viewall-visual-styles-color) doc page.

## How to test

Check if there is no instances of `blue` and if those instances are changed to `info`


mikemai2awesome (approved):
I see zero usage by searching on Gitlab, I think this is safe to remove without a Remove ticket. @colbytcook

colbytcook (approved):

PR #2292, opened 2021-08-05
DS-409 Update plop component generator

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-409

## Summary

Schema, twig and test templates were updated to our standards

## Details

A branch [feature/DS-408-Convert-Yeoman-to-Plop-alternative](https://github.com/boltdesignsystem/bolt/pull/2274) was pulled to this one.

- A new prompt was added to the component generator that will render a different JS, Jest Test and Twig file if this component is a web component.
- `title` and `content` were added to `component.schema`
- `component.docs.twig` was updated with single quotes
-  `component.twig` was updated with single quotes, `attributes` was changed with `props.class`, `title` and `content` were added to the template
- `component.test.js` was updated with commented code in case to use, according to the master schema
- `element.docs.twig` was updated with single quotes

## How to test

1. Pull the branch locally and run `yarn setup`.
2. Run `yarn cc` to generate component and check if all the files are generated correctly (remember to run `yarn` after).
3. On the "Is this component a web component?:" press enter ("n" is the default value)
4. Run `yarn cc` to generate component and check if all the files are generated correctly (remember to run `yarn` after).
5. On the "Is this component a web component?:" question type "y"
6. Run `yarn ce` to generate element and check if all the files are generated correctly (remember to run `yarn` after).


danielamorse (approved):
Looks good! Thanks @MarcinMr. I just made [one formatting fix](https://github.com/boltdesignsystem/bolt/pull/2292/commits/c67d342b767a2ac96e5069e1c58ad9e92320db91).

colbytcook (approved):

cjwhitedev (changes requested):
1. if we're trying to make less and less web components, do we want the default generator to be a web component out of the box? maybe we can add a step in the generator to ask if its a web component, then generate a separate directory.
2. in the twig file it doesnt look to me like we are properly leveraging `this.data` / validation  out of the box.
3. in the test file, are the right things commented out? if we are creating a web component by default, we should be testing a web component by default. if we go with any changes from my first comment above, then the test file should be rendered differently as well, right?

mikemai2awesome:
yes, we should chat about CJ's points.

adamszalapski (changes requested):
Why did we change the name of the package? If this was already. discuss we should find all `bolt-generator` occurrences and change it to `generator-bolt` because of other tests and some exceptions based on this name.

colbytcook (approved):

packages/generators/bolt-generator/__tests__/component/__snapshots__/bolt-generator.js.snap
```diff
@@ -199,34 +194,38 @@ beforeAll(async () => {
...
+        ...fixtures.defaultData,
+      },
+    );
+
+    await expect(results.ok).toBe(true);
+    await expect(results.html).toMatchSnapshot();
   });
 });
+
+// describe('Bolt Test prop -', () => {
```
danielamorse, line 213: @MarcinMr, @colbytcook if this is the new convention for Jest tests, would you please also update the [wiki](https://github.com/boltdesignsystem/bolt/wiki/Jest-Test:-Example-Jest-Test)?

danielamorse (commented):
This looks good except for the tests. I'm running into some Jest test errors locally that I'll need more time to debug. Tests pass individually, and then when I run `yarn test:js` several generator tests fail.

packages/generators/bolt-generator/__tests__/component/__snapshots__/bolt-generator.js.snap
```diff
@@ -199,34 +194,38 @@ beforeAll(async () => {
   fixtures = {
     defaultData,
   };
-
 });
 
 describe('Bolt Test component', () => {
```
danielamorse, line 199: @MarcinMr @colbytcook same comment here about updating the wiki with the "component/element" convention if that's the new standard.

adamszalapski (approved):
Looks good. I agree with @danielamorse all-new approaches/conventions should be updated/added to our WIKI.

cjwhitedev (approved):

packages/generators/bolt-generator/__tests__/component/__snapshots__/bolt-generator.js.snap
danielamorse: Updated: https://github.com/boltdesignsystem/bolt/wiki/Jest-Test:-Example-Jest-Test

packages/generators/bolt-generator/__tests__/component/__snapshots__/bolt-generator.js.snap
danielamorse: Updated: https://github.com/boltdesignsystem/bolt/wiki/Jest-Test:-Example-Jest-Test

danielamorse (changes requested):
I'm seeing Jest test failures when I do the following:
1. Run `yarn test:js` (all tests pass... although I was seeing unrelated intermittent `bolt-text` fails, disregard for now)
2. Run `npx jest packages/generators/bolt-generator/__tests__/component/bolt-generator.js` (pass)
3. Run `npx jest packages/generators/bolt-generator/__tests__/element/bolt-generator.js` (pass)
4. Run `yarn test:js` (fail, [see error](https://gist.github.com/danielamorse/5c5ad4651dd98b338f830bc510a0ff9d))

The second time you run `yarn test:js` it produces different snapshots. It seems like something is left over from previous tests (perhaps in `tmp`?) that changes the test results.

I discovered this when I had to update the individual generator snapshots. Then I ran all tests and saw the error. It's an edge case, but we should fix it before this gets merged.

@adamszalapski any thoughts on why this might happen?

adamszalapski:
Good catch @danielamorse 
This issue's related to test asynchronously and how our bolt-generator uses `tmp` folder for checking the file structure. I updated the approach how we use `tmp` folder and now the issue is gone. This was a tricky one to catch, good work @danielamorse 👍

adamszalapski:
@danielamorse I update whole folder structure for `bolt-generator` tests, right now temp folders are generated in specific (component or element) folder. At the moment all test passing without any issue, please double check if everything is good.
image

danielamorse (approved):
@adamszalapski works perfectly now. Thank you!

PR #2294, opened 2021-08-06
DS-329 Remove Button from Micro Journey Editor

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-329](https://pegadigitalit.atlassian.net/browse/DS-329)

## Summary

Removed Bolt-Button option from Micro Journey Editor

## Details

Bolt-Button has been deprecated and slated for removal. Micro Journeys is also to be removed eventually. So rather than resolve any issues one might cause the other, we thought it best just to remove the button from the editor.

## How to test

Review my code

## Release notes

Removed Bolt-Button option from Micro Journey Editor


danielamorse (commented):
AFAICT there is no actual usage of `bolt-button` in micro journeys content, so this should be ok. Having [Ross verify that](https://pegadigitalit.atlassian.net/browse/DS-329) is also reassuring.

@remydenton does WWO use this same editor? Does WWO use Bolt Button?

remydenton (approved):
This looks great, thanks for the updates @cjwhitedev.  I'm glad we realized we could just remove rather than sinking more time into making it work.

@danielamorse , WWO actually doesn't use the editor.  WWO instances are created with fixed markup written by a developer, not a content author.  It means a little less flexibility, but also way less complexity.

PR #2295, opened 2021-08-06
Fix: holy grail JS, side nav JS classes, and button theme colors

mikemai2awesome:
## Summary

Fixes Holy Grail overflow bug, side nav JS classes, and button theme colors.

## Details

1. Updated the overflow hidden util class to be applied based on breakpoint in Holy Grail JS
2. Updated JS selectors to use `js-` classes in side nav
3. Updated theme colors so that button text has enough contrast across all light themes
4. Updated Community pages to reflect design changes

## How to test

1. Check the community doc layout page and make sure overflow hidden isn't applied to the `<body>` when above the medium breakpoint and the side nav is expanded.
2. Check the side nav docs, make sure the expand/collapse function still works.
3. Check the theming system doc page, make sure button text color has good contrast with button background color in Navy-light, Teal, Yellow, Orange, Pink, Wine, Berry, and Violet themes.


danielamorse (approved):
Verified fixes to JS (my own changes) and color contrast on buttons.

PR #2298, opened 2021-08-17
Fix read-only repo publishing

remydenton:
## Jira

N/A

## Summary

Fixes broken publishing to read-only repos ([bolt_connect](https://github.com/boltdesignsystem/bolt_connect) and [core-php](https://github.com/boltdesignsystem/core-php))

## Details

Previously, changes weren't being published to these repos due to a hard-coded condition that only allowed publishing from the `release/1.x` or `release/2.x` branches.  This PR removes all conditions, under the assumption that any conditions for running or not running this script should happen outside this script itself.

Additionally, I updated the old domain (bolt-design-system.com) to the new (boltdesignsystem.com)

## How to test

Any merges to a `release/*` branch should run this script now.  Alternatively, you can run it locally.

Note that the script will update the two read-only github repos as long as you have permission to do so, but you may need to manually update the respective packagist projects (the webhook to push from Github to Packagist was broken last time we tried). 
https://packagist.org/packages/bolt-design-system/bolt_connect
https://packagist.org/packages/bolt-design-system/core-php

### DIFF
--- a/scripts/release/update-read-only-git-repos.sh
+++ b/scripts/release/update-read-only-git-repos.sh
@@ -12,13 +12,13 @@ if [ -n "$(git status --porcelain)" ]; then
 fi
 
 echo "Updating remote subsplit repos"
-./scripts/release/git-subsplit.sh init https://$GH_TOKEN@github.com/bolt-design-system/core-php.git
-./scripts/release/git-subsplit.sh publish --work-dir=$PWD packages/twig-integration/twig-extensions-shared:https://$GH_TOKEN@github.com/bolt-design-system/core-php.git \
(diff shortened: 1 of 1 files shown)


danielamorse (approved):
Makes sense and helps to remove unnecessary noise from this script.

If I understand correctly, this script is only ever run from [here](https://github.com/boltdesignsystem/bolt/blob/master/.travis.yml#L168), which is governed by the Auto-release ["if" statement](https://github.com/boltdesignsystem/bolt/blob/master/.travis.yml#L153) above it.

I'm not going to merge into release to test, so I say just :shipit:

PR #2301, opened 2021-08-20
DS-565 My Quick Links 2.0: Manage My Quick Links page (UX Mockup)

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-565

## Summary

A mockup of "My Quick Links 2.0" was created in the Pattern Lab sandbox.

#### Desktop:

- two columns are created with a list of links
- ~~visible scrollbar is added~~ (feedback request: add default scrollbar)
- skip links are added to switch between columns
- ~~aria-describedby is added to make items more accessible~~ (Changed to invisible descriptive`span`s)
- spaces between list items are increased
- descriptive titles of the columns are created
- tooltips are removed
- external link icons were added

#### Mobile

- "add Items to my quick links" column is switched to modal
- ~~fixed bottom bar with the modal trigger is added~~ (feedback request: modal trigger is static above the list)
- skip links are triggers to switch between modal and the list

## How to test

Pull the branch. Check the mockup if it's accurate to the sent wireframes. Check if UX and its elements are accessible.


docs-site/src/pages/pattern-lab/_patterns/50-pages/99999-bolt-dev-sandbox/15-my-quick-links-2.0.twig
```diff
@@ -0,0 +1,200 @@
...
+          tag: 'ul',
+          items: [
+            quick_link_1,
+            quick_link_2,
+            quick_link_3,
+          ]
+        } only %}
+
+      {% endset %}
+      {% include '@bolt-components-card-replacement/card-replacement.twig' with {
```
mikemai2awesome, line 142: Don't use a card for this. You will need to style a scrollable container around just the list of links (like in the wireframe), play around with the max-height, maybe 50vh. On mobile, you can adjust the height again to further optimize the experience when the 2 columns stack into 1.

docs-site/src/pages/pattern-lab/_patterns/50-pages/99999-bolt-dev-sandbox/15-my-quick-links-2.0.twig
```diff
@@ -0,0 +1,200 @@
...
+        body: {
+          content: demo,
+        },
+        theme: 'xlight'
+      } only %}
+    {% endcell %}
+
+    {% cell "o-bolt-grid__cell u-bolt-width-1/1 u-bolt-width-1/2@small" %}
```
mikemai2awesome, line 150: I believe you will need specific CSS to make this interface work well across all the breakpoints, so best to write something that will work for this specific case rather than using a generic grid.

mikemai2awesome (changes requested):

docs-site/src/pages/pattern-lab/_patterns/50-pages/99999-bolt-dev-sandbox/15-my-quick-links-2.0.twig
```diff
@@ -0,0 +1,200 @@
+{% extends "@pl/_default-page-template.twig" %}
+
+{% set icon_add %}
+  {% include "@bolt-components-icon/icon.twig" with {
+    name: "add-solid",
+    size: "medium",
```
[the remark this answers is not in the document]
MarcinMr: icon size removed

docs-site/src/pages/pattern-lab/_patterns/50-pages/99999-bolt-dev-sandbox/15-my-quick-links-2.0.twig
```diff
@@ -0,0 +1,200 @@
...
+
+{% set icon_trash %}
+  {% include "@bolt-components-icon/icon.twig" with {
+    name: "trash",
+    size: "medium",
+  } only %}
+{% endset %}
+
+{% set quick_link_1 %}
+  {% include '@bolt-elements-text-link/text-link.twig' with {
```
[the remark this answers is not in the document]
MarcinMr: reverse_underline added

docs-site/src/pages/pattern-lab/_patterns/50-pages/99999-bolt-dev-sandbox/15-my-quick-links-2.0.twig
MarcinMr: Card component removed. Created two columns layout with scrollable lists

docs-site/src/pages/pattern-lab/_patterns/50-pages/99999-bolt-dev-sandbox/15-my-quick-links-2.0.twig
MarcinMr: Generic grid deleted and replaced with specific CSS

colbytcook:
@MarcinMr @mikemai2awesome should we be including a tooltip that helps describe the action? When i pulled down the branch and started to interact with the page I didn't understand the proposed functionality immediately. Now this could be due to this being out of context with the expected page or me just not paying attention. But i figured I would bring it up anyways.

With that said would it make sense to add in a tooltip that will display the message "Delete Agenda Manager" or "Add Atlas"? Would that be too busy?  I also have a question regarding the use of the trash can icon here. Since the functionality of adding and subtracting from a list is grouped together in one design would it make sense to update to a "-" instead? To me, that seems to read closer to the list curation functionality.

MarcinMr:
@colbytcook @mikemai2awesome I was convinced that provided description in a subheading will be enough for the proposed functionality. I also added `aria` attributes for screen readers for better understanding. But yes, maybe we should consider adding tooltips ? I added them to the first list for comparison and to check if this won't be too busy. We can decide what would be better.

When it comes to icons, I absolutely agree that "minus" symbol would suit better. But do we have such an icon in our library ?The only reason I added the "trash" icon was I couldn't find "minus" icon. If we don't have "substract" symbol maybe we should upload new icons like these below but with opposite sign (-)

Screenshot 2021-09-01 at 09 58 15

mikemai2awesome:
@MarcinMr @colbytcook I added back the minus icons for you to try out and pulled in master, so you can use the icon element. I can't get the icon schema to update though, so I will leave that to you to fix.

MarcinMr:
> @MarcinMr @colbytcook I added back the minus icons for you to try out and pulled in master, so you can use the icon element. I can't get the icon schema to update though, so I will leave that to you to fix.

@mikemai2awesome Icon is switched and schema update with the minus icon.

MarcinMr:
> @MarcinMr That is much better, I agree. Please proceed with your proposed method.

Updated @mikemai2awesome

MarcinMr:
@mikemai2awesome 

There is one more thing. If we have two links in every single `<li>` the voice-over reads (for example first column) that there is 3 list item and that's true. But going through the list with the `tab` key we actually have to press it six times so this could be misleading I guess. I updated and now we have every single link in a single `<li>` so the screen reader says that there is a six items list (first column). Let me know if that's okay or I revert the commit if necessary to the previous one.

mikemai2awesome:
@MarcinMr It's good now. Thanks!

mikemai2awesome (approved):

mikemai2awesome (changes requested):
Sorry, I was looking at outdated code and misread what you said. The 2 links should be contained within just one `<li>` instead, not separate `<li>`s.

MarcinMr:
> Sorry, I was looking at outdated code and misread what you said. The 2 links should be contained within just one `<li>` instead, not separate `<li>`s.

Ok :) I reverted this commit to the previous one

mikemai2awesome (approved):

docs-site/src/pages/pattern-lab/_patterns/50-pages/80-portal/quick-links-styles.scss
```diff
@@ -0,0 +1,132 @@
+.c-portal {
```
mikemai2awesome, line 1: Not critical but this should be named `.c-portal-quick-links`.

colbytcook: To keep with the selector pattern should I update all of the classes to follow this change? For example `.c-portal__col` turns to o`.c-portal-quick-links__col`?

PR #2308, opened 2021-09-03
DS-593 Patternlab Icon Element Update

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-593

## Summary

Icon Component was replaced with Icon Element

## Details

- All the examples of the Icon Component were replaced with the Icon Element and Shape Element in `/doc-site` directory.
- New icon names were used.

## How to test

Check if all instances of `@bolt-components-icon/icon.twig` in `/doc-site` directory were replaced with `@bolt-elements-icon/icon.twig`

## Release notes

The `Icon Component` is deprecated and has been replaced with `Icon Element`.


MarcinMr (commented):
I didn't find a minus icon in our new icon element library. There is minus icon but in the component icon library. I needed that icon to update one of the pages on the academy so I uploaded a "minus-circle" for now. If we don't use the minus icon anymore just let me know and I'll revert this commit.

Screenshot 2021-09-03 at 11 21 07

remydenton:
I merged in the latest master to fix the build issues.  The replacements in general look right to me.

Regarding the minus icon, it looks like Academy is using `face-happy` and `face-sad` icons now for that section.  For example: https://academy.pega.com/topic/pcm-interactions/v1.  I would say update the demo to match those and remove the minus icon that was added.

Finally, since the ticket says to deprecate the icon component, I think you'll want to match what we're doing for other deprecated components like [link](https://github.com/boltdesignsystem/bolt/tree/feature/DS-593-patternlab-icon-element-update/docs-site/src/pages/pattern-lab/_patterns/40-components/link-deprecated), [video](https://github.com/boltdesignsystem/bolt/tree/feature/DS-593-patternlab-icon-element-update/docs-site/src/pages/pattern-lab/_patterns/40-components/video-deprecated), etc.  I don't think we have docs on how to do that, but just match the examples so that the main page in the docs site shows it as deprecated.

remydenton (changes requested):
See

docs-site/src/pages/pattern-lab/_patterns/40-components/hero/30-hero--image-min-width.twig
```diff
@@ -3,17 +3,22 @@
 
 <bolt-text>This demo creatively uses the <code>imageMinWidth</code> prop to "pin" the image to the right side of the screen on larger devices.</bolt-text>
 
+{% set icon %}
+  {% include "@bolt-elements-icon/icon.twig" with {
+    name: "pega-compass",
```
colbytcook, line 8: @MarcinMr @mikemai2awesome The only discrepancy I see during these icon updates is that **pega-compass** and **brand-operations** do not match. First, i wanted to double check that the change is expected (I know we are re-mapping old icons to new ones). Also if this is intended we should add a **Visual Changes** section to the PR description to outline this.

remydenton (approved):
Thanks for the updates @MarcinMr, looks good (though waiting to merge until others have had a chance to review).

docs-site/src/pages/pattern-lab/_patterns/40-components/hero/30-hero--image-min-width.twig
MarcinMr: @colbytcook I used that icon because I thought that we no longer have a brand-operations icon or some similar to that. But it turns out we have a similar icon -  it's a speed-dial icon I believe. I pushed changes.

colbytcook (approved):

docs-site/src/pages/pattern-lab/_patterns/40-components/hero/30-hero--image-min-width.twig
mikemai2awesome: yes, it has been renamed to speed-dial

mikemai2awesome (approved):

PR #2309, opened 2021-09-03
Upgrade chromedriver

danielamorse:
## Jira

n/a

## Summary

Force upgrade `chromedriver`.

## Details

Change `chromedriver` dependency syntax from `latest` to `*` to see if it makes any difference in the future. If this fails again, both npm and Travis are getting hardcoded. cc @remydenton @colbytcook @adamszalapski 

## How to test

Build is passing

### DIFF
--- a/packages/testing/testing-nightwatch/package.json
+++ b/packages/testing/testing-nightwatch/package.json
@@ -18,7 +18,7 @@
   "main": "nightwatch.js",
   "dependencies": {
     "@zeit/fetch-retry": "^4.0.1",
-    "chromedriver": "latest",
+    "chromedriver": "*",
     "ci-utils": "^0.6.0",
     "geckodriver": "^1.16.2",
     "globby": "^10.0.1",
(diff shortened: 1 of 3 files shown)


remydenton (approved):
As discussed in Slack, the additional change we want here is to lock the Chrome version that gets installed by Travis.  That way, the locked version of chromedriver matches the Travis version of Chrome until we manually update both together (i.e. not in the middle of other development).

I'm approving this as-is though in case we want to merge it to unblock other PRs before we figure out how to lock the Travis version.

danielamorse:
Thanks @remydenton. From the [Travis docs](https://docs.travis-ci.com/user/chrome#selecting-a-chrome-version):
> You can install the `stable` or the `beta` version of Chrome but you can’t select a specific numeric version.

There might be a way around this, but I'm going to stop here and merge this fix. I say we continue to manually update this until Nightwatch is gone.

PR #2314, opened 2021-09-09
WWW-960: mobile nav top link color

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/WWW-960

## Summary

Makes unselected top-level mobile links black for consistency

## Details

See screenshots in the "visual changes" section below.  All other colors should be unchanged.

## How to test

See screenshots in the "visual changes" section below.  All other colors should be unchanged.

### Visual changes

In the global nav, top-level links are now black instead of dark navy when not selected.  For example, on https://boltdesignsystem.com/pattern-lab/?p=components-page-header-example-main-site-with-subnav

Before:
Screen Shot 2021-09-10 at 10 44 12 AM

After:
Screen Shot 2021-09-10 at 10 44 24 AM

### DIFF
--- a/packages/components/bolt-page-header/src/_page-header-mobile.scss
+++ b/packages/components/bolt-page-header/src/_page-header-mobile.scss
@@ -183,11 +183,6 @@
     padding-left: $bolt-page-header-mobile-expanded-content-spacing-x;
     color: var(--m-bolt-text);
     border-radius: 0;
-
-    &:hover,
-    &:focus {
-      color: var(--m-bolt-headline);
-    }
   }
 
   .c-bolt-page-header__nav-list-item.is-selected .c-bolt-page-header__nav-link {
--- a/packages/components/bolt-page-header/src/_page-header-mobile.scss
+++ b/packages/components/bolt-page-header/src/_page-header-mobile.scss
@@ -183,6 +183,11 @@
     padding-left: $bolt-page-header-mobile-expanded-content-spacing-x;
     color: var(--m-bolt-text);
     border-radius: 0;
+
+    &:hover,
+    &:focus {
+      color: var(--m-bolt-headline);
+    }
   }
 
   .c-bolt-page-header__nav-list-item.is-selected .c-bolt-page-header__nav-link {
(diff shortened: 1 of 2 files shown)


mikemai2awesome (changes requested):
Page header is very complicated. Your change affected some other elements. I need a closer look tomorrow.

mikemai2awesome (approved):

mikemai2awesome:
All set.

remydenton:
With the latest changes, top level links have a different color in focus state (black)...
Screen Shot 2021-09-10 at 11 45 15 AM

... than other links in focus state (dark navy)
Screen Shot 2021-09-10 at 11 46 14 AM

I actually don't see a good reason to have an `[aria-expanded='true']` style for mobile links, since they have identical styling to other links, hence my commit comment in [a2e752](https://github.com/boltdesignsystem/bolt/pull/2314/commits/a2e752).

At the end of the day, I don't really care as long as stakeholders are happy, and you're the original author @mikemai2awesome, so your call.

mikemai2awesome:
@remydenton They are both in expanded state, so the color should be kept consistent.

colbytcook (approved):

remydenton:
> They are not both in expanded state (one is acting as a heading and expanded, the other is an item and not expanded), so the color should be kept consistent (expanded state is black).

I'm aware they're not both expanded.  If you're trying to show that they're in different states, why do they look the same when not in focus state?

IMO, they should look identical, unless you want to distinguish between expanded and not-expanded links, in which case they should be styled differently both when focused and not.

Pretty sure the original PR makes more sense, unless there are other situations I'm not aware of.

mikemai2awesome:
@remydenton They shouldn't look the same intially. I intended headlines to take on headline color, links to take on link color, but I'm forced to change them to black which makes no sense.

remydenton:
> I'm forced to change them to black which makes no sense.

Fair, I think that's probably the root cause of why it feels weird to me.  I'm fine with this as-is, but let me have stakeholders take a look at the Pattern Lab demo link first so that we don't run the risk of yet more change requests 😄 .

colbytcook:
@mikemai2awesome @remydenton it looks like Ross responded to the JIRA question, we should be good to merge this one!

PR #2316, opened 2021-09-10
DS-582 Add information about twig file name on schema section

adamszalapski:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-582

## Summary

Add information about the twig file name on the schema section.

## Details

Applies to components with multiple schemas:
- packages/components/bolt-listing-teaser
- packages/components/bolt-navbar
- packages/components/bolt-page-footer
- packages/components/bolt-page-header
- packages/components/bolt-side-nav
- packages/layouts/bolt-holy-grail
- packages/layouts/bolt-layout

## How to test

Run this code locally and check if the twig file name is shown on components with multiple schemas and it's not visible on single schemas.

### DIFF
--- a/docs-site/src/components/pattern-lab-utils/schema-docs.twig
+++ b/docs-site/src/components/pattern-lab-utils/schema-docs.twig
@@ -10,7 +10,12 @@
     {% for schema in schemas %}
       {% set title = schema.title %}
       <bolt-accordion-item{% if loop.first %} open{% endif %}>
-        {% if title %}<strong slot="trigger">{{ title }}</strong>{% endif %}
+        {% if title %}
+          <div slot="trigger">
+            <strong>{{ title }}</strong>
+            <span>({{ title|replace({' ': '-'})|lower }}.twig)</span>
+          </div>
+        {% endif %}
         {{ macros.schemaWrapper(schema) }}
       </bolt-accordion-item>
     {% endfor %}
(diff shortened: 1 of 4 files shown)


danielamorse (commented):
@adamszalapski code looks good 👍

Do you know why these changes aren't reflected on [Accordion](https://feature-ds-582.boltdesignsystem.com/pattern-lab/?p=viewall-components-accordion)? I recall the logic of the schema template is kind of convoluted. Accordion might match some condition that these other components do not.

[Page Header](https://feature-ds-582.boltdesignsystem.com/pattern-lab/?p=viewall-components-page-header
) looks good. [Navbar](https://feature-ds-582.boltdesignsystem.com/pattern-lab/?p=viewall-components-navbar) is showing incorrect Twig template names. Before we merge this, we should review each component with multiple schemas and fix the names in the schema where necessary. Let's discuss today in huddle.

colbytcook:
@danielamorse @adamszalapski the reason the Accordion and the Grid are not showing the twig names is because they are using the old version of including the nested schemas. The new version you list an array of schemas in the package.json, the older model you include those files directly in the schema file. 

I am not sure if we should proceed to just update those two components to the new standard way or try to make this PR work with them. I think we should update those two components to the latest method.

danielamorse (approved):
Checked again with @colbytcook's update. Verified all the components/layouts with multiple schemas match the filenames in the docs.

colbytcook (approved):
@adamszalapski As part of this PR can you update the schema "title" of

PR #2323, opened 2021-09-16
[DO NOT MERGE] DS 615 Convert 'string' to 'any' in schema where appropriate

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-615

## Summary

`String` props were changed to `any` across all `.schema`'s files.

## Details

`String` props were changed to `any` across all `.schema`'s in all:

- bolt elements
- bolt components

Props converting were omitted if an enumeration was found in a code block.

## How to test

Pull the branch, Make sure all `string` props were changed to `any`. Check if there is more to change or if some of them were changed unnecessarily.

### DIFF
--- a/packages/components/bolt-card-replacement/__tests__/__snapshots__/card-replacement.js.snap
+++ b/packages/components/bolt-card-replacement/__tests__/__snapshots__/card-replacement.js.snap
@@ -143,36 +143,10 @@ exports[`Bolt Card Props border_radius items: large 1`] = `
       <ssr-keep for="bolt-card-replacement-actions"
                 class="c-bolt-card_replacement__actions"
       >
-        <bolt-card-replacement-action url="https://pega.com">
+        <bolt-card-replacement-action>
           <ssr-keep for="bolt-card-replacement-action"
                     class="c-bolt-card_replacement__action"
           >
-            <a href="https://pega.com"
-               target="_self"
-               class="e-bolt-button e-bolt-button--transparent e-bolt-button--block"
-            >
-              This is a button
-              <span class="e-bolt-button__icon-after"
-                    aria-hidden="true"
-              >
-                <bolt-icon name="chevron-right"
-                           aria-hidden="true"
-                >
-                  <span class="c-bolt-icon c-bolt-icon--chevron-right">
-                    <svg class="c-bolt-icon__icon"
-                         xmlns="http://www.w3.org/2000/svg"
-                         viewbox="0 0 32 32"
-                    >
-                      <path fill="var(--bolt-theme-icon, currentColor)"
-                            fill-rule="evenodd"
-                            d="M13 7c-.6-.5-1.4-.5-2 0-.5.6-.5 1.4 0 2l7.1 7-7 7a1.3 1.3 0 00.9 2.3c.3 0 .7-.1 1-.4l8-8c.5-.5.5-1.3 0-1.8l-8-8z"
-                      >
-                      </path>
-                    </svg>
-                  </span>
-                </bolt-icon>
-              </span>
-            </a>
           </ssr-keep>
         </bolt-card-replacement-action>
       </ssr-keep>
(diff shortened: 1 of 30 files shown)


MarcinMr:
@remydenton 

I went thru `.schema` files and converted `string` to `any` I wasn't sure of each of them. So if you find some unnecessary changes or places where the change should be please let me know and I'll fix them.

I was not sure if there were examples such as:  `src`,`url`, `hrefs`, `ids`.

And how about `[string, object,]` or `[string, object, array]`. If there is more than one type of data shouldn't it be converted to `any`?

packages/components/bolt-card-replacement/__tests__/__snapshots__/card-replacement.js.snap
```diff
@@ -143,36 +143,10 @@ exports[`Bolt Card Props border_radius items: large 1`] = `
       <ssr-keep for="bolt-card-replacement-actions"
                 class="c-bolt-card_replacement__actions"
       >
-        <bolt-card-replacement-action url="https://pega.com">
```
remydenton, line 146: This is another test that is showing that something that's needed is now missing.

The lesson here @MarcinMr is that you should always double check when a test snapshot has failed and ask whether it makes sense.  Sometimes it does and you can safely just update the snapshot, but other times (like this) it is catching a real error.

packages/components/bolt-card-replacement/card-replacement.schema.js
```diff
@@ -125,11 +125,11 @@ module.exports = {
         type: 'object',
         properties: {
           text: {
-            type: 'string',
+            type: 'any',
             description: 'Action link text.',
           },
           url: {
-            type: 'string',
```
remydenton, line 132: The problem with changing this from `string` to `any` is that it will not longer pass through init() and appear appear in `this.data`:

https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig#L6

That's because it no longer returns true from `isAllowedSchemaType()`:

https://github.com/boltdesignsystem/bolt/blob/master/packages/twig-integration/twig-extensions-shared/src/Utils.php#L206

Maybe the right answer in this case is to not use this.data in the twig template, and instead use `{{ url }}`.  Thoughts @danielamorse ?

packages/components/bolt-card-replacement/card-replacement.schema.js
```diff
@@ -125,11 +125,11 @@ module.exports = {
         type: 'object',
         properties: {
           text: {
-            type: 'string',
```
remydenton, line 128: This will also break for the same reason as `url` (we're using this.data.text).

https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig#L23

Like `url`, I believe the correct answer here is to change the schema to `any` and not use `this.data` in the twig template.

packages/components/bolt-card-replacement/card-replacement.schema.js
```diff
@@ -138,7 +138,7 @@ module.exports = {
               'Indicate if this is an external link. If true, the link will open in a new tab.',
           },
           icon: {
-            type: 'string',
```
remydenton, line 141: This will break for the same reason as `url` (we're using `this.data.icon`):

https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig#L33

However, unlike for `url` and `text`, I think we want to keep the schema type as `string` and add an `enum` for what strings are valid for icon name (import the valid values from the icon component).  The description should also be updated to indicate that this only accepts an icon name (perhaps it should actually accept a rendered icon as well, but we'll save that for another day).

packages/components/bolt-copy-to-clipboard/copy-to-clipboard.schema.js
```diff
@@ -33,27 +33,27 @@ module.exports = {
...
     copiedText: {
-      type: 'string',
+      type: 'any',
       title: 'DEPRECATED',
       description:
         'Use the custom_confirmation property instead if you need to change the confirmation text.',
       default: 'Copied!',
     },
     iconSize: {
-      type: 'string',
```
remydenton, line 49: We probably shouldn't update `DEPRECATED` props at all.  But, I couldn't help but notice that iconSize should be have an `enum` and should not be of type `any`.

remydenton (changes requested):
I've only looked through the first few examples, but it's clear that this is a much more delicate proposition than going through and bulk changing `string` to `any`.  You have to really understand what's going on under the hood, making sure to not cause conflicts with the use of `init()`/`this.data`, and also recognize when an `enum` is actually implied and was just omitted by the original author.

colbytcook:
Closing this PR for now, we will revisit this with a more granular approach (the ticket will have reference to this PR and the branch).

PR #2334, opened 2021-09-29
DS-648 Update "information" banner colors

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-648

## Summary

Update the background and text colors of "information" banner.

## Details

As requested by Academy/Collaboration Center, banners with "information" status will now have light blue background and dark blue text.

## How to test

- Review code changes
- Review Banner demos, particularly "status" demo

### Visual changes
- Banner with status "information" are now light blue with dark blue text. Previously they were white with black text.

### DIFF
--- a/packages/components/bolt-banner/src/banner.scss
+++ b/packages/components/bolt-banner/src/banner.scss
@@ -76,26 +76,12 @@
 
 .c-bolt-banner--status-information {
   position: relative;
-  color: var(--m-bolt-text);
-  border-left-color: var(--m-bolt-link);
+  color: var(--bolt-color-info-dark);
+  border-left-color: var(--bolt-color-info-dark);
   border-left-style: $bolt-border-style;
   border-left-width: $bolt-border-width * 3;
   box-shadow: 0 0 var(--bolt-spacing-y-xsmall) var(--m-bolt-border);
-  background-color: var(--m-bolt-bg);
-
-  &:before {
-    content: '';
-    display: block;
-    opacity: 0.2;
-    position: absolute;
-    top: 0;
-    right: 0;
-    bottom: 0;
-    left: 0;
-    pointer-events: none;
-    user-select: none;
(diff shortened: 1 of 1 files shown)


packages/components/bolt-banner/src/banner.scss
```diff
@@ -76,26 +76,12 @@
 
 .c-bolt-banner--status-information {
   position: relative;
-  color: var(--m-bolt-text);
-  border-left-color: var(--m-bolt-link);
+  color: var(--bolt-color-info-dark);
+  border-left-color: var(--bolt-color-info-dark);
```
mikemai2awesome, line 80: Use `--bolt-color-info` for the border. And then add the following:

```
@include bolt-repeat-rule(
  ('bolt-banner[status=information]', ':host(bolt-banner[status=information])')
) {
  --m-bolt-text: var(--bolt-color-info-dark);
  --m-bolt-headline: var(--bolt-color-info-dark);
  --m-bolt-link: var(--bolt-color-info-dark);
}
```

mikemai2awesome (changes requested):

packages/components/bolt-banner/src/banner.scss
danielamorse: @mikemai2awesome updated.

I don't think any of [these `:host()` rules](https://github.com/boltdesignsystem/bolt/blob/d45bc41da99d38e1e2ae76b2e53738acd647e1a8/packages/components/bolt-banner/src/banner.scss#L28-L57) are necessary, but I'm fine adding one for `[status="information"]` for consistency until this is reworked. The only reason we'd need the `:host()` selector is if the WC was used in an environment where bolt-global styles weren't loaded or inside the Shadow DOM (and none of our components put Banner inside their Shadow DOM).

mikemai2awesome (approved):

PR #2336, opened 2021-09-30
DS-619 Add support for attributes to each form select option

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-619

## Summary

Support for attributes to each form select option was added.

## Details

- Every file from `/doc-site` which use `select-option` was updated with an `attributes object`.
- `Form-select.twig` was updated with `attributes object` usage.

## How to test

Make sure passed `props` added to `attributes object` renders properly and work as expected.

## Release notes

Added support for passing attributes to each form select option, see demo page (TODO: add demo page link) for reference.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/40-components/form/60-form-full-campaign-landing.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/40-components/form/60-form-full-campaign-landing.twig
@@ -54,22 +54,34 @@
       {
         'type': 'option',
         'value': '',
-        'label': '- Select a Country -'
+        'label': '- Select a Country -',
+        'attributes' : {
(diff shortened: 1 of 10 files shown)


colbytcook (approved):
@mikemai2awesome this PR looks good, I tested out the `<optgroup>` as well and it works properly. 

Quick note, I also tested putting `selected` and `disabled` attributes on a option to see if we can get it to display within the select. This does work, an option ("- Select a Country-") will be displayed on load but once a user selects a another option this option will be unclickable. Not sure if this a use case we need or care about (or even if this is accessible) but I wanted to pass this along in case a stakeholder or Drupal developer ask for this behavior.

PR #2339, opened 2021-10-01
DS-468 Fix input height on focus

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-468

## Summary

Prevent inputs with label "before" from shifting height on focus.

## Details

As part of https://pegadigitalit.atlassian.net/browse/DS-468, this PR fixes the shift in height that you see when you focus on an input with label "before". Fixing this helps us match the height between the current input and the Multi-select component.

## How to test

- View demo with `labelDisplay: "before"`: `/pattern-lab/?p=components-form-label-positions`
- Click into the input and verify input height does not change.

### Visual changes

Form inputs with label "before" the field (not floating inside) will no longer shift in height when they get focus.

### DIFF
--- a/packages/components/bolt-form/src/form.scss
+++ b/packages/components/bolt-form/src/form.scss
@@ -616,8 +616,13 @@ $bolt-strength-indicator-transition: var(--bolt-transition);
 
     .c-bolt-input.is-filled,
     .c-bolt-input:focus {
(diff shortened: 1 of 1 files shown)


mikemai2awesome (approved):
@danielamorse Let's roll with this. I don't see another way.

PR #2356, opened 2021-10-15
DS-590 Switch button componentv1

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-590

## Summary

Adds a new component, Switch Button, which handles boolean UI functions that happens instantly.

## Details

1. Added new component, Jest test and docs.
2. Only bare minimum props are created to handle basic needs for the notifications project. We can expand on this component down the road.

## How to test

Run the branch locally and proof read the Switch Button docs. Make sure the docs are intuitive. Inspect the demo and change on/off to some other language that might be longer, make sure it looks ok. Then check the component code for any mistakes in the code.


danielamorse (commented):
Looks awesome! Very slick.

I made some minor copy edits.

Do we want to add the following to the schema?:
1. `checked`: make it checked/unchecked by default
2. `uuid`: allow user to pass their own id rather than have it randomly generated by Twig? It would still default to random Twig id. This is how we do it in [Popover](https://github.com/boltdesignsystem/bolt/blob/master/packages/components/bolt-popover/src/popover.twig#L9).

Both of these could be a round 2 additions if we want to ship this now.

mikemai2awesome:
@danielamorse Good suggestions, I would named the prop `on` rather than `checked`, I've seen this naming in UISwtich for iOS.

colbytcook:
@mikemai2awesome @danielamorse this PR got weird (it spanned 3 releases at this point) and I don't trust the updates to it. I am going to close this PR since I cherry picked the commits and created a new PR, https://github.com/boltdesignsystem/bolt/pull/2382.

PR #2379, opened 2021-11-10
DS 656 Form Documentation Update

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-656

## Summary

Form Component was updated with the standard of the new docs.

## Details

- form inputs variations were updated with the new docs
- places with old form component usage were replaced with the new usage
- old docs were removed
- only `twig` docs examples were created for now

## How to test

Pull the branch. Check if the documentation is useful and there are no regressions.


mikemai2awesome (changes requested):
I updated the copy on a few pages, but I did not go through them all. I would like to tune up the rest when I come back.

mikemai2awesome (approved):
I finished updating the docs.

colbytcook (approved):
Approved, pushed up some typo fixes

PR #2380, opened 2021-11-10
Remove unwanted icon files

danielamorse:
## Jira

n/a

## Summary

Remove unwanted icon files that were accidentally added in a merge.

## Details

[These files](https://github.com/boltdesignsystem/bolt/tree/master/packages/elements/bolt-icon/src/svgs) were reverted upstream and then when we [pulled the latest](https://github.com/boltdesignsystem/bolt/commit/3b45c19e3256c03c2881f46b34dd68d90a29a00d) into a branch where these files were moved, removed, or renamed (@mikemai2awesome can you please confirm?) they were added back in. I don't think we need these files, or else only the SVG needs to be added to `packages/elements/bolt-icon/src/icons/svgs`.

## How to test

Verify that these files can be removed or move only the `minus-*.svg` files to `packages/elements/bolt-icon/src/icons/svgs` and run the build to generate the corresponding Twig files.


mikemai2awesome (changes requested):
Keep these icons, do not merge this.

PR #2385, opened 2021-11-17
Background Component Image Update

colbytcook:
## Summary

Updated the Background Component to use the Image Element instead of the Image Component

## Details

In order to allow for a more performant Hero, we need to remove all the Image Component. On of the steps to do that is to replace the Image Element from the Background Component so we can leverage more perforrmant methods (for example, appending `loading="eager"` to all images in a hero).

## How to test

* Review the code and pull down the branch.
* Review the Bolt Background, and the Bolt Band docs.
* Review the docs "Pages" folder

## Release notes

The Background Component now leverages the Image Element instead of the Image Component. This is a step to create a more performant Hero

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
@@ -12,8 +12,8 @@
         pattern: 'image',
         src: '/images/content/backgrounds/background-tall-4.jpg',
         srcset: '/images/content/backgrounds/background-tall-4-50.jpg 50w, /images/content/backgrounds/background-tall-4-100.jpg 100w, /images/content/backgrounds/background-tall-4-200.jpg 200w, /images/content/backgrounds/background-tall-4-320.jpg 320w, /images/content/backgrounds/background-tall-4-480.jpg 480w, /images/content/backgrounds/background-tall-4-640.jpg 640w, /images/content/backgrounds/background-tall-4-800.jpg 800w, /images/content/backgrounds/background-tall-4-1024.jpg 1024w, /images/content/backgrounds/background-tall-4-1366.jpg 1366w, /images/content/backgrounds/background-tall-4-1536.jpg 1536w, /images/content/backgrounds/background-tall-4-1920.jpg 1920w, /images/content/backgrounds/background-tall-4-2560.jpg 2560w, /images/content/backgrounds/background-tall-4-2880.jpg 2880w, /images/content/backgrounds/background-tall-4.jpg 3840w',
-        width: 823,
-        height: 400
+        width: 3840,
+        height: 2160
       }
     ]
 } only %}
--- a/docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
@@ -10,7 +10,10 @@
     content_items: [
       {
         pattern: 'image',
-        src: '/images/content/backgrounds/background-tall-4.jpg'
+        src: '/images/content/backgrounds/background-tall-4.jpg',
+        srcset: '/images/content/backgrounds/background-tall-4-50.jpg 50w, /images/content/backgrounds/background-tall-4-100.jpg 100w, /images/content/backgrounds/background-tall-4-200.jpg 200w, /images/content/backgrounds/background-tall-4-320.jpg 320w, /images/content/backgrounds/background-tall-4-480.jpg 480w, /images/content/backgrounds/background-tall-4-640.jpg 640w, /images/content/backgrounds/background-tall-4-800.jpg 800w, /images/content/backgrounds/background-tall-4-1024.jpg 1024w, /images/content/backgrounds/background-tall-4-1366.jpg 1366w, /images/content/backgrounds/background-tall-4-1536.jpg 1536w, /images/content/backgrounds/background-tall-4-1920.jpg 1920w, /images/content/backgrounds/background-tall-4-2560.jpg 2560w, /images/content/backgrounds/background-tall-4-2880.jpg 2880w, /images/content/backgrounds/background-tall-4.jpg 3840w',
+        width: 823,
+        height: 400
       }
     ]
 } only %}
--- a/packages/components/bolt-background/src/background-item.twig
+++ b/packages/components/bolt-background/src/background-item.twig
@@ -21,27 +21,39 @@
 {% if item.pattern %}
   <div class="{{ 'c-bolt-background__item' }}">
     {% if item.pattern == 'image' %}
-      {% set _srcset = item.srcset ? item.srcset : '' %}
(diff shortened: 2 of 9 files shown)


packages/components/bolt-background/src/background-item.twig
```diff
@@ -20,11 +20,36 @@
...
+      {% include '@bolt-elements-image/image.twig' with {
+        background: true,
+        attributes: {
+          src: item.src,
+          srcset: _srcset,
+          sizes: _sizes,
+          width: _width,
+          height: _height,
+          loading: _loading,
+          style: 'object-position: ' ~ _alignment ~ ';'
```
mikemai2awesome, line 43: Use the CSS var we set up rather than this.

docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
```diff
@@ -10,7 +10,10 @@
     content_items: [
       {
         pattern: 'image',
-        src: '/images/content/backgrounds/background-tall-4.jpg'
+        src: '/images/content/backgrounds/background-tall-4.jpg',
+        srcset: '/images/content/backgrounds/background-tall-4-50.jpg 50w, /images/content/backgrounds/background-tall-4-100.jpg 100w, /images/content/backgrounds/background-tall-4-200.jpg 200w, /images/content/backgrounds/background-tall-4-320.jpg 320w, /images/content/backgrounds/background-tall-4-480.jpg 480w, /images/content/backgrounds/background-tall-4-640.jpg 640w, /images/content/backgrounds/background-tall-4-800.jpg 800w, /images/content/backgrounds/background-tall-4-1024.jpg 1024w, /images/content/backgrounds/background-tall-4-1366.jpg 1366w, /images/content/backgrounds/background-tall-4-1536.jpg 1536w, /images/content/backgrounds/background-tall-4-1920.jpg 1920w, /images/content/backgrounds/background-tall-4-2560.jpg 2560w, /images/content/backgrounds/background-tall-4-2880.jpg 2880w, /images/content/backgrounds/background-tall-4.jpg 3840w',
+        width: 823,
```
mikemai2awesome, line 15: Put in the source image width and height for all of these demos.

docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
```diff
@@ -10,7 +10,10 @@
...
+        height: 400
       }
```
mikemai2awesome, line 17: Add `sizes: '100vw'` for all of these demos because they all take up full width of the viewport.

mikemai2awesome (changes requested):

docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
danielamorse: Also noticed that the Band demo has empty `width` and `height` attributes:
image

packages/components/bolt-background/src/background-item.twig
```diff
@@ -20,11 +20,36 @@
...
   <div class="{{ 'c-bolt-background__item' }}">
-    {% include pattern_template(item.pattern) with item|merge({
-      ratio: _ratio_value,
-      lazyload: item.lazyload ?? true,
-      cover: item.cover ?? true
-    }) only %}
+    {% if item.pattern == 'image' %}
+      {% set _srcset = item.srcset ? item.srcset : '' %}
+      {% set _width = item.width ? item.width : '' %}
+      {% set _height = item.height ? item.height : '' %}
```
danielamorse, line 26: Rather than setting these to empty strings, which will output empty attributes, you can do something like this:
```
{# conditional attribute #}
{% if item.srcset %}
  {% set attributes = attributes|merge({ src: item.srcset }) %}
{% endif %}

{% include '@bolt-elements-image/image.twig' with {
  background: true,
  attributes: attributes|merge({
    src: item.src, // unconditional attribute
  })
  ...
```

danielamorse (changes requested):

docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
colbytcook: @mikemai2awesome I added this as the default in the background-item.twig on line 27

packages/components/bolt-background/src/background-item.twig
colbytcook: This is all set

docs-site/src/pages/pattern-lab/_patterns/40-components/background/05-background.twig
colbytcook: width and height values are updated and will no longer show if there are now values

packages/components/bolt-background/src/background-item.twig
colbytcook: empty attributes will no longer show if there are now values

danielamorse (approved):

packages/components/bolt-background/src/background-item.twig
danielamorse: Perfect

mikemai2awesome (approved):

packages/components/bolt-background/src/background-item.twig
```diff
@@ -20,11 +20,48 @@
 {# todo: in v4.0 nix the pattern_template pattern #}
 {% if item.pattern %}
   <div class="{{ 'c-bolt-background__item' }}">
-    {% include pattern_template(item.pattern) with item|merge({
-      ratio: _ratio_value,
-      lazyload: item.lazyload ?? true,
-      cover: item.cover ?? true
-    }) only %}
+    {% if item.pattern == 'image' %}
```
remydenton, line 23: This is nicely backwards-compatible, but what's the plan going forward?  As it stands now, you _have_ to pass data as expected by the deprecated image component.  There's no way to do it "right" even if you know better.  Annoying at best, but this may come back to bit us when we try to replace other image components with image elements in Drupal.

Here's a proposal-- what if within `if item.pattern == image`, we have a second check: if there's either a `src` or `srcset` prop, then we assume that the deprecated image component pattern is being used (since those props exist in image component but not image element).  However, if neither of those props is present, we assume the user is actually passing data that's intended to be rendered with the image element.

Thoughts?  We can handle this in a follow up PR.

remydenton: Following up on this, I realized we already have this logic at the bottom of this control statement:
```
{% elseif (item is iterable) %}
  <div class="{{ 'c-bolt-background__item' }}">
    {{ item }}
  </div>
{% else %}
  {{ item }}
{% endif %}
```

That allows you to just pass an already rendered (or renderable) item without resorting to the horrible `pattern_template()` design pattern.  That is the preferred option for anyone who wants to use this pattern correctly.  Our schema docs need some updates, but otherwise this is fine as-is.

PR #2388, opened 2021-11-19
DS-669 side nav and holy grail layout enhancements

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-669

## Summary

Fixes the triangle symbol in Side Nav links to render better in Windows and Holy Grail sidebar to have a max-width.

## Details

1. Used a different HTML entity to render the triangle in side nav so it appears normal in Windows
2. Set a max-width on the holy grail sidebar so any long side nav link text would not make the sidebar so wide that it squishes the main content

## How to test

Run the branch locally and check the Community Doc Layout page in Windows. Make sure the triangle in the side nav looks similar to what you see in Mac. Also the side nav at any breakpoint is at a reasonable width that doesn't squish the main article content too hard.


colbytcook (approved):
Tested on Firefox on Windows and can confirm the arrow is not squished anymore and the content area is not squished on resize.

danielamorse (approved):
Tested on Chrome and Windows on Browserstack, works perfectly.

PR #2389, opened 2021-11-22
DS 677 notifications tag manager

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-677

## Summary

Notification Tag Manager modal mockup was created.

## Details

A mockup of four separate steps was created.

## How to test

Check if the Notification Tag Manager is similar to the design and if the standards of a11y and UX are acceptable

### Visual changes

The design of the four steps (white circles on the top right corner) was replaced with our progress bar component. Bolt-chip style was changed a little bit.


mikemai2awesome (changes requested):
Hi @MarcinMr, this is a good start. Let's take care of the following.

## General

1. We can move all mockups under User now. I've created one for you to follow: `05-user-notification-settings.twig`.
2. I created a new template called `_site-layout-template.twig`, this uses the Site Layout, Page Header, and Page Footer. This should be the standard going forward.
3. Remove all bands as we are trying to kill them. Layout should give you what you need. Reference how I did it in `05-user-notification-settings.twig`.
4. When writing custom code for this, use `c-base-` prefix, this will indicate the custom code is coming from the base theme. In your case, you can do `c-base-tag-manager-*`, do not abbreviate if it makes things hard to understand.
5. Render modals at the bottom of the page rather than nested them inside any elements.
6. Use all [current theme classes](https://boltdesignsystem.com/pattern-lab/?p=visual-styles-theming-system) when setting themes, do not use deprecated ones.
7. Do not combine a `c-this` with a `c-that`, components should not be merged together.

## Notification Settings Page

1. All the `c-ntm` custom code are unnecessary, remove them and use the grid component and listing teaser component instead.
2. Each listing links out to their respective page so make sure the heading is linked when you are using the listing teaser to build out the list.

## Tag Manager Modals

1. Set the headline one size smaller (xxlarge) and set it to `h2`.
2. Set the modal width to `regular` and scroll to `overall` (already done).
3. Replace switch buttons with regular checkbox. I discussed with other designers that checkboxes are more appropriate in a form context.
4. Remove as much custom code as possible. Grid, List, and Stack components, and [Spacing Util classes](https://boltdesignsystem.com/pattern-lab/?p=visual-styles-spacing-utilities) should take care of general layouts. Only write custom code if those don't give you what you need.

MarcinMr:
@mikemai2awesome 

I deleted all custom CSS for the Notification Settings page and used bolt components like: `listing-teaser`, `grid`, `stack`, and used some of the utility classes to recreate that design.

For the modals followed the same approach. The only custom code I left is for `header` (as we discussed on the slack that this could be the only element with the custom code). I'm wondering if this should be left (custom CSS) or I should remove the custom CSS and rewrite the `header` using the inline `bolt-list` component. Now the `header` markup is cleaner but we have a couple of lines of custom CSS. And using the inline `bolt-list` component will cause a little bit more of the markup but no CSS. 

What would be better ? Let me know what do you think @danielamorse and I'll fix if neccessary

danielamorse:
@MarcinMr I would say remove the remaining custom CSS and finish building the header with existing components (and maybe inline style or CSS vars for the background gradient?).

The main reason I say this is because you've already converted so much of this to use existing components and utility classes. Normally if we have an "element" class like `c-base-tag-manager__header`, we'd also have a "block" class wrapping it like `c-base-tag-manager`. In a component like this we'd also have `__body` and `__footer` classes as well, but since you did such a good job of converting to existing components :) I'd say just go all the way.

Otherwise, looks great visually and in code. Take care of this last step and we can merge it.

danielamorse (changes requested):
Requested changes: https://github.com/boltdesignsystem/bolt/pull/2389#issuecomment-981824782

danielamorse (approved):
@MarcinMr good question about whether or not to include the gradient. I went ahead and added it as an inline style so that we have it if we need it. Before removing, I'd want to run it by Matt (or whoever did the original designs) and since Piotr will start theming the app soon I want to hand this off ASAP. It's easy enough to remove or to add to a separate stylesheet on the Drupal-side once we hand it off.

Thank you! I'm going to start working with these template today in the React app.

PR #2390, opened 2021-11-22
DS-304 add subheadline sizes to match Drupal admin ui

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-304

## Summary

Fixes a mismatch between the headline component and Drupal admin ui's authoring options.

## Details

1. Enabled subheadline to use any size and tag options (although this is against the design guidelines, subheadline should only be used at large, xlarge, and xxlarge sizes, and its tag can be just headings or paragraph).
2. Fixed a doc error where the size variable is not rendering on the numbered headline docs.

We are making this change rather than fixing the Drupal implementation because it's safer. Efforts will be put in to refactor foundational Drupal paragraphs (including headline) in the future, and the headline component will need a refactor (downgrade to element) on the design system side as well.

## How to test

Run the branch locally and check the headline docs. Nothing should be visually different compared to master.

### DIFF
--- a/packages/components/bolt-headline/src/subheadline.twig
+++ b/packages/components/bolt-headline/src/subheadline.twig
@@ -1,19 +1,6 @@
-{% extends "@bolt/_typography.twig" %}
-{% set type = "subheadline" %}
-{# @todo @Salem, not sure if span should be allowed or not. There should not be two overrides of tags here. See headline as well #}
-{% set tags = ["h1", "h2", "h3", "h4", "h5", "h6", "p", "span"] %} {# Allowed Tags #}
-{% set tags = [
-  "h1",
-  "h2",
-  "h3",
-  "h4",
-  "h5",
-  "h6",
-  "p",
-] %}
-{% set sizes = ["xxlarge", "xlarge", "large"] %}
-{% set size = size | default("xlarge") %}
-{% set weight = "regular" %}
-
-{% set styles = ["normal", "italic"] %}
-{% set style = style | default("normal") %}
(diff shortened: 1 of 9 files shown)


colbytcook (approved):

packages/components/bolt-headline/src/text.twig
```diff
@@ -1,8 +1,6 @@
-{% extends "@bolt/_typography.twig" %}
-{% set type = "text" %}
-{% set tag = tag | default("p") %}
-{% set weights = ["regular", "bold", "semibold"] %}
-{% set styles = ["normal", "italic"] %}
```
remydenton, line 5: Do `weights` and `styles` even have any effect when used like this?  i.e. if they're defined in `_typography.twig`, I would think that those values would take preference anyway-- but I guess you're removing them in this PR because they previously did limit the possible options?

packages/components/bolt-headline/src/subheadline.twig
```diff
@@ -1,19 +1,6 @@
...
-{% set tags = ["h1", "h2", "h3", "h4", "h5", "h6", "p", "span"] %} {# Allowed Tags #}
-{% set tags = [
-  "h1",
-  "h2",
-  "h3",
-  "h4",
-  "h5",
-  "h6",
-  "p",
-] %}
```
remydenton, line 13: Jesus 🤦 

Thanks for cleaning this up.

remydenton (approved):

PR #2391, opened 2021-11-22
DS-676 Upgrade to Stylelint 14

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-676](https://pegadigitalit.atlassian.net/browse/DS-676)

## Summary

stylelint was a few versions and several breaking changes behind, so we wanted to update it.

## Details

I updated stylelint, then modified its config to prevent any modification of our rules. In the future we should take another look at the index file that i changed here and see what rules we should be applying.

## How to test

Pull the branch, run `yarn clean && yarn setup` then `yarn lint:scss`.


danielamorse (approved):
Look great, nice work 👏 

Incidentally, reviewing this PR helped me figure out that I wasn't actually using v14 in my React app project. I was running `npx stylelint` which was using my globally installed v12.

Another thing to consider... this config will be used outside of Bolt in Drupal. We can release it now in a minor release, but in the next major release, when Drupal upgrades build tools to 6.x, there will likely be SCSS to fix (Note to self and @remydenton @colbytcook).

PR #2399, opened 2021-12-07
DS-682 profile enhancements

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-682

## Summary

Design enhancements to Profile and Info Section components

## Details

1. Updated the size of avatar and cover in Profile
2. Updated heading weight for Info Section
3. Updated user profile demos

## How to test

Run the branch locally and compare with master. In the User Profile page, the avatar and cover should be smaller, the info section headings should be bold, the certificates should show 4 per row on desktop and 2 per row on mobile, and the badges should show 6 per row on desktop and 3 per row on mobile.


colbytcook (approved):
Can confirm that the avatar and banner reduced, the headlines are bold and the responsive behavior is correct in FF, Chrome, Edge and Safari.

PR #2400, opened 2021-12-08
DS-660 Ratio Element

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-660

## Summary

Ratio Element is created to replace the Ratio Component in the future.

## Details

A new Ratio Element was created and its docs

- schema
- twig template
- scss
- docs

## How to test

Pull the branch. Check if:

- the `schema` is compatible with our standards
- `styles` are efficient and sufficient for this element
- `ratio element` meets use cases and can replace the old `ratio component`
- `docs` are useful and readable

## Release notes

`Ratio Component` will be deprecated and replaced with the new `Ratio Element`.


mikemai2awesome (changes requested):
Can we reduce to just one container?

```
<div class="e-bolt-ratio" style="--e-bolt-aspect-ratio: 16/9">
  /* image or video goes here */
</div>
```

Also, CSS aspect ratio is only an experimental feature in Safari, so we cannot merge this just yet.

MarcinMr:
> Can we reduce to just one container?
> 
> ```
> <div class="e-bolt-ratio" style="--e-bolt-aspect-ratio: 16/9">
>   /* image or video goes here */
> </div>
> ```
> 
> Also, CSS aspect ratio is only an experimental feature in Safari, so we cannot merge this just yet.

@mikemai2awesome 

Yes, we could go like this:

(A)
bolt-ratio-1-div

but since we use here an `inline-style` with the CSS variable, when someone uses `attributes object`, for example: `attributes: { style: 'max-width: 150px' }` <- this will override our `inline-style` from the template with the CSS aspect-ratio variable.

So if we are sure that using ratio-element, `attributes object` with a `style` property inside the ratio element include, won't be used, we can utilize only a single `div` without nesting anything inside it, like in the screen above. This is why I nested one container in another - to make it possible using `attributes object` with a `style` property.

But we can also go like this:

(B)
bolt-ratio-merge-classes

this way we can use both our inline variable and styles from the `attributes object` if necessary. The thing is we don't have anything like: `attributes.addStyles(styles)` like the method for adding classes `attributes.addClass(classes)` which would be useful in this case.

Option B is now pushed. Let me know which is better


When it comes to Safari IOS usage, I provided `@supports` for that case. Isn't that enough to make it work on Safari IOS?

mikemai2awesome:
@MarcinMr that's great! Let's bring it up in dev huddle. @danielamorse and @remydenton would be better to think about the attributes stuff.

remydenton:
Good questions guys.  I think one container is definitely preferable, and we just need to manage the attributes on it.

Option A from @MarcinMr is close to what we want, but you're right we don't want to prevent additional style attributes from being passed.

I think this is a good case where the [|without() filter](https://www.drupal.org/docs/8/theming-drupal-8/using-attributes-in-templates#s-using-without-filter) for attributes would come in handy.  There's almost never a good reason to use this with classes (use addClass instead), but since there's no `addStyle()` method as you mention, I think this is the easiest way to manage the style attribute.  Something like this:

```
<div {{ attributes.addClass('classes')|without('style') }}
  style="--e-bolt-aspect-ratio: {{ ratio }}; {{ attributes.style }}"
>
  {{ children }}
</div>
```

MarcinMr:
> Good questions guys. I think one container is definitely preferable, and we just need to manage the attributes on it.
> 
> Option A from @MarcinMr is close to what we want, but you're right we don't want to prevent additional style attributes from being passed.
> 
> I think this is a good case where the [|without() filter](https://www.drupal.org/docs/8/theming-drupal-8/using-attributes-in-templates#s-using-without-filter) for attributes would come in handy. There's almost never a good reason to use this with classes (use addClass instead), but since there's no `addStyle()` method as you mention, I think this is the easiest way to manage the style attribute. Something like this:
> 
> ```
> <div {{ attributes.addClass('classes')|without('style') }}
>   style="--e-bolt-aspect-ratio: {{ ratio }}; {{ attributes.style }}"
> >
>   {{ children }}
> </div>
> ```

Thank you Remy :)

Your suggestion worked very well. But the approach changed a little bit and we will just add 3 common aspect-ratio from the `enum` and style them by modifier classes, and form more advanced options like choosing any `aspect-ratio` value we will use `attributes style`

mikemai2awesome (approved):
@MarcinMr As we discussed in dev huddle, we decided to remove the doc page about nearest aspect ratio since it might create confusion rather than providing useful tips.

I also added a doc page demonstrating image fit and position within ratio, those are expected to work. This was brought up in huddle by @adamszalapski .

MarcinMr:
@mikemai2awesome thanks for the updates. Solved conflict related to a cache

colbytcook (approved):
Approved. One note to make on using this in prod. Safari 15 for both OS and iOS support this feature but unfortunately Safari Mobile still has a significant user base (9.62% global usage). It is only a matter of time until this number drops but until then we probably shouldn't be using this element on production (yet). If we see the user base drop under 2% globally then we should be good to update the docs and components to leverage this (I will create tickets to deprecate the ratio component and update the docs and components with the ratio element).

danielamorse (approved):
Code looks good and got a visual demo last week.

PR #2402, opened 2021-12-09
DOCS-232 Fix holy grail layout content width

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DOCS-232

## Summary

Fixes a grid width issue with the Holy Grail Layout that causes responsive tables to not render correctly.

## Details

Use `1fr` in CSS grid template is causing certain components miscalculate the width of its parent container. It has been changed to `minmax(0, 1fr)` as a fix. 0 min-width straightens out the issue.

## How to test

Run the branch locally and check the Holy Grail Layout docs. The table in the use case demo should be able to horizontally scroll when the viewport is smaller than ~1000px.


colbytcook (approved):
Approved, confirmed using all the Mac browsers, Firefox, Edge, Chrome, Safari

danielamorse (approved):
@mikemai2awesome looks good, works as expected.

Should we add a comment here about why we use `minmax()` instead of just `1fr`?

PR #2406, opened 2021-12-16
Version Selector Logic Update

colbytcook:
## Summary

Updated the release list that will be displayed in the select version drop down (no "beta", "rc"...) and updated the drop down list so it only shows the current and previous releases and not one in the future

## Details

* Updated the logic on how the `bolt-releases.bolt.json` is being populated. Now anytime there is a word present the tag will be ignored. This is to prevent `beta`, `rc`, `test`... versions from showing up on the boltdesignsystem.com version selector.
* Updated the logic in which the version selector's options are not being populated. It will now only show the current and previous tags based on the `leran.json` version value.

## How to test

* Pull down the branch.
* Run `yarn build` and open up the `docs-site/.incache` file
* Search for "beta", you should have no results
* Open up the `lerna.json` file and change the `version` value to `3.5.4`
* Run `yarn build` and open up the `www/build/data/bolt-releases.bolt.json` file
* Confirm the first item in the `options` array is the `3.5.4`object
* Run `yarn start`
* On the local instance, observe that the version selector now defaults to `3.5.4` and the version selector drop down will only show `3.5.4` and later.
* On the local instance, observe that the version selector options do not have any instances of `beta`


danielamorse (changes requested):
@colbytcook works perfectly, but I have suggestions on some of the JS. See https://github.com/boltdesignsystem/bolt/pull/2408.

danielamorse (approved):

PR #2407, opened 2021-12-17
DS-671 Add rel="noopener"

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-671

## Summary

The attribute `rel="noopener"` was added to the `twig` demos and` js` files

## Details

The attribute `rel="noopener"` was added to:

- twig includes
- link demos
- js files
- test files

 :small_orange_diamond: `Bolt-trigger` and `bolt-chip` were updated due to adding `rel` prop to the schema and then updating the templates and demos. When we check the `bolt-trigger` and `bolt-chip` schemas we can see that they don't use `object attributes` to provide a `target` attribute but there is a custom prop target - so I follow the same way for the `rel` attr.

## How to test

Pull the branch. Check the codebase if `rel="noopener"` is added where should be

### DIFF
--- a/packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
+++ b/packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
@@ -81,9 +81,7 @@ class BoltCardReplacementAction extends withContext(BoltElement) {
                 target="${ifDefined(
                   this.url ? (this.external ? '_blank' : '_self') : undefined,
                 )}"
-                rel="${ifDefined(
-                  this.url ? (this.external ? 'noopener' : '') : undefined,
-                )}"
+                ${this.external ? 'rel="noopener"' : undefined}
               >
                 ${this.slotify('default')}${renderedIcon
                   ? // prettier-ignore
--- a/packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
+++ b/packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
@@ -81,6 +81,9 @@ class BoltCardReplacementAction extends withContext(BoltElement) {
                 target="${ifDefined(
                   this.url ? (this.external ? '_blank' : '_self') : undefined,
                 )}"
+                rel="${ifDefined(
+                  this.url ? (this.external ? 'noopener' : '') : undefined,
+                )}"
               >
                 ${this.slotify('default')}${renderedIcon
                   ? // prettier-ignore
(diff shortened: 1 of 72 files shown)


mikemai2awesome (approved):

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
```diff
@@ -81,6 +81,9 @@ class BoltCardReplacementAction extends withContext(BoltElement) {
                 target="${ifDefined(
                   this.url ? (this.external ? '_blank' : '_self') : undefined,
                 )}"
+                rel="${ifDefined(
+                  this.url ? (this.external ? 'noopener' : '') : undefined,
+                )}"
```
colbytcook, line 86: Shouldn't the attribute ("rel") be the conditional here? In the current iteration, the "rel" will show on every card action without a value which isn't ideal. We would want the conditional to control the display of the attribute and not just the value.

colbytcook:
There is a couple of examples within the components where `target="_blank"` is being applied but the `rel="noopener"` is not being added. The following is a list of examples where we might have to adjust the `target` attribute logic to include the `rel`,
* Chip Components
* Pega Footer Nav LI
* Teaser
* Trigger

colbytcook (changes requested):
I think we need to address some of the remaining component's target logic.

MarcinMr:
> @MarcinMr There is a couple of examples within the components where `target="_blank"` is being applied but the `rel="noopener"` is not being added. The following is a list of examples where we might have to adjust the `target` attribute logic to include the `rel`,
> 
> * Chip Components
> * Pega Footer Nav LI
> * Teaser
> * Trigger

@colbytcook 

For Pega Nav Li `rel` attr can be set using the 'link` prop. You can check that in the schema.

`Bolt-trigger` and `bolt-chip` were updated due to adding `rel` prop to the schema and then updating the templates and demos. When we check the `bolt-trigger` and `bolt-chip` schemas we can see that they don't use `object attributes` to provide a `target` attribute but there is a custom prop target - so I follow the same way for the `rel` attr.

packages/components/bolt-card-replacement/src/card-replacement-actions/card-replacement-action.js
MarcinMr: First of all, here I fixed the [card-replacement twig template](https://github.com/boltdesignsystem/bolt/blob/2507bf2340a10e2d211fcb072e896dfcb9154174/packages/components/bolt-card-replacement/src/card-replacement-actions/_card-replacement-action.twig). Because the `rel` attr showed without a value due to wrong if statement.

colbytcook (approved):

PR #2408, opened 2021-12-17
JS refactoring

danielamorse:
## Jira

n/a

## Summary

Refactors some JS in #2406.

## Details

- Remove `semver-compare` dependency. The `semver` package we already use has a function for comparing.
- Get `lerna.json` using `process.cwd()` rather than relative path from package.
- Refactor a couple JS loops to use ES6 methods that are shorter and do the same thing.
- Use semver validation instead of RegEx to filter out tags. Regex works just fine here, but I thought the semver validation would be a more explicit check for what we're looking for, and I prefer to use RegEx only as a last resort.


packages/build-tools/tasks/api-tasks/bolt-versions.js
```diff
@@ -8,51 +8,43 @@ const InCache = require('incache');
...
 const { throttling } = require('@octokit/plugin-throttling');
 const MyOctokit = Octokit.plugin(throttling);
-const cmp = require('semver-compare');
+
 let versionSpinner;
 
 const { getConfig } = require('@bolt/build-utils/config-store');
 const { fileExists } = require('@bolt/build-utils/general');
-const lernaConfig = require('./../../../../lerna.json');
+const lernaConfig = require(path.join(process.cwd(), '../lerna.json'));
```
colbytcook, line 16: @danielamorse I should I mentioned this but the `path.join` fails depending on what command is run. the `../` works on `yarn start` but fails on `yarn build` (I think) due the directory changing to be the parent folder of the Bolt repository. We might want to turn this back the ugly relative link.

danielamorse: Makes sense. I reverted my change.

colbytcook (approved):

PR #2413, opened 2021-12-27
Hotfix: v5.1.1 fix build errors

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-688

## Summary

Fix build errors found when pulling in 5.x to Drupal.

### Lerna
We were requiring a script that assumed the `lerna.json` file was accessible. This is a safe assumption in the monorepo but not in the Drupal environment. The solution is to skip this script in Drupal.

### Stylelint
A few of our stylelint-related packages are not compatible with stylelint < v14, and one of the packages has a missing peer dependency (`postcss`). I upgraded the packages and added the expected peer dependency.

The takeaway here is to pay attention to the peer dependency warnings we see when running `yarn install`. Mismatches here at some point became actual bugs.

## How to test

- Check out this branch and run `yarn`, `yarn lint:scss`, `yarn fix`, `yarn start`, and `yarn build`.
- Travis is passing.

### DIFF
--- a/packages/build-tools/tasks/task-collections.js
+++ b/packages/build-tools/tasks/task-collections.js
@@ -11,7 +11,6 @@ const imageTasks = require('./image-tasks');
 const iconComponentTasks = require('./icon-component-tasks');
 const iconTasks = require('./icon-tasks');
 
-const { writeBoltVersions } = require('./api-tasks/bolt-versions');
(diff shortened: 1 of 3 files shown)


remydenton (approved):
This looks great, thanks for the clear explanations @danielamorse

PR #2414, opened 2022-01-06
WWW-1157: iOS modal bug

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/WWW-1157
https://pegadigitalit.atlassian.net/browse/DS-448

## Summary

Added a new utility class to help solve a iOS scroll bug with the modal, updated the modal CSS for more consistent behavior

## Details

* Added a utility CSS class that will apply a color background directly behind the modal and fix a regression bug with table modals
* Adjusted the modal styling to provide more consistent behavior on iOS Safari

## How to test
Pull down the branch and complete the following tasks.

*Non-iOS Devices*
1. Review the Patternlab modal examples for any regressions.
1. Confirm that on Chrome, Edge, Safari (on mac and not iOS), and Firefox, the "u-bolt-ios-content-fill" is NOT being added to the body (I have confirmed this on Mac but not on Windows though I cannot see that there would be a difference).

*iOS Devices (either an iPhone or an iPhone emulator)*
1. On a iOS device, review the Patternlab modal examples and confirm that the class "u-bolt-ios-content-fill" is being added to the body after a modal is opened.
2. Navigate to the [BoC Gallery Page](http://localhost:3000/pattern-lab/?p=pages-boc-gallery-phase2) and confirm that when the iOS URL bar hides (this requires scrolling up), the modal remains the same size and a navy color is filling the space below the modal (this behavior can be found on the [Carbon Design System Modal](https://www.carbondesignsystem.com/components/modal/code)). This is to check the [DS-448](https://pegadigitalit.atlassian.net/browse/DS-448) regression.
3. Navigate to [Next in Tech: 2022 Forecast](https://staging-www.pega.com/webinars/next-tech-2022-forecast-dec) and click on the "Register Now" button
4. Inspect the modal and update the CSS with the [modal.scss](https://github.com/boltdesignsystem/bolt/pull/2414/files#diff-e7b7f8255942c28b9ab1c4f8d821e14fee92504fb5f638274187489b8ed97a19) changes.
5. Add a ::before element to the body and update this new element with the CSS found in the [_utilities-ios-content-fill.scss](https://github.com/boltdesignsystem/bolt/pull/2414/files#diff-f5347f123f40bc623a4720ede28b92148b584d0ecb74ddf37fa81f8757dbf673) file.
6. Scroll so the iOS URL bar is hidden and observe that you can no longer see the page content.  This is to confirm that this fixes the bug reported in [WWW-1157](https://pegadigitalit.atlassian.net/browse/WWW-1157).

### Visual changes

On iOS Safari there will now be a navy blue background behind the modal, this can only be seen by having thee URL bar at the bottom of the screen and scrolling up (the URL bar should disappear off the bottom of the screen). If the URL bar hides these will be a navy blue color between the bottom of the modal and the bottom of thee view port.


packages/global/styles/07-utilities/_utilities-ios-content-fill.scss
```diff
@@ -0,0 +1,46 @@
...
+        top: 0;
+        right: 0;
+        bottom: 0;
+        left: 0;
+        z-index: bolt-z-index('modal') - 1; // [3]
+        height: 150vh; // [4]
+        background-color: var(--bolt-color-navy);
+      }
+
+      .c-bolt-table {
```
mikemai2awesome, line 37: This is very suspicious and I don't understand the comment, and it's not stated in the tickets neither. Again, do we actually have a problem with table since we didn't change to use css grid for modal?

packages/components/bolt-modal/src/modal.scss
```diff
@@ -36,10 +36,14 @@ bolt-modal:not([ready]) {
   opacity: bolt-opacity(0);
   position: fixed;
   top: 0;
+  right: 0;
+  bottom: 0;
   left: 0;
   width: 100%; // Use % instead of vh or modal scrollbar is hidden behind bold scrollbar in IE11
   height: 100vh;
```
mikemai2awesome, line 43: Wanna double check one more time, I think ios safari has problem with `vh` specifically. Did you try `height: 100%` instead?

packages/global/styles/07-utilities/_utilities-ios-content-fill.scss
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+// Dev Notes
+// 1. Make this only show on smaller devices to align with the modal display change (when it starts to cover the screen).
+// 2. Target only Safari with this fix (this is to help prevent regressions).
+// 3. Place this background directly behind the modal to prevent other elements from displaying over it.
+// 4. Forces the background color to spill out the bottom of the page to cover the browser scrrolling/painting issues.
+// 5. This fixes a table specific regression that didn't allow the model to show over the background color.
+
+.u-bolt-ios-content-fill {
```
mikemai2awesome, line 19: This is way too much for a utility class and very specific to modal. I prefer if it's tied to modal instead. This might be an example for data attribute. @danielamorse

mikemai2awesome (changes requested):

packages/global/styles/07-utilities/_utilities-ios-content-fill.scss
```diff
@@ -0,0 +1,46 @@
...
+      // [2]
+      &:before {
+        content: '';
+        display: block;
+        position: fixed;
+        top: 0;
+        right: 0;
+        bottom: 0;
+        left: 0;
+        z-index: bolt-z-index('modal') - 1; // [3]
```
mikemai2awesome, line 32: Please avoid doing math with z-index values from our system. Since modals are rendered at the bottom of `<body>`, they will cover up anything before them even if the same z-index is used, because the modals are higher than the `:before` element in the dom. So i believe this can just be `bolt-z-index('modal')`.

packages/global/styles/07-utilities/_utilities-ios-content-fill.scss
```diff
@@ -0,0 +1,46 @@
...
+    // [1]
+    @supports (-webkit-touch-callout: none) {
+      // [2]
+      &:before {
+        content: '';
+        display: block;
+        position: fixed;
+        top: 0;
+        right: 0;
+        bottom: 0;
```
mikemai2awesome, line 30: `bottom: 0` conflicts with `150vh`, i don't think we need it.

docs-site/src/pages/pattern-lab/_patterns/999-tests/table/00-table-modal.twig
```diff
@@ -0,0 +1,93 @@
...
+{% endset %}
+
+{% set main_content %}
+  <p>This test is to check for regressions regarding the following issues.</p>
+  <ul>
+    <li><a href="https://pegadigitalit.atlassian.net/browse/WWW-1157" target="_blank" rel="noopner">https://pegadigitalit.atlassian.net/browse/WWW-1157</a></li>
+    <li><a href="https://pegadigitalit.atlassian.net/browse/DS-448" target="_blank" rel="noopner">https://pegadigitalit.atlassian.net/browse/DS-448</a></li>
+  </ul>
+
+  <bolt-table>
```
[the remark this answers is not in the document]
colbytcook: @mikemai2awesome I thought is was originally but when I added the `min-height` bit to the CSS it seemed to solve it by itself, so I reverted the `grid` back to soo owe didn't introduce any regressions.

packages/components/bolt-modal/src/modal.js
```diff
@@ -11,6 +11,14 @@ const tabbable = require('tabbable');
 const ESCAPE_KEY = 27;
 let cx = classNames.bind(styles);
 
+// This conditional will target iOS Safari on mobile, if true, it will add the "u-bolt-ios-content-fill" class to the body.
+// Refer to the documentation on "packages/global/styles/07-utilities/_utilities-ios-content-fill.scss" for more details
+const isiOS =
+  window.matchMedia('(max-width: 37.5em)') &&
```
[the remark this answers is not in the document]
colbytcook: I agree, I will update this.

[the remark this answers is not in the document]
colbytcook: @mikemai2awesome When I wrote this logic I did look around in the code base and found only two examples, `bolt-holy-grail` and the `bolt-page-header`, both of which are using hard coded values. 

@danielamorse do you have any insights here? Do we have examples of the JS using any of the values in a break point map (just so I use the existing patterns)? Or even the JS using any existing data maps (I did look but did not find any but maybe I was not looking for the right thing)? If not, I can spend some time to get this to work with the breakpoint map.

packages/global/styles/07-utilities/_utilities-ios-content-fill.scss
colbytcook: @mikemai2awesome this is exactly why I called this one out in the stand up, I had the same feeling that it is very  closely tied to the modal and should live there but it is placed on the body which seems odd. 

@danielamorse what is your opinion on this?

packages/global/styles/07-utilities/_utilities-ios-content-fill.scss
colbytcook: @mikemai2awesome removed, I confirmed we don't need it (it wasn't breaking it but it was redundant). Good catch.

packages/global/styles/07-utilities/_utilities-ios-content-fill.scss
colbytcook: @mikemai2awesome updated

packages/components/bolt-modal/src/modal.scss
colbytcook: @mikemai2awesome double checked again, I think safari has an issue with scrolling and `vh` and not `vh` specifically (`%` will at least attempt to re paint itself when the URL bar hides, `vh` does nothing). In this fix we need the modal to do nothing and remain the same height (both `vh` and `-webkit-fill-available` do this). While `%` does seem to work I would again rather only change CSS if is pertains too the bug fix here, `height: 100%` doesn't really do anything different than `height: 100vh` already does in this case.

mikemai2awesome (approved):
Approving. But feels weird to use a component class on the body, thoughts, @danielamorse?

colbytcook:
@danielamorse I updated this PR based on our conversation, the color overlay is now scoped to the `bolt-modal[open]` attribute, removed the breakpoints from the`modal.js`, the utility class is not removed and refactored the modal docs  to remove all examples of a modal nested inside a table since it is not supported anymore.

@mikemai2awesome please take a look at the update modal docs and let me know if you have any suggestions on improving them (I based the UI on the elements page structure).

docs-site/src/pages/pattern-lab/_patterns/40-components/modal/25-modal-scroll-variations.twig
```diff
@@ -10,49 +8,85 @@
...
+      container_modal,
+      overall_modal,
+    ]
+  } only %}
+{% endset %}
+
+{% set twig_markup %}
+{% verbatim %}
+{% include '@bolt-components-modal/modal.twig' with {
+  content: 'This is a modal set to overall scrolling.',
```
mikemai2awesome, line 77: We should note when doing an image modal, you can set `spacing` and `theme` to none to show only the image.

docs-site/src/pages/pattern-lab/_patterns/40-components/modal/30-modal-trigger-variations.twig
```diff
@@ -1,97 +1,111 @@
...
+      link_modal,
+    ]
+  } only %}
+{% endset %}
+
+{% include '@utils/pattern-doc-page.twig' with {
+  title: 'Modal Trigger Options',
+  description: description,
+  notes: notes,
+  demo: demo,
```
mikemai2awesome, line 110: This is missing code snippets. And the image modal is broken, the modal is still taking full width of screen instead of the width of the image.

mikemai2awesome (changes requested):

danielamorse (approved):
@mikemai2awesome @colbytcook updated the CSS to leverage the existing overlay, [see commit](https://github.com/boltdesignsystem/bolt/pull/2414/commits/f07b207865725cac0918199e29c47d0251df20a7).

docs-site/src/pages/pattern-lab/_patterns/40-components/modal/30-modal-trigger-variations.twig
colbytcook: @mikemai2awesome I am seeing this bug in Safari, but it seems like it is an existing bug (https://boltdesignsystem.com/pattern-lab/?p=components-modal-trigger-variations). I will spend sometime trying to fix this. If not, I will create a bug ticket to handle this another time.

colbytcook: @mikemai2awesome I created https://pegadigitalit.atlassian.net/browse/DS-699 to track this. I was not able to track down what was causing this bug.

mikemai2awesome (approved):

PR #2415, opened 2022-01-11
DS-517 Type element

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-517

## Summary

Adds a new Type element that will replace both Text and Headline components.

## Details

The element has a set of typographic styles that can be controlled via props. Twig and plain HTML usage are supported. Note that it is named `Type` to avoid Pattern Lab conflicts with the Text component.

## How to test

Run the branch locally and go through all docs under Elements > Type.


packages/elements/bolt-type/src/type.twig
```diff
@@ -0,0 +1,41 @@
...
+  {% if link_attributes %}
+    {% include '@bolt-elements-text-link/text-link.twig' with {
+      content: this.data.content.value,
+      icon_before: icon_before ? icon_before : null,
+      icon_after: icon_after ? icon_after : null,
+      reversed_underline: true ? this.data.hierarchy.value != schema.properties.hierarchy.default,
+      attributes: link_attributes,
+    } only %}
+  {% else %}
+    {% if icon_before %}<span class="e-bolt-type__icon-before" aria-hidden="true">{{ icon_before|spaceless }}</span>{% endif %}{{ this.data.content.value|spaceless }}{% if icon_after %}<span class="e-bolt-type__icon-after" aria-hidden="true">{{ icon_after|spaceless }}</span>{% endif %}
```
remydenton, line 39: Why `this.data.content.value` and not just `content`?  I think we want to follow the example of text link element, which just does [content|spaceless](https://github.com/boltdesignsystem/bolt/blob/master/packages/elements/bolt-text-link/src/text-link.twig#L33).  Otherwise, we'll run into issues when someone wants to pass something that isn't a string, like a renderable object.

packages/elements/bolt-type/type.schema.js
```diff
@@ -0,0 +1,56 @@
...
+  title: 'Type',
+  type: 'object',
+  properties: {
+    attributes: {
+      type: 'object',
+      description:
+        'A Drupal-style attributes object with extra attributes to append to this element.',
+    },
+    content: {
+      type: 'string',
```
remydenton, line 12: Same idea as my other comment, this should probably be
```suggestion
      type: 'any',
```

remydenton (changes requested):

packages/elements/bolt-type/src/type.twig
mikemai2awesome: updated

remydenton (approved):

docs-site/src/pages/pattern-lab/_patterns/20-elements/type/05-type.twig
```diff
@@ -0,0 +1,33 @@
...
+{% endverbatim %}{% endset %}
+
+{% set html_markup %}{% verbatim %}
+<p class="e-bolt-type">This is a type element. Use common inline HTML elements to <em>emphasize</em>, <strong>bold</strong>, <a href="#!">link</a>, and more.</p>
+<p>If no props are set, the above is exactly the same as a <p> HTML element.</p>
+{% endverbatim %}{% endset %}
+
+{% include '@utils/pattern-doc-page.twig' with {
+  title: 'Basic Type',
+  description: 'The type element displays text in Pega branded typographic styles.',
```
danielamorse, line 28: Should we just say "branded typographic styles" to avoid mention of "Pega"?

danielamorse (commented):
@mikemai2awesome this looks great 😎 

Only remaining issue is the schema isn't showing in the docs. At first glance, everything looks correct. I will have to debug more on Monday.

docs-site/src/pages/pattern-lab/_patterns/20-elements/type/05-type.twig
mikemai2awesome: I think it's only `pega.com` that's problematic.

danielamorse:
Schema issue fixed in https://github.com/boltdesignsystem/bolt/pull/2499. Will merge once next build passes.

cjwhitedev (approved):

PR #2418, opened 2022-01-19
DS-697 Prevent Listing teaser from outputting empty div

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-697

## Summary

The if conditional was added to prevent rendering empty div - `c-bolt-listing-teaser__flag-content`

## Details

`c-bolt-listing-teaser__flag-content` is rendered if one of the following properties is added: 

- `meta_items`
- `description`
- `warning`
- `chip_list`
- `reply`

## How to test

Check if div with the class `c-bolt-listing-teaser__flag-content` is rendered only when one or more properties are added


MarcinMr:
@danielamorse template updated

danielamorse (approved):
Tested locally, works perfectly.

mikemai2awesome (approved):

PR #2419, opened 2022-01-19
DS-692 Animation Tool Set

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-692

## Summary

Animated classes with keyframes were added to `packages/global/styles/animations`.

## Details

There were created **fade**, **bounce**, **zoom** (**ins**/**outs**) combined with slide effects (**top**, **bottom**, **left**, **right**) + modifier classes controlling the cascading appearance of elements and delays options.

### fade
- fade in / out / in-up, out-up / in-down, out-down / in-left, out-left / in-right, out-right

### zoom
- zoom in / out / in-up, out-up / in-down, out-down / in-left, out-left / in-right, out-right

### bounce
- bounce in / out / in-up, out-up / in-down, out-down / in-left, out-left / in-right, out-right

---
Test pages were created in the `tests/animations` folder for experimenting.

## How to test

Check if animations work properly by adding specific classes to the elements considered to be animated. Make sure the performance of animating elements is sufficient.


packages/global/styles/06-animations/_animations-base.scss
```diff
@@ -0,0 +1,73 @@
...
+
+.a-bolt-fade {
+  visibility: hidden; //Safari performance issue
+  animation-delay: var(--bolt-animation-delay);
+  animation-duration: var(--bolt-animation-duration);
+  animation-fill-mode: both;
+  animation-play-state: paused;
+
+  @media print, (prefers-reduced-motion: reduce) {
+    opacity: 1 !important;
```
mikemai2awesome, line 18: I think this will be a problem, certain components use opacity, we have to avoid overriding it on the base class. I think you can just do `animation: none !important;` and remove everything.

packages/global/styles/06-animations/_animations-fade-in-down.scss
```diff
@@ -0,0 +1,20 @@
+/* ------------------------------------ *\
+   Animation fade in down
+\* ------------------------------------ */
+
+@keyframes fade-in-down {
```
mikemai2awesome, line 5: We'll need to use the `a-bolt-*` convention for all keyframe names as well to avoid conflicts with custom code.

packages/global/styles/06-animations/_animations-on-load-test.js
```diff
@@ -0,0 +1,9 @@
+function playStateRunningOnLoad() {
```
mikemai2awesome, line 1: I am not sure about the handling of JS here. Ideally, we don't run this JS at all if "prefer reduced motion" is on.

packages/global/styles/06-animations/_animations-base.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+.a-bolt-fade--delay-fast {
+  --bolt-animation-delay: 100ms;
+  animation-delay: var(--bolt-animation-delay);
+}
+
+//these classes fade animation one by one
+@for $i from 1 through 1000 {
+  .a-bolt-fade--successively:nth-of-type(#{$i}) {
```
mikemai2awesome, line 61: I understand what `a-bolt-fade--successively` is doing but I think the naming is confusing.

@danielamorse Grammar boy, how do we name this behavior?

mikemai2awesome (changes requested):
@MarcinMr Good job! I only have a few inline comments and we'll discuss performance with the team in dev huddle.

packages/global/styles/06-animations/_animations-base.scss
MarcinMr: Updated

packages/global/styles/06-animations/_animations-fade-in-down.scss
MarcinMr: Updated

packages/global/styles/06-animations/_animations-base.scss
MarcinMr: Yeah... I wasn't sure how to name it. I was thinking also about: `.a-bolt-fade--gradually`, `.a-bolt-fade--one-by-one`, `.a-bolt-fade--step-by-step`, `.a-bolt-fade--little-by-little`, `.a-bolt-fade--in-succession` :) there's a lot of synonyms, but I'm not sure which works best for english

packages/global/styles/06-animations/_animations-on-load-test.js
MarcinMr: I think we have to use `function playStateRunningOnLoad()` as we have our `.a-bolt-fade` with `animation-play-state: paused` and `visibility: hidden` at the begining, and `function playStateRunningOnLoad()` adds `.a-bolt-fade--running` which overrides css `animation-play-state` and `visibility`.

What could I do is write an if statement that triggers `function playStateRunningOnLoad()` only when reduced-motion is off, but then I have to add `visibility: visible` besides `animation: none !important;`, to the media query like below 

```
@media print, (prefers-reduced-motion: reduce) {
    visibility: visible;
    animation: none !important;

    &[class*='--out'] {
      visibility: visible;
      animation: none !important;
    }
  }
  ```
  
  Is that ok?

**Comment by MarcinMr** [comment]
@mikemai2awesome 
I made change requests with some comments. We can go over it on Monday if necessary

File: packages/global/styles/06-animations/_animations-base.scss (hunk 4 above)
Comment (mikemai2awesome) [reply packages/global/styles/06-animations/_animations-base.scss:57]: I've think "cascade" describes this but we'll discuss in huddle.

File: packages/global/styles/06-animations/_animations-on-load-test.js (hunk 3 above)
Comment (mikemai2awesome) [reply packages/global/styles/06-animations/_animations-on-load-test.js:1]: Overriding visibility is basically the same problem as overriding opacity, I don't think that's ideal. We'll have to discuss JS with @danielamorse in huddle.

File: packages/global/styles/06-animations/_animations-base.scss [hunk 5]
```diff
@@ -0,0 +1,69 @@
...
+
+.a-bolt-fade {
+  visibility: hidden; //Safari performance issue
+  animation-delay: var(--bolt-animation-delay);
+  animation-duration: var(--bolt-animation-duration);
+  animation-fill-mode: both;
+  animation-play-state: paused;
+
+  @media print, (prefers-reduced-motion: reduce) {
+    &[class*='--out'] {
```
mikemai2awesome, line 18: We need a better condition for this, `--out` is too broad, it could override classes that matches that string. And you can test to find out, I don't think we need this condition since we are using `!important` already.

mikemai2awesome (changes requested):
Thanks @MarcinMr , we'll discuss the remaining issues monday.

packages/global/styles/06-animations/_animations-base.scss
MarcinMr: @mikemai2awesome yes, `animation: none !important;` is enough to stop the animation when `reduced motion` is on.  

The one thing that puzzles me is should all **fade-outs** be invisible when `reduce motion` is on?

Right now when `reduce motion` is on every **fade-ins** is rendered without animation and visible on the screen, and that's ok.

But the same is with **fade-outs**. All of them are rendered and visible on the screen. But since the animation of **fade-outs** is about being invisible at the end, shouldn't they be invisible when `reduce motion` is on?

packages/global/styles/06-animations/_animations-on-load-test.js
MarcinMr: ok

mikemai2awesome (changes requested):
@MarcinMr I did some more thinking and would like you to try the following.

```scss
// _animations.scss
[class*='a-bolt-'] {
  animation-delay: var(--bolt-animation-delay);
  animation-duration: var(--bolt-animation-duration);
  animation-fill-mode: both;
  animation-play-state: paused;

  @media print, (prefers-reduced-motion: reduce) {
    animation: none !important;
  }
}
[class*='a-bolt-in'] { 
  animation-timing-function: ease-out; 
}
[class*='a-bolt-out'] { 
  animation-timing-function: ease-in;
  
  @media print, (prefers-reduced-motion: reduce) {
    opacity: 0;
  }
}
.a-bolt-running {}
.a-bolt-duration-long {}
.a-bolt-duration-short {}
.a-bolt-delay-long {}
.a-bolt-delay-short {}
.a-bolt-cascade-slow {}
.a-bolt-cascade-fast {}

// _animations-fade-in.scss
.a-bolt-in--fade {}
.a-bolt-in--fade-up {}
.a-bolt-in--fade-down {}
.a-bolt-in--fade-left {}
.a-bolt-in--fade-right {}

// _animations-fade-out.scss
.a-bolt-out--fade {}
.a-bolt-out--fade-up {}
.a-bolt-out--fade-down {}
.a-bolt-out--fade-left {}
.a-bolt-out--fade-right {}

// More to come
.a-bolt-in--bounce {}
.a-bolt-in--bounce-up {}
.a-bolt-in--bounce-down {}
.a-bolt-in--bounce-left {}
.a-bolt-in--bounce-right {}
.a-bolt-in--zoom {}
.a-bolt-in--zoom-up {}
.a-bolt-in--zoom-down {}
.a-bolt-in--zoom-left {}
.a-bolt-in--zoom-right {}
```

I want to organize with a convention that will scale, so we don't end up with too many files within the animations folder. Fade is just one type of animations, it shouldn't be a base class. And the delay, duration, and cascade can all be set with any type of animations (bounce, zoom, slide, etc.), they are not limited to fade. Same goes for in and out, they can apply to any type of animations.

You are correct that things that are supposed to go out should be invisible when reduced motion is enabled. So I think it's ok to set opacity 0 with the media query.

mikemai2awesome (changes requested):

packages/global/styles/06-animations/_animations.scss
```diff
@@ -0,0 +1,82 @@
...
+   Base animations tool set
+\* ------------------------------------ */
+
+:root {
+  --bolt-animation-duration: 200ms;
+  --bolt-animation-delay: 0ms;
+}
+
+[class*='a-bolt-'] {
+  visibility: hidden; //Safari performance issue
```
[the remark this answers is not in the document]
MarcinMr: If I don't set `visibility: hidden` Safari flickers the icons like in the record below. Some icons appear in random places and after that animation starts.

https://user-images.githubusercontent.com/85670248/151324110-a480c5bd-0743-4740-9687-6f6539e412b8.mov

packages/global/styles/06-animations/_animations.scss
```diff
@@ -0,0 +1,82 @@
...
+.a-bolt-duration-short {
+  animation-duration: calc(var(--bolt-animation-duration) / 2);
+}
+
+.a-bolt-delay-long {
+  --bolt-animation-delay: 500ms;
+  animation-delay: var(--bolt-animation-delay);
+}
+
+.a-bolt-delay-medium {
```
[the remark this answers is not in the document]
MarcinMr: removed

packages/global/styles/06-animations/_animations.scss
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.a-bolt-delay-short {
+  --bolt-animation-delay: 100ms;
+  animation-delay: var(--bolt-animation-delay);
+}
+
+//these classes fade animation one by one
+@for $i from 1 through 150 {
+  .a-bolt-cascade:nth-of-type(#{$i}) {
```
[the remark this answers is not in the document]
MarcinMr: removed

packages/global/styles/06-animations/_animations.scss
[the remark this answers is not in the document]
mikemai2awesome: I see. I added your explanation to the comment in css.

mikemai2awesome (approved):
@MarcinMr I reduced cascade to 25 to keep the file size down. We will need to figure out how to do this better later on.

packages/global/styles/06-animations/_animations.scss
```diff
@@ -0,0 +1,67 @@
...
+  animation-delay: var(--bolt-animation-delay);
+}
+
+.a-bolt-delay-short {
+  --bolt-animation-delay: 100ms;
+  animation-delay: var(--bolt-animation-delay);
+}
+
+/* [2] */
+@for $i from 1 through 150 {
```
danielamorse, line 58: @MarcinMr we _should_ try to use CSS instead of JS when possible, but this is really better handled by JS because:
1. We should avoid magic numbers (150)
2. The code inside this block gets repeated in the compiled code 150x, adding +40kb of CSS to bolt-global.css

JS can do this kind of thing programmatically with much less code ([see JS](https://github.com/boltdesignsystem/bolt/pull/2428/files#diff-497dcda9f7d8f9424038f7d20c5d34675a2af1b04c495ee65b96014bf7cfe985)).

Also, just a reminder to always use `@for` loops with caution in SCSS :) They are super helpful but can generate a lot of code.

docs-site/src/pages/pattern-lab/_patterns/999-tests/animations/banner-test-fade-ins.twig
```diff
@@ -0,0 +1,40 @@
...
+    setTimeout(() => {
+      e.target.classList.add(...runAnimationClasses)
+    }, 20)
+  }
+
+  fadeEl.forEach((el) => {
+    el.addEventListener('click', triggerThisEl)
+  })
+
+</script>
```
danielamorse, line 39: @MarcinMr well done on the JS! I had [a few suggestions](https://github.com/boltdesignsystem/bolt/pull/2428/files#diff-56b89f9594b79a0edb031b5954ac02b38c9526e0266f5c8d67777f3942d8be70). In my edits, I:
- Work with `classList` (an array) rather than `className` (a string). It's just nicer to work with for what you're doing here.
- [Destructure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) the `target` argument, since that's the only part of `e` we're using.
- Changed the setTimeout from `20` to `0`. `0` should be enough to trigger the animation, it just says to execute this after, not all at the same time.

We can now start sending you more JS work ;)

danielamorse (changes requested):
@MarcinMr I made a separate PR so you could review my suggestions. Feel free to merge _that_ PR if it all makes sense: https://github.com/boltdesignsystem/bolt/pull/2428

This all looks great, just had some notes on the cascade animation and the demo JS. See inline comments.

cc @mikemai2awesome @colbytcook

packages/global/styles/06-animations/_999-animations-on-load-test.js
```diff
@@ -0,0 +1,9 @@
+function playStateRunningOnLoad() {
+  document.querySelectorAll('[class*="a-bolt-"]').forEach(fade => {
+    fade.classList.add('a-bolt-running');
+  });
+}
+
+window.onload = () => {
```
danielamorse, line 7: Instead of `window.onload` we should use `window.addEventListener('load', () => { })`. I updated this [on my branch](https://github.com/boltdesignsystem/bolt/pull/2428/files#diff-497dcda9f7d8f9424038f7d20c5d34675a2af1b04c495ee65b96014bf7cfe985R39).

In most cases you don't want to use `window.onload` because you might unintentionally [overwrite other event listeners](https://carstenbehrens.com/never-use-window-onload/).

packages/global/styles/06-animations/_animations.scss
mikemai2awesome: Thanks Dan, I would love to remove this chunk of CSS.

MarcinMr: Thanks Dan for that solution. That SCSS solution only works when we are 100% that there are for example 10 elements to animate, but doesn't make sense if we don't know the number of elements which gonna be animated, so the SCSS hardcoded number doesn't work here

colbytcook (approved):

danielamorse (approved):
My JS edits have been merged and everything else looks great.

I just merged in the latest from `master` to fix the failing build (chromedriver).

PR #2421, opened 2022-01-19
SN-96 Skip JS validation on React form elements

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/SN-96

## Summary

Updates the selector used to apply JS validation so that React form elements are skipped.

## Details

You can now add the class `js-bolt-react-component` to any React input which also uses the `c-bolt-input` class for styles and opt out of the default Bolt validation.

On Accounts this validation is handled via a separate React library.

## How to test

Review code and verify existing for JS works as expected. For example, blurring a required field triggers an error message.


remydenton (approved):
Looks great!

PR #2423, opened 2022-01-21
DS-541 Floating Action Buttons

cjwhitedev:
## Jira

- [DS-541](https://pegadigitalit.atlassian.net/browse/DS-541)
- [DS-540](https://pegadigitalit.atlassian.net/browse/DS-540)

## Summary

Create a component for floating action buttons.

## Details

1. Added the new Floating Action Buttons component
2. Created full page demos to showcase the component usage
3. Fixed a bug in Button element where icon is way too big for small icon-only buttons

## How to test

1. Pull Down the Branch and update.
2. Check out the demo pages.

## Release notes

New component named Floating Action Buttons added.

### Visual changes

Icon-only button with size set to `small` renders smaller icon now. Overall button size is not affected.


packages/components/bolt-satellite/satellite.schema.js
```diff
@@ -0,0 +1,42 @@
...
+      type: 'object',
+      description:
+        'A Drupal attributes object. Applies extra HTML attributes to the &lt;bolt-satellite&gt; tag.',
+    },
+    content: {
+      type: 'any',
+      description:
+        'The main content. Typically, this will be a number of icon-only buttons.',
+    },
+    position: {
```
mikemai2awesome, line 16: We are lacking offset options to move this out of the way when there are sticky things on the page that's out of our control, such as Chatbot button and Cookie Consent panel.

Also, let's keep out testing things out of our schemas.

mikemai2awesome (changes requested):
@cjwhitedev This is a good first pass. First, let's keep this simple and just name this Floating Action Buttons rather than a fancy term that doesn't resonate.

Secondly, let's go over some visual enhancements and semantics fixes.

## Visual

Let's create some information hierarchy here. If the more button toggles on additional buttons, these would be considered secondary. Rough sketch below:

fab

Additionally, it'd be great to smooth out the transition with some motion effects [like this](https://codepen.io/th3no0b/pen/ZvVMZe).

## Semantics

This is a list of buttons, let's apply the correct semantics in the markup.

```html
<ul>
  <li>
    <a href="">more</a>
    <ul>
      <li>
        <a href="">
          rocket
        </a>
      </li>
      <li>
        <a href="">
          rocket
        </a>
      </li>
      <li>
        <a href="">
          rocket
        </a>
      </li>
    </ul>
  </li>
  <li>
    <a href="">share</a>
  </li>
  <li>
    <a href="">scroll to top</a>
  </li>
</ul>
```

packages/components/bolt-satellite/satellite.schema.js
cjwhitedev: @mikemai2awesome 

1. There is a CSS var to control the offset. Should it be a prop instead?
2. How can I use it in the demos if its not in the schema?

mikemai2awesome: @cjwhitedev You can document it similar to [what we did with Image](https://boltdesignsystem.com/pattern-lab/?p=elements-image-use-case-background-fit-and-position). Have a dedicated page for advanced usage.

cjwhitedev: @mikemai2awesome I believe I've hit every point here. Please re-review this PR.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
+export class BoltFloatingActionButtons {
+  constructor(el) {
+    if (!el) return;
+    this.el = el;
+    this.init();
+  }
+
+  init() {
+    this.visibleItems = this.el.getAttribute('data-visible-items');
```
danielamorse, line 9: From JS use: `this.el.dataset.visibleItems`

https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+    this.contentElement = this.el.querySelector(
+      '.c-bolt-floating-action-buttons__content',
+    );
+    this.showOnScroll = this.el.getAttribute('show-on-scroll');
+    this.showOnLoad = this.el.getAttribute('data-show-on-load');
```
danielamorse, line 14: Ditto, use `dataset`

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+      '.c-bolt-floating-action-buttons__more',
+    );
+
+    if (this.visibleItems > 0) {
+      this.setDefaultVisibility();
+    }
+
+    if (this.showOnScroll) {
+      this.handleshowOnScroll();
+    } else if (this.showOnLoad != null) {
```
danielamorse, line 25: The `null` check is probably unnecessary here, just `!this.showOnLoad` should do.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+    this.expandButton = this.el.querySelector(
+      '.c-bolt-floating-action-buttons__more',
+    );
+
+    if (this.visibleItems > 0) {
+      this.setDefaultVisibility();
+    }
+
+    if (this.showOnScroll) {
+      this.handleshowOnScroll();
```
danielamorse, line 24: Fix camelcase, should be `handleShowOnScroll`.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+    } else if (this.showOnLoad != null) {
+      this.show();
+    }
+
+    if (this.expandButton) {
+      this.expandButton.onclick = () => {
+        this.toggleShowMore();
```
danielamorse, line 31: Do "expandButton" and "showMore" refer to the same piece of UI? If so, let's use the same name for both. This makes me think they are two separate things.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+      };
+    }
+
+    // add the ready attribute at the end of init
+    // delay it just long enough to avoid CSS animation lag
+    setTimeout(() => {
+      this.el.setAttribute('data-bolt-ready', '');
+    }, 200);
```
danielamorse, line 39: What exactly does this fix? What animation lag happens without a delay?

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  show() {
+    this.el.setAttribute('aria-hidden', false);
+  }
+
+  getScrollPositionFromProp() {
+    var revealPosition = 0;
```
danielamorse, line 47: Use `let`/`const`

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+      revealPosition = scrollInt;
+    } else if (this.showOnScroll.includes('%')) {
+      // set a pixel value based on the percentage value.
+      revealPosition = pageHeight * (scrollInt / 100);
+    }
+    return revealPosition;
+  }
+
+  handleshowOnScroll() {
+    var scrollPosition = window.scrollY;
```
danielamorse, line 61: Use `let`/`const`

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+    var scrollInt = parseInt(this.showOnScroll, 10);
+    var pageHeight = window.innerHeight;
+
+    if (this.showOnScroll.includes('px')) {
+      revealPosition = scrollInt;
+    } else if (this.showOnScroll.includes('%')) {
+      // set a pixel value based on the percentage value.
+      revealPosition = pageHeight * (scrollInt / 100);
+    }
+    return revealPosition;
```
danielamorse, line 57: Nice way to handle `px`/`%`

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  handleshowOnScroll() {
+    var scrollPosition = window.scrollY;
+    var revealPosition = this.getScrollPositionFromProp();
+
+    const superFunction = () => {
```
danielamorse, line 64: `superFunction` is fun, but `scrollHandler` is more descriptive.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+    };
+
+    window.addEventListener('scroll', superFunction);
+  }
+
+  toggleShowMore() {
+    if (this.el.getAttribute('aria-expanded') === 'false') {
+      this.el.setAttribute('aria-expanded', 'true');
+      // show all contentItems
+      for (const contentItem of this.contentElement.children) {
```
danielamorse, line 79: I wouldn't use `children`. I would add a querySelector to `init()` that finds these elements based on classname. Use a `js-` classname [per the convention](https://gitlab.com/pegadigital/pega-digital-standards-and-conventions/-/wikis/Standards-Naming#front-end-examples).

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+        contentItem.setAttribute('aria-hidden', false);
```
danielamorse, line 80: Should this be 'false'?

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+      this.el.setAttribute('aria-expanded', 'false');
+      this.setDefaultVisibility();
+    }
+  }
+
+  setDefaultVisibility() {
+    let index = 1;
+    for (const contentItem of this.contentElement.children) {
+      if (index > this.visibleItems) {
+        contentItem.setAttribute('aria-hidden', true);
```
danielamorse, line 92: Ditto, 'true'?

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+      } else {
+        contentItem.setAttribute('aria-hidden', false);
```
danielamorse, line 94: Ditto, 'false'?

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+    } else {
+      this.el.setAttribute('aria-expanded', 'false');
+      this.setDefaultVisibility();
+    }
+  }
+
+  setDefaultVisibility() {
+    let index = 1;
+    for (const contentItem of this.contentElement.children) {
```
danielamorse, line 90: See comment above about using a `js-` selector to find the children. Then, try using `map()` here. Map is good when you want to perform an operation on each item of an array. Let me know if you have any questions refactoring.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

danielamorse (changes requested):
@cjwhitedev This is really coming together. Almost there! I left you some inline comments.

Another thought - In our demos maybe we should position tooltips to the left when FAB is vertical so that the tooltip doesn't block the button above?

image

mikemai2awesome:
@cjwhitedev The markup still doesn't make sense, let's try again.

**Currently this:**
```
<div class="c-bolt-floating-action-buttons c-bolt-floating-action-buttons--position-bottom-right" aria-hidden="false" aria-expanded="true" data-visible-items="2" data-show-on-load="" data-bolt-ready="">
  <ul class="c-bolt-floating-action-buttons__content">
    <li class="c-bolt-floating-action-buttons__content-item">
      <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Share"></button>
    </li>
    <li class="c-bolt-floating-action-buttons__content-item">
      <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Back to top"></button>
    </li>
    <li class="c-bolt-floating-action-buttons__content-item">
      <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Secondary"></button>
    </li>
    <li class="c-bolt-floating-action-buttons__content-item">
      <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Secondary"></button>
    </li>
    <li class="c-bolt-floating-action-buttons__content-item">
      <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Secondary"></button>
    </li>
  </ul>
  <div class="c-bolt-floating-action-buttons__more">
    <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="More"></button>
  </div>
</div>
```

**What we want to get to is this:**
```html
<div class="c-bolt-floating-action-buttons c-bolt-floating-action-buttons--position-bottom-right" aria-hidden="false" data-visible-items="2" data-show-on-load data-bolt-ready>
  <ul class="c-bolt-floating-action-buttons__list">
    <li class="c-bolt-floating-action-buttons__list-item">
      <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Share"></button>
    </li>
    <li class="c-bolt-floating-action-buttons__list-item">
      <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Back to top"></button>
    </li>
    <li class="c-bolt-floating-action-buttons__list-item">
      <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="More" aria-expanded="false"></button>
      <ul class="c-bolt-floating-action-buttons__list">
        <li class="c-bolt-floating-action-buttons__list-item">
          <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Secondary"></button>
        </li>
        <li class="c-bolt-floating-action-buttons__list-item">
          <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Secondary"></button>
        </li>
        <li class="c-bolt-floating-action-buttons__list-item">
          <button type="button" class="e-bolt-button e-bolt-button--icon-only" aria-label="Secondary"></button>
        </li>
      </ul>
    </li>
  </ul>
</div>
```

`aria-expanded` is to be used on a `<button>` not a `<div>`. Also reference my rough design previously, when "more" is clicked on to expand, the ellipsis changes to an X.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
cjwhitedev: as far as i can tell, `dataset` (or getAttribute) doesn't translate to booleans. the attribute is either there or it isn't. is there another way to approach this?

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
cjwhitedev: Without the delay, on page load we can see the component for a moment before it hides itself. With the delay, the CSS animations happen behind the scenes for that .2 second before it becomes visible. Is there a better way to do this because I don't feel great about it as-is.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
cjwhitedev: @danielamorse does it need the js-if there is an existing class name to target? look at `this.contentElement` in this file to see what I mean.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
cjwhitedev: I don't think so - setting hidden to true would mean they are hidden, wouldn't it? I'm trying to show all the items here.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
cjwhitedev: see above

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
danielamorse: Yes, we use `js-` as a way of saying "JS depends upon this class". It's more verbose but clearer and harder to break.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
danielamorse: I meant `'false'` vs `false`, string vs Boolean.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
danielamorse: Oh you are totally right. I think [forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) is better here. You get the `index` automatically as the second arg.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.scss
```diff
@@ -0,0 +1,78 @@
...
+    transition: all var(--bolt-transition);
+  }
+
+  .c-bolt-floating-action-buttons__toggle {
+    position: relative;
+    z-index: 1;
+  }
+
+  // Hide secondary list
+  .c-bolt-floating-action-buttons__toggle[aria-expanded='false']
```
mikemai2awesome, line 41: By removing the `.c-bolt-floating-action-buttons__toggle` class, you can just use `e-bolt-button` here.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.scss
```diff
@@ -0,0 +1,78 @@
...
+    + .c-bolt-floating-action-buttons__list {
+    > .c-bolt-floating-action-buttons__list-item {
+      visibility: hidden;
+      opacity: 0;
+      margin-bottom: -100%;
+    }
+  }
+
+  .c-bolt-floating-action-buttons__toggle-icon {
```
mikemai2awesome, line 50: Also remove this class, we can use `nth-child` instead to handle the 2 icons. We should avoid mixing component classes with another component/element class.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.scss
```diff
@@ -0,0 +1,78 @@
...
+$default-x-offset: var(--bolt-spacing-x-xsmall);
+$default-y-offset: var(--bolt-spacing-y-xsmall);
+
+.c-bolt-floating-action-buttons {
+  z-index: bolt-z-index('fab');
+
+  // must both be ready && not have aria-hidden before showing
+  &:not([data-bolt-ready]),
+  &[aria-hidden='true'] {
+    @include bolt-visuallyhidden;
```
mikemai2awesome, line 19: When it's hidden, it needs to be completely hidden, not visually hidden. You should not be able to tab to these buttons if they are hidden.

mikemai2awesome (changes requested):
@cjwhitedev Great job setting up the list structure. I have some inline comments and the following:

## Transition

1. The transition of the children action buttons are a bit choppy, might have to do with using `margin` which costs more than `transform`. Consult our animation expert @MarcinMr on improving this.
2. The button element tooltip transition is not working that well, you can see it in the icon only button doc page, sometimes the tooltip doesn't trigger at all (in Safari). Again, consult Marcin about this.
3. `show_on_scroll` needs fade in and fade out transitions.
4. Remove transitions when `@media (prefers-reduced-motion) {}` is true.

## Hidden

`aria-hidden` is supposed to be used for hiding things from just screen readers, which doesn't really apply to what we are doing with this component, because we are hiding things entirely. We shouldn't need to write css based on `aria-hidden`.

For the main container, you can toggle a `--hidden` modifier class when `show_on_scroll` is set.

For the toggle button, you handle the css with just `aria-expanded`:

```css
.e-bolt-button[aria-expand='false'] ~ .c-bolt-floating-action-buttons__list > .c-bolt-floating-action-buttons__list-item  {
  // hide the list items
}
```

packages/components/bolt-floating-action-buttons/src/floating-action-buttons-toggle-button.twig
```diff
@@ -0,0 +1,40 @@
...
+  'e-bolt-button',
+  'e-bolt-button--icon-only',
+  'c-bolt-floating-action-buttons__toggle',
+  'js-bolt-floating-action-buttons-toggle',
+  this.data.size.value ? 'e-bolt-button--' ~ this.data.size.value : '',
+  this.data.hierarchy.value ? 'e-bolt-button--' ~ this.data.hierarchy.value : '',
+  this.data.border_radius.value ? 'e-bolt-button--border-radius-' ~ this.data.border_radius.value : '',
+] %}
+
+<button
```
danielamorse, line 21: Is there a particular reason we're using HTML here and not the Button element Twig template? If it's to add a class somewhere the template doesn't allow can we explain that in a comment please?

packages/elements/bolt-button/button.schema.js
```diff
@@ -28,6 +28,13 @@ module.exports = {
       description:
         'Append an icon to the button content and visually hide the text content. This prop trumps icon_before and icon_after.',
     },
+    icon_only_tooltip: {
```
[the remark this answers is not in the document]
danielamorse: When you're ready to look at this one, let's have a quick call @cjwhitedev. I need to get up to speed.

danielamorse (changes requested):
@cjwhitedev I took another pass through and left comments.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.scss
cjwhitedev: @mikemai2awesome I was just trying this out and was running into an issue where it was hidding the children in the buttons i dont want it to. i think if we were to keep the `toggle` class from above this would be more feasible. Do you have thoughts on an alternate approach?

packages/components/bolt-floating-action-buttons/floating-action-buttons.schema.js
```diff
@@ -0,0 +1,34 @@
...
+      type: 'string',
+      description: 'The desired location of the floating element.',
+      enum: ['bottom-right', 'middle-right', 'top-right'],
+      default: 'bottom-right',
+    },
+    show_on_load: {
+      type: 'boolean',
+      description:
+        'Please leave this set to true unless you plan on adding javascript to show the element yourself.',
+      default: true,
```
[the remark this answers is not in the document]
mikemai2awesome: that sounds reasonable, then we'd need to follow our convention of boolean defaults to `false`. something like `hide_on_load`. @danielamorse @cjwhitedev 

let's say we are to do that, i think having 2 props to achieve one thing is kind of excessive:

```
{
  hide_on_load: true,
  show_on_scroll: '20%',
}
```

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.scss
mikemai2awesome: The different ways of hidden as discussed in our chat:

```
html:
aria-hidden="true" -> hide only from screen readers
hidden -> hide completely

css:
bolt-visuallyhidden -> hide only from visual
display: none -> hide completely
visibility: hidden -> hide completely but takes up space
opacity: 0 -> hide only from visual but takes up space
```

To hide/show the main container, it requires css `visibility: hidden` and `opacity` already, other methods are not to be used.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.scss
mikemai2awesome: As discussed in our chat, using the `children` prop as the logic to render a modifier class for the list item will allow us to not write additional classes on the button.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.scss
mikemai2awesome: As discussed in our chat, using the children prop as the logic to render a modifier class for the list item will allow us to not write additional classes on the button.

MarcinMr:
@mikemai2awesome @cjwhitedev I did an example of the children's btns animation. Had to change js for testing purposes, this might need to be refactored, didn't make cascade variables dynamic, but hardcoded at this moment in CSS as I don't know if this is what we are looking for.

mikemai2awesome (changes requested):
@cjwhitedev As we discussed over chat, here are the things to look into:

1. The component needs to work without JS (the toggle button requires JS to expand/collapse, but we cannot hide the entire thing due to 1 button, similar to page header, it still appears without JS but the mega nav expand/collapse function is not expected to work).
2. Based on 1, the component should not have a hidden class by default.
3. The JS doesn't boot if a toggle button is not passed to the component (eg. passing only the "back to top" button and share button).
4. The JS hangs on every few refresh. Reproduce this by refreshing the page a few times till you hit one that doesn't load ready data attribute.
5. The `right-start` placement option does not account for the sticky page header, I've added a CSS hack for now. Figure out how to do it properly with Dan.

I am going to wrap up the following:

1. Animate the toggle button's icons.
2. Finish reformatting the docs.

cc @danielamorse

mikemai2awesome:
@cjwhitedev This is working great now. We have `show_on_scroll`, but how can we hide the FAB again if user scrolls back up? The scroll to top button on Collaboration Center fades away when user [scrolls back up](https://collaborate.pega.com/question/database-commit-operation-slow-after-oracle-19c-upgrade), we need to cover this use case.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,124 @@
...
+            el.classList.add(
+              'c-bolt-floating-action-buttons__list-item--hidden',
+            );
+          }
+        }
+      });
+    };
+
+    const observer = new IntersectionObserver(callback);
+    observer.observe(document.querySelector(el.dataset.showOnScrollSelector));
```
danielamorse, line 122: Assign `document.querySelector(el.dataset.showOnScrollSelector)` to a variable and check for it at the very top of this function. If it's not present `return`.

packages/components/bolt-floating-action-buttons/src/floating-action-buttons.js
```diff
@@ -0,0 +1,124 @@
...
+    this.primaryList = this.el.querySelector(
+      '.js-bolt-floating-action-buttons-list',
+    );
+    this.secondaryList = this.primaryList.querySelector(
+      '.js-bolt-floating-action-buttons-list',
+    );
+    this.toggleButton = this.primaryList.querySelector(
+      '.js-bolt-floating-action-buttons-toggle',
+    );
+    this.hiddenListItems = this.el.querySelectorAll(
```
danielamorse, line 18: Instead of using the presence of the hidden class, use the show on scroll data attribute.

danielamorse (changes requested):
@cjwhitedev left a few inline requests. Please also update demo titles and description to reflect the new prop name (still see examples of "show-on-scroll-position"), and remember to prefix commit messages with the ticket number 🙏

danielamorse (approved):
@cjwhitedev approved! 🎉  Thanks for sticking with this one. I think it turned out great.

I made two last tweaks. One is a very minor JS style thing - I tend to skip brackets for single-line `if` statement. It's totally fine to include brackets (some prefer that), but for consistency I removed them ([see commit](https://github.com/boltdesignsystem/bolt/pull/2423/commits/d13f397c18f337ebec9eb0e6fa79ad88eaa97117)). Second, I removed the remaining scroll-position demos ([see commit](https://github.com/boltdesignsystem/bolt/pull/2423/commits/5d2983ccd0712c40d1470d523ce09961bc87fd14)).

Before merging I just want to double-check this has Mike's approval. I think he approved before many recent changes.

mikemai2awesome (approved):
@danielamorse @cjwhitedev I fixed some docs and tested with screen readers. Good to go.

PR #2430, opened 2022-01-31
Fix deploy script

danielamorse:
## Jira

n/a

## Summary

Fixes bug where pre-release tags get deployed to prod.

## Details

Add check to prevent pre-release (beta, rc, etc.) branches from being aliased to production site.

I tested this myself by creating a temporary `v9.9.9-beta-2` tag (now deleted), pushing it, and verifying it does not get aliased to production.

## How to test

- Review code

### DIFF
--- a/scripts/deploy-tagged-release.js
+++ b/scripts/deploy-tagged-release.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 const { promisify } = require('util');
+const semver = require('semver');
 const gitSemverTags = require('git-semver-tags');
 const promisifyGitTags = promisify(gitSemverTags);
 const { getLatestDeploy } = require('./utils');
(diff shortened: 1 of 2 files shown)


scripts/deploy-tagged-release.js
```diff
@@ -16,7 +17,13 @@ getLatestDeploy()
       aliasNowUrl(url, latestTag);
     }
 
-    if (TRAVIS_TAG && TRAVIS_TAG === latestTag && !latestTag.includes('rc')) {
+    // Alias prod (boltdesignsystem.com) if this is the latest semver tag and it is NOT a pre-release
+    if (
+      TRAVIS_TAG &&
+      TRAVIS_TAG === latestTag &&
+      !semver.valid(latestTag) &&
```
colbytcook, line 24: @danielamorse this line seems wrong unless I am misunderstanding? Wouldn't this be checking to see if the latest tag is NOT valid?

danielamorse: You are totally right. Nice catch! Fixed.

colbytcook (approved):

scripts/deploy-tagged-release.js
```diff
@@ -16,7 +17,13 @@ getLatestDeploy()
       aliasNowUrl(url, latestTag);
     }
 
-    if (TRAVIS_TAG && TRAVIS_TAG === latestTag && !latestTag.includes('rc')) {
+    // Alias prod (boltdesignsystem.com) if this is the latest semver tag and it is NOT a pre-release
```
remydenton, line 20: Appreciate the comment here-- definitely adds some clarity.

remydenton (approved):

PR #2434, opened 2022-02-04
Fix: add max width to regular modal

mikemai2awesome:
## Summary

Adds a max-width to regular modals.

## Details

1. Updated CSS for the `regular` option for Modal component's width prop
2. The max width is set to be exactly the same as the overall site max width (1400px)

## How to test

Run the branch locally and check the modal docs. Modals with width set to `regular` should be no wider than 1400px.

### Visual changes

Any modal that has `width` set to `regular` now has a max-width of 1400px. This means the modal width will not get wider than 1400px. For modals that require going beyond that max-width, use `width: full` instead.


danielamorse (approved):
Tested locally, works perfectly. Thanks for including the clear release notes.

PR #2437, opened 2022-02-14
DS 516 table component rework

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-516

## Summary

Table WC is replaced by a new responsive table component

## Details

- the table was created using the semantic HTML tags,
- the table can have headers (top and aside), the body of the table, and a footer,
- on a smaller screens table is scrollable horizontally,
- aside headers get sticky on small screens,
- top header gets sticky when the height of the table is more than 75vh,

Some of the [UI enhancements from here](https://github.com/boltdesignsystem/bolt/wiki/POC%3A-Table-Enhancements) were added: 

- toggling aside checkboxes
- toggling hidden rows
- toggling filter's menu, checkboxes, and a reset filters button feature
- toggling sort button states

## How to test

Focus on the responsiveness of the table and an accessibility


mikemai2awesome (changes requested):
@MarcinMr This is a good start, let's take care of the following:

1. When the table side scrolls, the content gets chopped off. There is no visual cue that the table is scrollable. Try to work in some shadows as we previously had or do another design.
2. Fix the inconsistent border color (vertical borders in between cells are lighter than horizontal borders), they should be the same color.
3. Fix the vertical border misalignment, avoid using `:before` and `:after` for the borders which could be the cause of the bug.
3. Add a prop to go completely borderless as you can see in the POC design.
4. Match the sortable table headers exactly as you see in the POC design.
5. For each cell with a checkbox, make sure the text align when it wraps to the next line. Do not wrap text under the checkbox.
6. The filter button would open a popover for filter options (see POC design), mock this up in the demo.
7. As for the expandable row feature, [read up on this post first](https://adrianroselli.com/2019/09/table-with-expando-rows.html).
8. The expand button doesn't work, write some JS to make the demo work as expected, we'll refine it later.
9. Fix the colors. Some colors are not displaying well in various color themes.

Lastly let's rework the schema to make sure devs have attribute control on every table elements, we are gonna need to build multiple templates:

```twig
{% set header %}
  {% include '@bolt-components-table/table-cell.twig' with {
    content: 'This is a header cell.',
    header: true,
  } only %}
  {# Add more header cells here #}
{% endset %}

{% set row %}
  {% set cells %}
    {% include '@bolt-components-table/table-cell.twig' with {
      content: 'This is a regular cell.',
    } only %}
    {% include '@bolt-components-table/table-cell.twig' with {
      content: 'This cell spans 2 columns and has an utility class.',
      attributes: {
        colspan: 2,
        class: 'u-bolt-color-navy-light'
      }
    } only %}
    {# Add more cells here #}
  {% endset %}
  {% include '@bolt-components-table/table-row.twig' with {
    content: cells,
  } only %}
{% endset %}

{% set footer %}
  {% include '@bolt-components-table/table-cell.twig' with {
    content: 'This is a footer cell.',
  } only %}
  {# Add more cells here #}
{% endset %}

{% include '@bolt-components-table/table.twig' with {
  header: {
    content: header,
    attributes: {},
  },
  body: {
    content: [
      row,
      row,
      row,
    ],
    attributes: {},
  },
  footer: {
    content: footer,
    attributes: {},
  },
  caption: {
    content: 'This is the table caption.',
    attributes: {},
  },
} only %}
```

To do a row with a header as the first cell, it would be like this:

```twig
{% set row_with_side_header %}
  {% set cells %}
    {% include '@bolt-components-table/table-cell.twig' with {
      content: 'This is a header cell.',
      header: true,
    } only %}
    {% include '@bolt-components-table/table-cell.twig' with {
      content: 'This is a regular cell.',
    } only %}
    {# Add more cells here #}
  {% endset %}
  {% include '@bolt-components-table/table-row.twig' with {
    content: cells,
  } only %}
{% endset %}

{% include '@bolt-components-table/table.twig' with {
  body: {
    content: [
      row_with_side_header,
      row_with_side_header,
      row_with_side_header,
    ],
    attributes: {},
  },
} only %}
```

To do expandable row, it would be like this:

```twig
{% set row_expandable %}
  {% set cells %}
    {% set expand_button %}
      {% include '@bolt-components-table/table-expand-button.twig' with {
        content: 'Expand more rows',
        ids: [
          'hidden-row-1',
          'hidden-row-2',
        ],
        attributes: {},
      } only %}
    {% endset %}

    {% include '@bolt-components-table/table-cell.twig' with {
      content: expand_button,
    } only %}
    {% include '@bolt-components-table/table-cell.twig' with {
      content: 'This is a regular cell.',
    } only %}
    {# Add more cells here #}
  {% endset %}

  {% include '@bolt-components-table/table-row.twig' with {
    content: cells,
  } only %}
{% endset %}

{% set row_hidden_1 %}
  {% set cells %}
    {% include '@bolt-components-table/table-cell.twig' with {
      content: 'This is a regular cell.',
    } only %}
    {# Add more cells here #}
  {% endset %}
  {% include '@bolt-components-table/table-row.twig' with {
    content: cells,
    hidden: true,
    attributes: {
      id: 'hidden-row-1',
    },
  } only %}
{% endset %}

{% set row_hidden_2 %}
  {% set cells %}
    {% include '@bolt-components-table/table-cell.twig' with {
      content: 'This is a regular cell.',
    } only %}
    {# Add more cells here #}
  {% endset %}
  {% include '@bolt-components-table/table-row.twig' with {
    content: cells,
    hidden: true,
    attributes: {
      id: 'hidden-row-2',
    },
  } only %}
{% endset %}

{% include '@bolt-components-table/table.twig' with {
  body: {
    content: [
      row_expandable,
      row_hidden_1,
      row_hidden_2,
    ],
    attributes: {},
  },
} only %}
```

mikemai2awesome:
@MarcinMr Thanks. Let me think about the shadows and borders to see if there's another way.

mikemai2awesome:
@MarcinMr I pushed up some `.test-table` CSS for you to reference, that solves the scrolling visual cue by using box-shadow to mimic an overall border. Can you use that code as the foundation to work out the other things?

MarcinMr:
> @MarcinMr I pushed up some `.test-table` CSS for you to reference, that solves the scrolling visual cue by using box-shadow to mimic an overall border (I think it works just as well as the old shadow). Can you use that code as the foundation to work out the other things? See the example on the first doc page. I had to create 2 new global variables but I think this works nicely.

@mikemai2awesome Thanks Mike, I used the code you provided. Here is how it is:

1. Using `margin: -1px` will cause the bottom and right border a little bit thicker, this is because of `width: 100%` but we need it to stretch the table. Otherwise, we will have a grey background to the right if the table does not fill the whole width space.

margin -1px

2. Without `margin: -1px` looks better, outer border is the same thickness on every side, but still a little bit thicker than the cell borders
margin 0

3. But your solutions make it possible to use sticky on the left-column-header and the top-row-header - borders are now visible on the stuck cells, which is good I think, I had to remove `overflow-y: hidden` from the global vars, if we want to have sticky top headers.

https://user-images.githubusercontent.com/85670248/154475442-92480471-9ba4-4cbd-8d70-2a8b917a194a.mov

4. And when we have a caption we have a border around it. In my opinion that doesn't look bad, but maybe we can avoid it somehow?
table caption border


_What about points 7 and 8 from the previous conversation?_

mikemai2awesome:
@MarcinMr I created a few templates and simplified the table cell template:

```twig
// Sortable table header
{% set sorting_text_button %}
  {% include '@bolt-components-responsive-table/table-sorting-text-button.twig' with {
    content: content,
  } only %}
{% endset %}
{% set filters %}
  {% include '@bolt-components-responsive-table/table-filter-button.twig' with {
    content: content,
  } only %}
{% endset %}
{% include '@bolt-components-responsive-table/table-cell.twig' with {
  content: sorting_text_button,
  filters: filters,
  header: true,
} only %}

// Checkbox table cell
{% set checkbox_input %}
  {% include '@bolt-components-responsive-table/table-checkbox-input.twig' with {
    content: content,
  } only %}
{% endset %}
{% include '@bolt-components-responsive-table/table-cell.twig' with {
  content: checkbox_input,
} only %}
```

Use these new templates inside table cell, this will allow flexibility for devs to use our templates for these features or build their own. It's all up to them based on what they passed to table cell's `content` and `filters` prop.

I need you to finish up these templates.

1. For `table-sorting-text-button.twig`, you'll see 3 different states in there: unsorted, ascending, and descending. Clicking on the buttons will toggle between the 3 states. I added new icons to make the design work.
2. For the `table-filter-button.twig`, the content prop would accept a list of checkbox inputs. Try to see if you can just use the existing form component inputs here.
3. For the `table-checkbox-input.twig`, I copied over the input and label you created, but this template might not be necessary if you could just use the existing form component input. So test the existing component first. You will need to rebuild the table checkbox doc page as well.

docs-site/src/pages/pattern-lab/_patterns/40-components/responsive-table/20-responsive-table-sort-and-filter.twig
```diff
@@ -0,0 +1,617 @@
...
+  twig_markup: twig_markup,
+  html_markup: html_markup
+} only %}
+
+
+<script>
+  function toggleFilters() {
+    document.querySelector('body').addEventListener('click', event => {
+    
+      if (event.target.matches('.c-bolt-input--checkbox')) {
```
mikemai2awesome, line 595: Use `js-` classes for targeting. Add `.js-filters-checkbox` to each checkbox.

docs-site/src/pages/pattern-lab/_patterns/40-components/responsive-table/20-responsive-table-sort-and-filter.twig
```diff
@@ -0,0 +1,617 @@
...
+        const resetBtn = event.target.closest('.c-bolt-fieldset').nextElementSibling;
```
mikemai2awesome, line 596: Same here, use a `js-` class.

docs-site/src/pages/pattern-lab/_patterns/40-components/responsive-table/20-responsive-table-sort-and-filter.twig
```diff
@@ -0,0 +1,617 @@
...
+        if(resetBtn) {
+          resetBtn.classList.add('visible')
```
mikemai2awesome, line 598: Use an existing util class `u-bolt-hidden`.

mikemai2awesome (changes requested):

docs-site/src/pages/pattern-lab/_patterns/40-components/responsive-table/05-responsive-table-checkbox.twig
```diff
@@ -0,0 +1,474 @@
...
+  notes: notes,
+  demo: demo,
+  twig_markup: twig_markup,
+  html_markup: html_markup
+} only %}
+
+<script>
+  function toggleCheckbox() {
+    const mainCheckbox = document.querySelector(
+      '.c-bolt-responsive-table__head .c-bolt-input--checkbox',
```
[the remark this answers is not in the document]
MarcinMr: Fixed

docs-site/src/pages/pattern-lab/_patterns/40-components/responsive-table/20-responsive-table-sort-and-filter.twig
MarcinMr: used `u-bolt-hidden`

docs-site/src/pages/pattern-lab/_patterns/40-components/responsive-table/20-responsive-table-sort-and-filter.twig
MarcinMr: fixed

packages/components/bolt-responsive-table/src/responsive-table.scss
```diff
@@ -0,0 +1,307 @@
...
+  gap: var(--bolt-spacing-x-small);
+  align-items: center;
+
+  .c-bolt-responsive-table__button {
+    display: flex;
+    color: var(--m-bolt-headline);
+  }
+}
+
+.tippy-box {
```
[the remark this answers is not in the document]
MarcinMr: removed styling `tippy-box` and used `u-bolt-hidden` for that

docs-site/src/pages/pattern-lab/_patterns/40-components/responsive-table/20-responsive-table-sort-and-filter.twig
MarcinMr: event target class changed

mikemai2awesome:
@MarcinMr Please take a look at the sticky headers, they are not working as expected:

sticky-headers

We'll discuss in huddle to see if we want to roll this out as breaking change for table component or as a separate component. Let's clean up the code after that's decided.

MarcinMr:
@mikemai2awesome The sticky should work with the basic table, table with side headers, and expandable rows, it should work now I hope. We need to use `c-bolt-responsive-table__first-cell-sticky` class in `attributes` on a `thead` if we want the first cell (top left corner to be sticky).

https://user-images.githubusercontent.com/85670248/155335230-de253f3d-6456-40ac-84af-ab6805fcbc8c.mov

https://user-images.githubusercontent.com/85670248/155335248-8d61d646-d089-4570-bc07-601e38a2b6a6.mov

https://user-images.githubusercontent.com/85670248/155335261-98e58fcc-4934-4ea9-80b0-8fd8e043961e.mov

When it comes to those doc pages with the prefix `old` I think we can remove them, I've kept them only for my reference, the same as the old `responsive-table` template. And since we decide to use our existing `input` component for checkboxes I believe the template chunk `table-checkbox-input.twig` can also be removed as we don't use it.

mikemai2awesome (changes requested):
@MarcinMr 

1. We had a discussion in dev huddle, we decided to remove classes on inner table elements. Only keeping classes on the parent container. I've refactored the CSS for that.
2. I also updated the expand button template to use just a text link element rather than something. Please hook up the `attributes` because I broke that.
3. I created modifiers `c-bolt-responsive-table--sticky-side-headers` and `c-bolt-responsive-table--sticky-top-headers`. We need a prop named `sticky_header` and `enum` set to `top`, `side`, and `none`. We'll make sticky header an option instead of default. Also, top sticky headers cannot work with side sticky headers so we'll only allow one at a time.
4. Lastly, let's rename the component to `Table`, we have decided to go with a breaking change.
5. Once renamed and cleaned up. Make schema validations work in all templates.

MarcinMr:
> @MarcinMr
> 
> 1. We had a team discussion in dev huddle, we decided to remove classes on inner table elements. Only keeping classes on the parent container. I've refactored the CSS for that. I didn't update the twig templates, you'll need to change those to remove the classes.
> 2. I also updated the expand button template to use just a text link element rather than something new. Please hook up the `attributes` because I broke that.
> 3. I created modifiers `c-bolt-responsive-table--sticky-side-headers` and `c-bolt-responsive-table--sticky-top-headers`. We need a prop named `sticky_headers` and `enum` set to `top`, `side`, and `none`. We'll make sticky header an option instead of default. Also, top sticky headers cannot work with side sticky headers so we'll only allow one at a time.
> 4. Lastly, let's rename the component to `Table`, we have decided to go with a breaking change.
> 5. Once renamed and cleaned up. Make schema validations work in all templates.

@mikemai2awesome did 1,2,3,4,5, but the schema is not working properly as I mentioned on slack

MarcinMr:
@mikemai2awesome 

I think we should reset all sort button states after a user clicks a specific sorting button. For example, the toggling feature so far looked like this:

https://user-images.githubusercontent.com/85670248/156571207-a36476a9-a020-4404-aeb3-8286376c3db9.mov

As you can see when I click on the first button then the second and the third, all previous clicked button's states remain. But actually, we will have the table sorted after the recent button is clicked. So this can be misleading because we have still arrow indicators on the first and the second button.

Since we are sorting tables that means we actually sort table rows (not columns), so I think we should reset all button's states and have the active state on the recently clicked button:

https://user-images.githubusercontent.com/85670248/156572053-c3d2c2d7-4c12-439a-9b04-dc50d06e7ad8.mov

mikemai2awesome:
@MarcinMr Agree, we'll let @danielamorse handle that with the JS work.

packages/core-v3.x/styles/02-tools/tools-scrolling/_tools-scrolling.scss
```diff
@@ -25,6 +25,5 @@
 /// }
 @mixin bolt-horizontal-scroll {
   overflow-x: auto;
-  overflow-y: hidden;
   -webkit-overflow-scrolling: touch;
```
mikemai2awesome, line 28: Don't change the horizontal mixin. If you are looking for both horizontal and vertical scrolling, you need to create a new one that does it.

```
@mixin bolt-scroll {
  overflow: auto;
  -webkit-overflow-scrolling: touch;
}
```

packages/components/bolt-table/src/table.twig
```diff
@@ -1,156 +1,48 @@
...
+{% set this = init(schema) %}
+{% set attributes_wrapper = create_attribute(attributes|default({})) %}
+{% set header_attributes = create_attribute(header.attributes|default({})) %}
+{% set body_attributes = create_attribute(body.attributes|default({})) %}
+{% set footer_attributes = create_attribute(footer.attributes|default({})) %}
+{% set caption_attributes = create_attribute(caption.attributes|default({})) %}
+
+{# Array of classes based on the defined + default props #}
+{% set classes = [
+  'js-bolt-table',
```
mikemai2awesome, line 16: Do we need this JS class?

mikemai2awesome (changes requested):

packages/components/bolt-table/src/table.twig
MarcinMr: I think this is necessary when we will have template scripts in `bolt-table/src/table.js` to target sort and filter logic. Btw shouldn't the already existing doc's scripts be moved to the `bolt-table/src/table.js`

mikemai2awesome: Remove the JS for now. Dan's going to do that in a separate ticket. We can release table with JS first.

MarcinMr: `js-bolt-table` class is removed from template

MarcinMr:
@mikemai2awesome One more thing is the `<caption/>` element. I moved it as a first child of the table because I read the table docs from MDN ----> [table caption spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption). And there is a statement:

table caption

So I think, if the `<caption>` is used, must be the first child of the `<table>` element.

mikemai2awesome (approved):
@MarcinMr Good job! I am approving this and will create a separate ticket to refine theme colors for table. And Dan has a separate ticket for complex table JS.

colbytcook (commented):
@MarcinMr @mikemai2awesome I do not know if this is intentional or not but I am not seeing the shadow on the right side of a responsive table (this would be in Firefox). If this is expected then I approve this PR otherwise there might be some more work here.

mikemai2awesome:
@colbytcook Shadow requires the old approach so we can't use it. We've switched to use a slightly different design with borders.

colbytcook (approved):

MarcinMr:
@mikemai2awesome  I found a typo in the schema.
typo schema

there should be generates `<caption`>element

updated and pushed

mikemai2awesome:
Get the updated colors from this https://github.com/boltdesignsystem/bolt/pull/2476 before merging.

packages/build-tools/tasks/api-tasks/bolt-status-board.js
```diff
@@ -11,25 +11,78 @@ const events = require('@bolt/build-utils/events');
...
         class: [
-          "t-bolt-xlight"
+          't-bolt-xlight',
+          'u-bolt-block'
         ],
-        id: "component-status"
+        id: 'component-status',
+        style: 'max-height: none;'
       },
+<<<<<<< HEAD
```
danielamorse, line 84: @MarcinMr would you please resolve these merge conflict tags?

MarcinMr: Yes, I'll do it

danielamorse (changes requested):
@MarcinMr left a couple inline comments/requests. Also, I removed the JS for now since it's not setup yet and the next version will most likely use React and look a bit different.

packages/global/styles/00-vars/_vars-mode.scss
```diff
@@ -20,6 +20,9 @@
   --m-bolt-disabled: var(--bolt-color-gray-light);
   --m-bolt-text-on-disabled: var(--bolt-color-gray-xdark);
   --m-bolt-border: var(--m-bolt-tertiary);
+  --m-bolt-table-header: var(--bolt-color-gray-xlight);
```
danielamorse, line 23: @MarcinMr it seems odd that we have to add component-specific variables here, when everything else is generic/global. What's the reason? Did you cover this with @mikemai2awesome already?

packages/build-tools/tasks/api-tasks/bolt-status-board.js
MarcinMr: @danielamorse Should be ok now

packages/global/styles/00-vars/_vars-mode.scss
MarcinMr: @danielamorse Honestly, I can't remember why we did that. But I checked the history and [here link to the comment](https://github.com/boltdesignsystem/bolt/pull/2437#issuecomment-1042375408) that says that new variables were created. Maybe that was for better readability?

danielamorse (approved):

PR #2440, opened 2022-02-24
Ds 729 Scope button state styles to a and button elements

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-729

## Summary

HTML elements different than `<a>` and `<button>` can look like the bolt button element but without the interactive features.

## Details

Interactive features of `<button>` like:

- hover
- active
- cursor pointer

were scoped only to semantic HTML `<button>` and `<a>` tags.

A test page of the examples was added to the Tests folder.

## How to test

Pull the branch. Try to add `e-bolt-button`  class to different HTML tags than `<a>` and `<button>`. For example `<span>` or `<div>`.

### Visual changes

If an HTML tag with an `e-bolt-button` class is not a semantic `<button>` or `<a>`, the element will only look like the button (shape, color, etc), without the interactive features of the button.

### DIFF
--- a/packages/elements/bolt-button/src/button.scss
+++ b/packages/elements/bolt-button/src/button.scss
@@ -104,23 +104,8 @@ a.e-bolt-button {
 $_bolt-button-hierarchy: primary, secondary, tertiary, transparent;
 
 @each $hierarchy-name in $_bolt-button-hierarchy {
-  .e-bolt-button--#{$hierarchy-name} {
-    @if $hierarchy-name == tertiary or $hierarchy-name == transparent {
-      button.e-bolt-button,
-      a.e-bolt-button {
-        &:hover {
-          transform: translate3d(0, 0, 0);
-
-          &:after {
-            opacity: 0.2;
-            box-shadow: none;
-            background-color: var(--bolt-color-gray);
-          }
-        }
-      }
-    }
-
-    @if $hierarchy-name != transparent {
+  @if $hierarchy-name != transparent {
+    .e-bolt-button--#{$hierarchy-name} {
       --e-bolt-button-text-color: var(--m-bolt-text-on-#{$hierarchy-name});
       --e-bolt-button-bg-color: var(--m-bolt-#{$hierarchy-name});
 
@@ -141,7 +126,9 @@ $_bolt-button-hierarchy: primary, secondary, tertiary, transparent;
           box-shadow: 0 0 0 2px var(--m-bolt-primary);
         }
       }
-    } @else {
+    }
+  } @else {
+    .e-bolt-button--#{$hierarchy-name} {
       --e-bolt-button-text-color: var(--m-bolt-link);
       --e-bolt-button-bg-color: transparent;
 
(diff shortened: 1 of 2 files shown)


packages/elements/bolt-button/src/button.scss
```diff
@@ -61,20 +61,26 @@
     opacity: 0;
   }
 
-  &:hover {
-    transform: translate3d(0, -2px, 0);
+  &[type='button'],
```
mikemai2awesome, line 64: All the conditions should be scoped to `a` and `button`, not these attributes.

```css
a.e-bolt-button,
button.e-bolt-button {}
```

mikemai2awesome (changes requested):

packages/elements/bolt-button/src/button.scss
```diff
@@ -69,39 +84,38 @@
...
   .e-bolt-button--#{$hierarchy-name} {
     @if $hierarchy-name == tertiary or $hierarchy-name == transparent {
-      &:hover {
-        transform: translate3d(0, 0, 0);
-
-        &:after {
-          opacity: 0.2;
-          box-shadow: none;
-          background-color: var(--bolt-color-gray);
+      button.e-bolt-button,
```
mikemai2awesome, line 109: This will not work, it's going to compile:

```scss
.e-bolt-button--tertiary {
    button.e-bolt-button,
    a.e-bolt-button { ... }
}
```

That is targeting button inside a button.

You'll need to scope like this:

```scss
@each $hierarchy-name in $_bolt-button-hierarchy {
  @if $hierarchy-name != transparent {
    .e-bolt-button--#{$hierarchy-name} {
      --e-bolt-button-text-color: var(--m-bolt-text-on-#{$hierarchy-name});
      --e-bolt-button-bg-color: var(--m-bolt-#{$hierarchy-name});

      @if $hierarchy-name == secondary {
        &:before {
          opacity: 0.4;
        }
      }

      @if $hierarchy-name == tertiary {
        backdrop-filter: blur(0.1em);

        &:before {
          opacity: 0;
        }

        &:after {
          box-shadow: 0 0 0 2px var(--m-bolt-primary);
        }
      }
    } @else {
      --e-bolt-button-text-color: var(--m-bolt-link);
      --e-bolt-button-bg-color: transparent;

      &:before,
      &:after {
        display: none;
      }
    }
  }

  @if $hierarchy-name == tertiary or $hierarchy-name == transparent {
    button.e-bolt-button--#{$hierarchy-name},
    a.e-bolt-button--#{$hierarchy-name} {
      &:hover {
        transform: translate3d(0, 0, 0);

        &:after {
          opacity: 0.2;
          box-shadow: none;
          background-color: var(--bolt-color-gray);
        }
      }
    }
  }

  @if $hierarchy-name == transparent {
    button.e-bolt-button--#{$hierarchy-name},
    a.e-bolt-button--#{$hierarchy-name} {
      &:hover {
        text-decoration: underline;
        text-decoration-thickness: 2px;
        text-underline-offset: 3px;
      }
    }
  }
}
```

mikemai2awesome (changes requested):

packages/elements/bolt-button/src/button.scss
MarcinMr: Yeah, that's true Mike. I didn't notice that, because in the `tests`  doc page I rendered only the primary button. In fact when I added other options: `secondary`, `transparent`, `tertiary` the `:hover` was not what it should be in those button options. I hope now the scope should be fine.

MarcinMr:
> @MarcinMr Sorry I found one more regression, we broke [this one](https://feature-ds-729-scope-button-state-styles.boltdesignsystem.com/pattern-lab/?p=elements-button-upload). We need to add one more scope:
> 
> ```css
> button.e-bolt-button,
> a.e-bolt-button,
> input[type='file'].e-bolt-button { ... }
> ```

updated

mikemai2awesome (approved):

danielamorse (approved):

PR #2441, opened 2022-02-24
WWW-1289 fix page footer button's chevron placement in iOS 14

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/WWW-1289

## Summary

Fixes an issue where the Page Footer button's chevron icon is out of place.

## Details

Updated CSS to use flex instead of grid.

## How to test

Run the branch locally and check the page footer doc with viewport set to 480px or smaller. Make sure the chevron icon that  is in each section heading is displayed on the same line as the text and to the right edge.


remydenton (approved):
Works as expected, thanks!

PR #2449, opened 2022-03-07
DS-498 Close Nav on Anchor Click

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-498](https://pegadigitalit.atlassian.net/browse/DS-498)

## Summary

Close the Page Header after clicking an anchor link.

## Details

After clicking an anchor link to an item listed on the same page, it is expected that the menu would close and reveal the related content. This was especially visible on mobile. I added an event listener to handle this action.

## How to test

1. Add an `ID` to a header tag on the doc site
    a. I did it on the Most Popular heading in `_patterns/50-pages/65-best-of-content/_boc-gallery.twig`
2. Add a link to that `ID`.
    a. I did it here: `_patterns/40-components/page-header/_demo-main-site-page-header.twig`
3. Start your local server and click the demo link.
4. Confirm the menu closes as expected.

## Release notes

Resolved expected behavior for page header menu to close when clinking an on-page anchor link.


danielamorse (approved):
@cjwhitedev nice, looks good. I made one tweak which was changing this:
```
if (e.target.hash || e.target.parentNode.hash) {
```
to this:
```
if (e.target.closest('a')?.hash) {
```

The latter covers you if the target is nested more than one level, which is less common but could happen (perhaps with an icon inside a span).

I wasn't crazy about having both an "external" click handler and an "anchor" click handler when they're both listening for clicks on the whole document, but after thinking about it I landed the same place as you did. In this case, I'd say it's better to have two listeners that handle specific cases than to have one which does it all but is harder to understand. Both event handlers are properly removed when the menu closes so I'm good with all of this 👌

PR #2452, opened 2022-03-17
DS-630: Screenreader Keypress Navigation in Page Header

cjwhitedev:
## Jira
[https://pegadigitalit.atlassian.net/browse/DS-630](https://pegadigitalit.atlassian.net/browse/DS-630)

## Summary

`Enter` and `Space` key interaction not working with screen readers active.

## Details

The screen readers are sending `click` events instead of `keypress` events when using `Enter` or `Space`. This ticket is to add the click `click` events and allow users to navigate the page header while using a screen reader.

## How to test

Test in BrowserStack with NVDA on, or otherwise use PAWS and NVDA, then make sure you can navigate the pege header.

## Release notes
Modify page-header keypress interaction.

### DIFF
--- a/packages/components/bolt-page-header/src/page-header-nav.js
+++ b/packages/components/bolt-page-header/src/page-header-nav.js
@@ -94,6 +94,7 @@ export class BoltPageHeaderNav {
   setupDesktopMenu() {
     this.addHoverHandler(this.menu);
     this.addKeypressHandler(this.menu);
+    this.addClickHandler(this.menu);
     this.state.desktopIsSetup = true;
   }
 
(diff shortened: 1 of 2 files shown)


mikemai2awesome (approved):
Tested and it works as expected.

danielamorse (commented):
Ok, I remember the rationale for adding _keypress_ handlers not _click_ handlers. On desktop the Menu opens on hover, not on click. If you are navigating via keyboard there is no hover, so a keypress makes sense. If you listen for both a hover an a click event, you get a user experience like this: user hovers and menu shows, user clicks and menu hides. Is this what we want? cc @mikemai2awesome @cjwhitedev

mikemai2awesome:
@danielamorse yeah that's fine. clicking can close it.

danielamorse (approved):
@mikemai2awesome, @cjwhitedev I did some cleanup of the original code now that we no longer listen for space/enter keyboard events. Specifically, I consolidated the escape key handlers. Basically, now every menu listens for escape key hits with the same function and there's no need for an option to turn it on/off.

I tested pretty thoroughly. Would you please retest this branch once more before merging? Be sure to check all the interactive menus on desktop and mobile, using both mouse and keyboard. 🙏

PR #2454, opened 2022-03-18
DS-745 Support PHP 8

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-745

## Summary

Removes dependency on two composer packages that are incompatible with PHP 8

## Details

This PR removes the following two dependencies, which are incompatible with PHP 8 and abandoned:

1. [tooleks/php-avg-color-picker](https://github.com/tooleks/php-avg-color-picker).  The last commit to this repo was 5 years ago and it was only used in one optional place to provide a color while images are loading.
2. [fzaninotto/faker](https://github.com/fzaninotto/faker).  A mostly useless shiny object library that also happens to be an abandoned project.  We had a whole custom PHP class to override it to get it to work with our test snapshots, and we weren't using it in any recent examples.  I removed a few examples that were preventing PL from building (there are some others in the archive directory that as-is I left because I felt it was a waste of time to update them-- in the off chance we ever need them again, we can replace the text with Lorem Ipsum at that time).

For faker, I also had to remove it as a sub dependency of our fork of `basaltinc/twig-tools`: https://github.com/boltdesignsystem/twig-tools/pull/2.  We should first merge that PR, then update this PR to use the master branch before merging it.

## How to test

Confirm no regressions on this branch.  Confirm that faker text has been replaced with hard-coded lorem ipsum.  Confirm that any differences in the color of the placeholder for bolt images while they load is minor/unnoticeable.


andrewmriv:
The changes looked good and made sense to me. Unfortunately, it looks like there is a build error.

andrewmriv (approved):

danielamorse:
@remydenton This looks good to me. Agree it's a good time to remove these libraries.

I noticed there are some instances of `faker` still in the `docs-site/src/pages/pattern-lab/_patterns/999-archive/academy` folder. If these are intentionally ignored that's totally fine. If not, you might want to update these before merging (optional).

PR #2455, opened 2022-03-22
Fix: typeahead search button misalignment

mikemai2awesome:
## Summary

Fixes a visual regression where the typeahead's search button is misaligned.

## Details

1. Cleaned up Typeahead's HTML and CSS
2. Removed unnecessary  CSS that is prone to bugs

## How to test

Run the branch locally and check the typeadead docs with and without JS. When tabbing to the icon-only search button within the typeahead, it should not be misaligned.


danielamorse (approved):
Reviewed code, tested locally. Fixes bug as expected, nice cleanup of additional cruft 👌

PR #2464, opened 2022-03-29
DS-752 Remove ssr-server and all related code (Part 1)

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-752

## Summary

Remove ssr-server and all other experimental SSR code.

## Details

SSR was an experiment to pass HTML to a back-end service and get back rendered web component HTML. It mostly worked but was never used and isn't needed going forward. We're removing it as part of repo clean up.

## How to test

- Review code changes
- Twig continues to render as expected (doesn't touch any normal Twig rendering)
- Travis builds as expected


packages/build-tools/cli.js
```diff
@@ -125,9 +125,6 @@ program
         config.prod =
           typeof program.prod === 'undefined' ? config.prod : program.prod;
 
-        // automatically set enableSSR to true in prod mode and false in dev mode, unless manually set.
-        config.enableSSR = false;
```
remydenton, line 129: The comment must be inaccurate?  Looks to me like it's disabled in both prod and dev modes...

remydenton (approved):
Does the fact that I love seeing deleted code mean I'm getting old?  I'll just tell myself it's the fact that the deleted code had incorrect comments and logic bugs.  In any case, 😍 .

The code changes make sense to me, and it looks like the build is passing, approved.

colbytcook (approved):

PR #2466, opened 2022-03-30
DS-442: Holy Grail Layout Cookie

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-442](https://pegadigitalit.atlassian.net/browse/DS-442)

## Summary

We want to remember the user's preference when interacting with the holy grail side nav.

## Details

I added a cookie library to the holy grail layout and used it to set and get a cookie that will show or hide the side nav based on previous user interaction.

## How to test

- Use a cookie manager to confirm the cookie is being set after interaction with the element.
- confirm that this is the desired behavior by interacting with the element and refreshing the page.

### DIFF
--- a/packages/layouts/bolt-holy-grail/package.json
+++ b/packages/layouts/bolt-holy-grail/package.json
@@ -16,7 +16,8 @@
   "style": "index.scss",
   "dependencies": {
     "@bolt/components-icon": "^5.2.0",
-    "@bolt/core-v3.x": "^5.0.0"
+    "@bolt/core-v3.x": "^5.0.0",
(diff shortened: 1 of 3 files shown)


mikemai2awesome (approved):
Working as expected! I'll have @danielamorse check the JS.

danielamorse (approved):
@cjwhitedev looks good! I just added an expiration to the cookie. Otherwise, it gets removed when the user closes the browser (not just the window).

```
Cookies.set('l-bolt-holy-grail--hidden', 'false', { expires: 90 });
```

PR #2471, opened 2022-04-06
DS-712 Remove twig whitespace added around tooltips

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-712

## Summary

Whitespace around the tooltip trigger was removed.

## Details

Whitespace was removed using CSS properties. `Spaceless` didn't want to work as mentioned in the Jira description.

## How to test

- Review code
- Check out this branch locally and see the new test page: `999-tests --> Tooltip Whitespace` (`/pattern-lab/?p=tests-tooltip-whitespace`)
- Check if there is no more whitespace around the tooltip trigger.


colbytcook (approved):
Confirmed this fix works on Safari (OS and iOS), Firefox, Chrome and Edge (Mac)

docs-site/src/pages/pattern-lab/_patterns/40-components/tooltip/15-tooltip-inverted.twig
```diff
@@ -1,7 +1,7 @@
 {% set tooltip %}
   {# Start component specific code #}
   {% include '@bolt-components-tooltip/tooltip.twig' with {
-    trigger: 'this example',
+    trigger: 'this example.',
```
mikemai2awesome, line 4: Why did this doc change? The period should be outside of the tooltip.

docs-site/src/pages/pattern-lab/_patterns/999-tests/tooltip/20-tooltip-whitespace.twig
```diff
@@ -0,0 +1,9 @@
+{% set tooltip %}
```
mikemai2awesome, line 1: This test is not sufficient. To test if the tooltip is truly inline without undesired white space, you'd need to layout all the scenarios such as:

This is a paragraph of text, a tooltip can appear anywhere in here, in the middle of the sentence **_like this_** or at the end of the sentence before a punctuation **_like this_**.

$**_A tooltip_** can also begin a sentence where a special character is used before it.

mikemai2awesome (changes requested):

docs-site/src/pages/pattern-lab/_patterns/40-components/tooltip/15-tooltip-inverted.twig
MarcinMr: dot-space

dot-space-2

I've changed it because when the dot is outside the trigger there is still space between the text and the dot, like in the examples above.

Here `/pattern-lab/?p=tests-tooltip-whitespace` I updated examples of tooltip triggers.

mikemai2awesome: That is exactly the problem. We need to solve the undesired white-space in between the text and the period. Not to change the doc to hide the problem.

docs-site/src/pages/pattern-lab/_patterns/999-tests/tooltip/20-tooltip-whitespace.twig
mikemai2awesome: Good. Now this test perfectly demonstrates the undesired white space, which is a side effect of `inline-block`.

whitespace

docs-site/src/pages/pattern-lab/_patterns/40-components/tooltip/15-tooltip-inverted.twig
MarcinMr: I'm not sure the tooltip itself and the `display: inline-block;` cause the issue.

Here is an example of `display: inline;` We can see that the dotted line stretches from the start of the string to the end of the dot.

display-inline

Here is the `display: inline-block;` and the dotted line stretches from the start of the string to the end of the string. So this is how it should look like but instead of this whitespace before the dot.

display-inline-block

Is the tooltip itself causing the issue and not the abandoned dot? In the dev tools if you check the "this example" string, whitespace is only before that string. After the string, there is no whitespace. But the dot has whitespace in front of it and back of it.

tooltip whitespace

Look at this record. Changing the "this example" string doesn't cause anything, but deleting the whitespace from the dot fix the issue.

https://user-images.githubusercontent.com/85670248/163108972-a8056a6f-df21-4810-82cf-58ac260dea5c.mov

mikemai2awesome: This is a problem we need to solve, we have used `spaceless` elsewhere.

MarcinMr:
@danielamorse 

I used a few `<spaceless>` tag combinations and some indentation changing and that works but only for the dot at the end of the tooltip. There is still space in front of the tooltip when we use a special character.

It looks like this in the dev tools, there is no space around tooltip and around the dot, but still exist near special character in front of the tooltip string:
special sign whitespace

I tried to remove everything and reduce the code to one line for testing purposes like this:
spaceless one line doesnt work
I also tried adding `|spaceless` filter to the above code like this: `{{ trigger|spaceless }}`, but didn't help.

Tried also put there basic string:
whitespace test string
An the filter removes spaces around the tooltip but there is still whitespace near the special character:
white-space issue

I tried also with `|trim` but didn't help :(

The only thing that worked was using the `|spacless` filter but in the demo pages, with the combination of `display: inline-block` in CSS.
display-inline-block-spaceless-demo-doc

I don't understand why the `spaceless` filter works on the demo pages but when I added it in the template and reduce code to one line to be sure there are no spaces, that didn't help.

I'm still working on that and will try also tomorrow. Dan, do you have any more ideas for that?

docs-site/src/pages/pattern-lab/_patterns/40-components/tooltip/15-tooltip-inverted.twig
MarcinMr: Filter `|spaceless` works in demo pages but with a combination of `display: inline-block`. that filter doesn't work with `display: inline`.

danielamorse (changes requested):
@MarcinMr you've actually solved the issue we care about. There is no longer extra space _inside_ the tooltip [here](https://feature-ds-712-remove-twig-whitespace-add.boltdesignsystem.com/pattern-lab/?p=components-tooltip).

In the other examples you mention, the extra space is getting added _outside_ the tooltip when the variable is assigned.

```
{% set tooltip %}// This line-break is adding whitespace
  {# Start component specific code #}
  {% include '@bolt-components-tooltip/tooltip.twig' with {
    trigger: 'this example',
    content: 'This is an inverted tooltip.',
    inverted: true
  } only %}
  {# End component specific code #}
{% endset %}
```

You'd have to do this:

```
{% set tooltip %}{% spaceless %}
  {# Start component specific code #}
  {% include '@bolt-components-tooltip/tooltip.twig' with {
    trigger: 'this example',
    content: 'This is an inverted tooltip.',
    inverted: true
  } only %}
  {# End component specific code #}
{% endspaceless %}{% endset %}
```

There is nothing you can do to trim that space. You must use `{% spaceless %}` when you assign the variable,  `|spaceless` when you print the variable, or simply don't add a line-break in the first place. _That_ space is out of our control (aside from [hacks we don't want to employ](https://css-tricks.com/fighting-the-space-between-inline-block-elements/) here). We must manually trim it in our demos and we should probably add a note saying something like: "When adding an inline tooltip be sure to trim any unwanted space around the element."

Thanks for adding the test page. Would you please update that test page and the other demo you modified to use `{% spaceless %}` and add a comment to the basic Tooltip demo explaining how you must trim the whitespace around the tooltip?

If you have any questions we can chat tomorrow 🙏

danielamorse:
@MarcinMr I also noticed that space is added by the top of the Tooltip Twig template [where we assign variables](https://github.com/boltdesignsystem/bolt/blob/bee535c7e333201fea6a0772fbcc831f90b196c2/packages/components/bolt-tooltip/src/tooltip.twig#L1-L58). To eliminate that space we'd either have to do something like this:

```
{% spaceless %}
{% set schema = bolt.data.components["@bolt-components-tooltip"].schema %}
...
```

Or just recommend that anytime you use a Tooltip inline that you must use spaceless in your variable assignment or variable output.

What do you think? Let's talk more tomorrow morning.

MarcinMr:
> @MarcinMr I also noticed that space is added by the top of the Tooltip Twig template [where we assign variables](https://github.com/boltdesignsystem/bolt/blob/bee535c7e333201fea6a0772fbcc831f90b196c2/packages/components/bolt-tooltip/src/tooltip.twig#L1-L58). To eliminate that space we'd either have to do something like this:
> 
> ```
> {% spaceless %}
> {% set schema = bolt.data.components["@bolt-components-tooltip"].schema %}
> ...
> ```
> 
> Or just recommend that anytime you use a Tooltip inline that you must use spaceless in your variable assignment or variable output.
> 
> What do you think? Let's talk more tomorrow morning.

I added your recommended solution but we still need to add a `spaceless` or `|spaceless` filter in demos, when we have a special character at the beginning of a tooltip.

> @MarcinMr you've actually solved the issue we care about. There is no longer extra space _inside_ the tooltip [here](https://feature-ds-712-remove-twig-whitespace-add.boltdesignsystem.com/pattern-lab/?p=components-tooltip).
> 
> In the other examples you mention, the extra space is getting added _outside_ the tooltip when the variable is assigned.
> 
> ```
> {% set tooltip %}// This line-break is adding whitespace
>   {# Start component specific code #}
>   {% include '@bolt-components-tooltip/tooltip.twig' with {
>     trigger: 'this example',
>     content: 'This is an inverted tooltip.',
>     inverted: true
>   } only %}
>   {# End component specific code #}
> {% endset %}
> ```
> 
> You'd have to do this:
> 
> ```
> {% set tooltip %}{% spaceless %}
>   {# Start component specific code #}
>   {% include '@bolt-components-tooltip/tooltip.twig' with {
>     trigger: 'this example',
>     content: 'This is an inverted tooltip.',
>     inverted: true
>   } only %}
>   {# End component specific code #}
> {% endspaceless %}{% endset %}
> ```
> 
> There is nothing you can do to trim that space. You must use `{% spaceless %}` when you assign the variable, `|spaceless` when you print the variable, or simply don't add a line-break in the first place. _That_ space is out of our control (aside from [hacks we don't want to employ](https://css-tricks.com/fighting-the-space-between-inline-block-elements/) here). We must manually trim it in our demos and we should probably add a note saying something like: "When adding an inline tooltip be sure to trim any unwanted space around the element."
> 
> Thanks for adding the test page. Would you please update that test page and the other demo you modified to use `{% spaceless %}` and add a comment to the basic Tooltip demo explaining how you must trim the whitespace around the tooltip?
> 
> If you have any questions we can chat tomorrow 🙏

@danielamorse thanks for explaining that. I also noticed that some of the comments could cause problems with whitespace, so I deleted them. I added a note in the tooltip docs and also updated the test page. Let me know if that's sufficient and if you spot anything there that is causing any issues related to whitespace.

danielamorse (approved):
Everything looks good now. Nice job on the docs!

PR #2477, opened 2022-04-12
DS-760 fix switch button label text

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-760

## Summary

Fixes a bug where the switch button's label text is announced twice.

## Details

1. Aria attributes added to prevent the redundant announcments
1. Doc page updated with better copy

## How to test

1. Run the branch locally
2. Go to the Switch Button doc
3. Use screen reader and tab to the switch button in the demo.
4. Press either enter or space
5. Listen to screen reader


danielamorse (approved):
@mikemai2awesome verified fixed. I noticed this only happens in Safari, not in Chrome.

PR #2479, opened 2022-04-15
[DO NOT MRGE] DS-761 replace image element in hero

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-761

## Summary

Removes image component from the hero component

## How to test

- Confirm no usages of the image component within the hero component
- Confirm no visual changes in the hero component examples in pattern lab


remydenton:
Known issue with spacing:

Screen Shot 2022-04-15 at 10 25 48 AM

packages/components/bolt-hero/hero.twig
```diff
@@ -38,17 +38,23 @@
         u-bolt-padding-top-{{_imagePaddingTop}}@{{_mediumBreakpoint}}
         u-bolt-padding-bottom-{{_imagePaddingBottom}}@{{_mediumBreakpoint}}
       ">
-      {% include "@bolt-components-image/image.twig" with {
-        src: this.data["image"].value,
-        srcset: this.data["image_srcset"].value ? this.data["image_srcset"].value : this.data["image"].value,
-        max_width: this.data["image_max_width"].value,
+
+      {% set style = 'min-width: ' ~ (this.data['image_min_width'].value ? this.data['image_min_width'].value : 'none') ~ ';' %}
```
mikemai2awesome, line 42: Do not render `min-width:none`, when there's no value, render nothing.

packages/components/bolt-hero/hero.twig
```diff
@@ -38,17 +38,23 @@
...
+      {% if this.data['image_max_width'].value %}
+        {% set style = style ~ 'max-width:' ~ this.data['image_max_width'].value ~ ';' %}
+      {% endif %}
+
+      {% include '@bolt-elements-image/image.twig' with {
         attributes: {
```
mikemai2awesome, line 48: As we discussed, this template is still blocking the user to pass width and height attributes to this image element, which will result in layout shift (image loading without height, stretches the layout when it has a height once loaded).

mikemai2awesome (changes requested):

colbytcook (approved):
@remydenton this looks good to me (baring Mikes comments). I noticed a bug that is similar but not related to this work (https://pegadigitalit.atlassian.net/browse/DS-766).

packages/components/bolt-hero/hero.twig
remydenton: I updated the hero component to accept props for image width and height.  Note that if you don't pass these attributes (which will be the case for existing content initially), it reverts to using the deprecated image component.  My intent is to remove that with the next major release.

mikemai2awesome (approved):

PR #2481, opened 2022-04-18
DS-540: scroll to top demo warning

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-540](https://pegadigitalit.atlassian.net/browse/DS-540)

## Summary

Demo was added with #2423, this PR is to add a warning about pattern-lab.

## Details

Add a warning that the smooth animation does not work inside the pattern-lab wrapper and that it should be opened in a new tab.

## How to test

Read the demo text added in PR.


colbytcook (approved):
Approved

PR #2483, opened 2022-05-04
DS-777 button element type validation

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-777

## Summary

Pilots a new strategy for validation in button elements

## Details

* Previously, when button type and href were both missing, no tag would be set and the button would render with a missing tag (the following screenshot shows two valid buttons followed by a button that is missing the type attribute):
Screen Shot 2022-05-03 at 12 17 20 PM
This PR adds a default type of `button` in that scenario.  If a button like this somehow makes it to prod, it will probably work as originally intended.
* However, to warn developers that they shouldn't omit the button type, if twig debugging is enabled (we use `enable_json_schema_validation` as a proxy), the button will be replaced with an error message telling them what's wrong.
Screen Shot 2022-05-04 at 2 17 11 PM

## How to test

1. Create a button with no type attribute in pattern lab:
```
{% include '@bolt-elements-button/button.twig' with {
    content: 'This is a button with no "type" attribute',
} only %}
```
2. Confirm that the button is replaced with a fat red error message as long as twig debugging is enabled (see screenshot in the `Details` section).

A similar warning should display if _both_ the `href` and `type` attributes are set.

If twig debugging is disabled (true in the case of the jest test snapshots added in this PR), the buttons should display normally, even though the data passed in is missing critical parts.

### DIFF
--- a/packages/elements/bolt-button/src/button.twig
+++ b/packages/elements/bolt-button/src/button.twig
@@ -8,7 +8,7 @@
  * Dev notes
  * 1. The markup is written in one line (combined with spaceless) to avoid any unwanted white-space.
  * 2. Zero width no-break space is passed to the icon container's :before and :after pseudo elements, combine that with nowrap white-space on the icon container, and it will make sure that an icon never wraps to the next line by itself, it will always wrap with the final word of the text. Do not remove the icon container spans. See: https://codepen.io/mikemai2awesome/pen/OJXbqwe?editors=0100
- * 3. If type is empty for a button, an error message will be shown in place of the button to remind the developer that type is required.  However, in a prod environment, we will set a default for them to try to recover.
+ * 3. If the type attribute is empty for a semantic <button>, an error message will be shown in place of the button to remind the developer that type is required. However, in a prod environment, we will set a default (type="button") for them to try to recover.
  */
 #}
 
@@ -18,11 +18,11 @@
   {{ validate_data_schema(schema, _self)|raw }}
 
   {% if attributes.href is empty and attributes.type is empty %}
-    {% set error_message = 'The type attribute is required for any button that is not a link'|t %}
+    {% set error_message = 'The type attribute is required for a semantic <button>, while the href attribute is required for a semantic <a>. Please pass one of those attributes.'|t %}
   {% endif %}
 
   {% if attributes.href and attributes.type %}
-    {% set error_message = 'The type attribute is not supported for buttons that are links'|t %}
+    {% set error_message = 'The type attribute should be used for a semantic <button>, while the href attribute should be used for a semantic <a>. If you have a reason for using both at the same time, please contact the core team.'|t %}
   {% endif %}
 {% endif %}
 
@@ -46,8 +46,14 @@
 ] %}
 
 {% if error_message %}
+  {% set message %}
+    {% include '@bolt-elements-icon/icon.twig' with {
+      name: 'close-circle',
+    } only %}
+    {{ error_message|e }}
+  {% endset %}
   {% include '@bolt-components-banner/banner.twig' with {
-    content: error_message,
+    content: message,
     status: 'error',
   } only %}
 {% else %}
--- a/packages/elements/bolt-button/src/button.twig
+++ b/packages/elements/bolt-button/src/button.twig
@@ -8,6 +8,7 @@
  * Dev notes
  * 1. The markup is written in one line (combined with spaceless) to avoid any unwanted white-space.
  * 2. Zero width no-break space is passed to the icon container's :before and :after pseudo elements, combine that with nowrap white-space on the icon container, and it will make sure that an icon never wraps to the next line by itself, it will always wrap with the final word of the text. Do not remove the icon container spans. See: https://codepen.io/mikemai2awesome/pen/OJXbqwe?editors=0100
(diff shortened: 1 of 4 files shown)


packages/elements/bolt-button/src/button.twig
```diff
@@ -8,19 +8,33 @@
...
  */
 #}
 
 {# Schema Validation #}
 {% set schema = bolt.data.components['@bolt-elements-button'].schema %}
 {% if enable_json_schema_validation %}
   {{ validate_data_schema(schema, _self)|raw }}
+
+  {% if attributes.href is empty and attributes.type is empty %}
+    {% set error_message = 'The type attribute is required for any button that is not a link'|t %}
```
mikemai2awesome, line 21: Update the copy: "The type attribute is required for a semantic `<button>`, while the href attribute is required for a semantic `<a>`. Please pass one of those attributes."

packages/elements/bolt-button/src/button.twig
```diff
@@ -8,19 +8,33 @@
...
+  {% endif %}
+
+  {% if attributes.href and attributes.type %}
+    {% set error_message = 'The type attribute is not supported for buttons that are links'|t %}
```
mikemai2awesome, line 25: Update the copy: "The type attribute should be used for a semantic `<button>`, while the href attribute should be used for a semantic `<a>`. If you have a reason for using both at the same time, please contact the core team."

packages/elements/bolt-button/src/button.twig
```diff
@@ -31,6 +45,13 @@
   icon_only ? 'e-bolt-button--icon-only',
 ] %}
 
+{% if error_message %}
+  {% include '@bolt-components-banner/banner.twig' with {
+    content: error_message,
```
mikemai2awesome, line 50: Let's add the `close-circle` icon in front of the text.

packages/elements/bolt-button/src/button.twig
```diff
@@ -8,19 +8,33 @@
  * Dev notes
  * 1. The markup is written in one line (combined with spaceless) to avoid any unwanted white-space.
  * 2. Zero width no-break space is passed to the icon container's :before and :after pseudo elements, combine that with nowrap white-space on the icon container, and it will make sure that an icon never wraps to the next line by itself, it will always wrap with the final word of the text. Do not remove the icon container spans. See: https://codepen.io/mikemai2awesome/pen/OJXbqwe?editors=0100
+ * 3. If type is empty for a button, an error message will be shown in place of the button to remind the developer that type is required.  However, in a prod environment, we will set a default for them to try to recover.
```
mikemai2awesome, line 11: Update copy: "If the type attribute is empty for a semantic <button>, an error message will be shown in place of the button to remind the developer that type is required. However, in a prod environment, we will set a default (type="button") for them to try to recover."

I think I need some explanation on how this works, so we can address @danielamorse's concern about the error message showing up unexpectedly on live pages.

mikemai2awesome (changes requested):
The template looks good, just a few minor adjustments.

packages/elements/bolt-button/src/button.twig
remydenton: I've updated the copy.

To explain a little about @danielamorse's concern, the key logic is this:
```
  ...
  {% set error_message = 'Some message' %}
```

My original assumption before looking into this was that `enable_json_schema_validation` would be false if twig debug mode was set to TRUE.  On further investigation though, that appears to not be the case, and this variable is basically [hard-coded to true](https://github.com/boltdesignsystem/bolt/blob/master/packages/twig-integration/twig-extensions-shared/src/TwigExtensions/BoltCore.php#L40).  I pushed an update to this PR that would update that, but it probably means this should wait for a major release to ensure sites actually pull it in before potentially breaking their buttons.

I think it's still worth getting some reviews and opinions on whether this kind of validation is a good direction to begin with.

adamszalapski (approved):
From the user's/developer's perspective should we have this error above the button?

colbytcook (approved):

remydenton:
I've moved the error after the button instead of replacing the button with the error, as it was doing previously.  Thanks for the feedback on that guys.

@danielamorse `isDebug` is a method on all Twig environments and isn't specific to Bolt.  There's [Drupal documentation](https://api.drupal.org/api/drupal/vendor%21twig%21twig%21lib%21Twig%21Environment.php/function/Twig_Environment%3A%3AisDebug/8.2.x) on it, and [the twig docs ](https://twig.symfony.com/doc/3.x/api.html#environment-options) discuss the `debug` boolean when initializing a Twig environment object.  

Pattern lab sets this this with its own `twigDebug` configuration, which we set to `true` here: https://github.com/boltdesignsystem/bolt/blob/master/docs-site/config/config.yml#L56.  I'm not sure if/how it gets set to `false` for production builds, but in theory it should be.  Originally added in d82d4756e69 ("back out code changes probably not needed" 🤦 ).

TLDR, this is intended to work in both Bolt and Drupal.

danielamorse (approved):
Thanks for making the update @remydenton.

It looks like `twigDebug` does not actually get set to `false` in Patternlab prod mode. I tested by running `yarn build && yarn serve` and saw the button error message there. Not ideal, but not something I'd bother fixing unless it's a trivial change.

:shipit:

PR #2494, opened 2022-05-24
DS-782 Refactor dialog close button

mikemai2awesome:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-782

## Summary

Refactored dialog's close button and added the `transparent` prop.

## Details

1. The close button is positioned using CSS grid rather than absolute position.
1. `transparent` prop added to enhance image/video only dialogs.
1. Revised all dialog docs.

## How to test

Run the branch locally and go through all the dialog docs. Proofread everything and test to see if all props work as intended.


mikemai2awesome:
Need help looking into why `esc` is not closing dialog anymore, and tooltips/popover inside dialog don't show up.

@danielamorse @cjwhitedev

danielamorse (approved):
@mikemai2awesome really nice job on the docs. They are clear and well-organized. Also, like the grid solution for the close button problem 🎉 

I fixed Popover and Tooltip by appending to the `<dialog>` itself.


Approved... just one minor comment: it looks like the spacing "none" class has no styles associated with it. We are still outputting the spacing "none" class in the Twig. Should we omit it? It's not doing any harm so I'm fine leaving it, but wanted to point that out.

mikemai2awesome:
@danielamorse Thanks for pointing that out. I actually found a lot of typos with the modifier classes. All fixed now.

packages/components/bolt-dialog/src/dialog.scss
```diff
@@ -116,49 +150,49 @@
...
     inline-size: 75ch;
   }
 
-  .c-bolt-dialog__width-full {
+  .c-bolt-dialog--width-full {
     inline-size: calc(100% - var(--bolt-spacing-x-medium) * 2);
   }
 }
 
 // Container spacing options
```
cjwhitedev, line 166: Should we add spacing for the close button here too? Just to keep the button inline with the content on the right side?

mikemai2awesome: I played around with this one. Only the large spacing feel a bit odd, but if I add the large spacing, it creates more problem:

- Proximity of the close button would be drastically off
- The huge chunk of spacing would squeeze the heading text too much

It's better to not move the close button.

cjwhitedev (approved):

PR #2495, opened 2022-05-26
DS-423: Typeahead Button only Clickable When Menu is Open

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-769](https://pegadigitalit.atlassian.net/browse/DS-769)

## Summary

This PR is to fix the typeahead button still being clickable even when the menu is closed.

## Details

The submit button was still clickable even when the menu was closed. This moves the `pointer-events: all;` line to only be applicable when the menu is open.

## How to test

Follow the steps-to-reproduce shown in the ticket.

### DIFF
--- a/packages/components/bolt-page-header/src/_page-header-desktop.scss
+++ b/packages/components/bolt-page-header/src/_page-header-desktop.scss
@@ -79,6 +79,11 @@
       pointer-events: all;
     }
 
+    @at-root .c-bolt-page-header__action-trigger--search[aria-expanded='false']
+        ~ #{&} {
+      pointer-events: none;
+    }
+
     &:before {
       content: '';
       opacity: 0.8;
--- a/packages/components/bolt-page-header/src/_page-header-desktop.scss
+++ b/packages/components/bolt-page-header/src/_page-header-desktop.scss
@@ -79,11 +79,6 @@
       pointer-events: all;
     }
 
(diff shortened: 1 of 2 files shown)


mikemai2awesome (approved):

danielamorse (changes requested):
I can still click on the "clear text" button:

https://user-images.githubusercontent.com/36336/170712045-a8288eec-91c0-4ebc-9c05-2b07cebf9f24.mov

Would you please look into that?

mikemai2awesome (changes requested):
@cjwhitedev This only takes care of mouse pointer events, the clear button can still be clicked on/focused on via screen reader after search panel is collapsed. Do we have a more permanent solution?

cc @danielamorse

danielamorse:
@mikemai2awesome @cjwhitedev this should be fixed now.

We were using the `visibility` CSS property to hide/show the "Clear" button. When you set `visibility: visible` screen readers will speak the element. This happens even if the container is set to `visibility: hidden`. There seemed to be no downside to using `display` instead of `visibility` to hide/show the "Clear" button, so that's what I did.

Once I made that change, I could also remove part of the CSS @cjwhitedev had to add earlier.

This should solve it for mouseover, focus, and screen readers. Please review to verify 🙏

mikemai2awesome (approved):
This works now! @danielamorse

danielamorse (approved):

PR #2497, opened 2022-06-21
DS-771 Carousel breaks inside layout grid

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-771

## Summary

The carousel placed inside `.l-bolt-site` should stay centered without left or right misaligning.

## Details

The carousel should have an active slide centered horizontally in the viewport in all carousel variations such as these below:

- buttons inside
- buttons outside
- buttons inside with a property `overflow` set to `true`
- buttons outside with a property `overflow` set to `true`
- all slides per view examples (with buttons inside and outside, with and without `overflow` prop)

## How to test

Pull the beach. Go to the `/pattern-lab/?p=tests-carousel-in-layout` and see the carousel variations. Check if the carousel's active slide is always centered.


danielamorse (approved):
@MarcinMr I think you found the root issue, nicely done! It seems there were rules applying only to "no-overflow" that should've applied to both "overflow" and "no-overflow".

I used the test page to reproduce the old bug and to verify the fix 🎉

PR #2501, opened 2022-07-06
DS-796 Remove Ratio Object

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-796

## Summary

Ratio object is removed from the `global > styles > 05-objects folder`.

## Details

Referring to the previously [research ticket](https://pegadigitalit.atlassian.net/browse/DS-652) we don't need the ratio object anymore.

## How to test

Pull the branch. Check if there's no ratio object in the Bolt repository and usage of that.

## Release notes

The ratio object is no longer in use. We have the old `ratio-component` and the new `ratio-element`.


colbytcook (approved):

colbytcook:
@remydenton Adding you to this PR for review. Currently there is a [couple places](https://gitlab.com/search?group_id=5804310&repository_ref=&scope=blobs&search=o-bolt-ratio&snippets=false) where this object is being used on the Drupal side so I wanted to make sure we shouldn't have this be merged into 6.0 instead of master since it could be a breaking change?

remydenton:
Thanks @colbytcook.  It looks like the places it's still in use in Drupal are all non-functional (a hack for the hapyak player which hasn't been used in years and a file for search summary which was only ever used for demos, and is not a good example anyway).

That said, I think for consistency and to not be jerks by knowingly removing things that once worked without telling anyone, we should merge this into 6.0 instead.

PR #2515, opened 2022-07-14
DS-799 Object Inline List deprecation

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-799

## Summary

Usage of `o-bolt-inline-list` were replaced with bolt-component `List` as was recommended in the [research ticket](https://pegadigitalit.atlassian.net/browse/DS-652)

## Details

Places where `inline-list object` was updated with Bolt-component `List`:

- form-icons.twig
- event-detail-twig
- 05-footer.twig

## How to test

Check if there are no regressions due to the replacement `inline-list object` with the `List` component.

## Release notes

Object inline-list is deprecated. Please use the Bolt component `List`.


packages/components/bolt-form/src/form.scss
```diff
@@ -414,7 +414,7 @@ $bolt-input-transition: var(--bolt-transition);
   top: 1px;
   right: 1px;
   padding-top: calc(
-    var(--bolt-spacing-y-medium) / 2 - 0.25rem
+    var(--bolt-spacing-y-medium) / 2.5 - 0.25rem
```
colbytcook, line 417: @MarcinMr is this change intended? I don't see any mention of it in the description so I wanted to double check.

colbytcook (approved):
@MarcinMr approved with one minor comment

packages/components/bolt-form/src/form.scss
MarcinMr: @colbytcook After replacing `o-bolt-list` with the `list` component I saw a small vertical misalignment of icons inside form input, this is why I adjusted it a little bit.

danielamorse: @MarcinMr minor suggestion: since we're already using one "magic number" here, `.25rem`, I would leave the divisor as `2` and tweak the `.25rem` value as needed.

Also, see Mike's comment on the next line and please update the comment with the new value. You can test the Japanese alignment by changing the `[lang]` tag on the `<html>` element to `lang="ja"`.

danielamorse (changes requested):

packages/components/bolt-form/src/form.scss
MarcinMr: @danielamorse thanks for the suggestion, I updated the numbers and comment

MarcinMr: **Vertical alignment before changes**
Old number

**Vertical alignment after changes**
New number

danielamorse (approved):
Nice, thanks @MarcinMr!

PR #2525, opened 2022-08-11
[DO NOT MERGE] Table Test Page

colbytcook:
## Jira

https://pegadigitalit.atlassian.net/browse/CC-1017

## Summary

Added a test page to test out a table fix that will prevent the table from locking the column widths

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/999-tests/table/00-table-column-cc-test.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/999-tests/table/00-table-column-cc-test.twig
@@ -1,0 +1,68 @@
+{% set main_content %}
+  <style>
+    .table-layout-auto { table-layout: auto; }
+    .cc-post { word-break: break-word; }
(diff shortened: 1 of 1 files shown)


danielamorse (commented):
@colbytcook, unsetting the word-break seems to get you the desired column width (when combined with the layout-fixed class).

However, the responsive behavior is not working in this demo. It seems to break as soon as you put the table into the Layout component.

I suspect it's something to do with Layout using `display: grid`. Does the responsive table actually work inside Layout? @MarcinMr we ran into something like this with Carousel. Any insights?

MarcinMr:
@colbytcook @danielamorse 

Can we use the attributes object and add there an inline style to make responsivness back?

```
{% include '@bolt-layouts-site/site.twig' with {
  main: {
    content: main_content,
    attributes: {
      style: 'min-width: 0;'
    }
  }
} only %}
```

colbytcook:
@MarcinMr @danielamorse thanks for the review but I am going to close this PR for now, we are going to explore another option.

PR #2529, opened 2022-08-18
DS-733: make tabs show-more button invisble by default

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-733](https://pegadigitalit.atlassian.net/browse/DS-733)

## Summary

Make the tabs show-more button invisble on page-load to prevent an unstyled flash

## Details

Make the tabs show-more button invisble on page-load to prevent an unstyled flash

## How to test

refresh a page with tabs on it, confirm that you are no longer seeing the more button flash before hiding again.

### DIFF
--- a/packages/components/bolt-tabs/src/tabs.js
+++ b/packages/components/bolt-tabs/src/tabs.js
@@ -240,13 +240,12 @@ class BoltTabs extends withContext(BoltElement) {
     // If button is hidden, we need to show it before we get the width.
     // Explicitly check for 'is-hidden', as we'll be re-adding the class when we're done.
     if (this.showMoreItem.classList.contains('is-hidden')) {
+      this.showMoreItem.classList.add('is-invisible');
       this.showMoreItem.classList.remove('is-hidden');
       width = this.dropdownButton.offsetWidth;
       this.showMoreItem.classList.add('is-hidden');
+      this.showMoreItem.classList.remove('is-invisible');
     }
-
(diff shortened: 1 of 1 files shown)


MarcinMr (approved):
There isn't any flash of the `more` button when refreshing. I tested also with the dev tools options network fast 3G  and slow 3G, seems like it works nice now.

packages/components/bolt-tabs/src/tabs.js
```diff
@@ -593,7 +594,13 @@ class BoltTabs extends withContext(BoltElement) {
 
     const dropdown = () => {
       return html`
-        <div class="${cx('c-bolt-tabs__item', 'c-bolt-tabs__show-more')}">
+        <div
+          class="${cx(
+            'c-bolt-tabs__item',
+            'c-bolt-tabs__show-more',
+            'is-invisible',
```
danielamorse, line 601: @cjwhitedev thinking about how this works, it would be better if we simply set `is-hidden` here and then left lines 243-247 as is. That way, any time we check the width of the button we also make it invisible, not just on load.

danielamorse (changes requested):
@cjwhitedev would you please try out my suggestion and let me know if it works?

MarcinMr (approved):
Seems like still works as expected

danielamorse (approved):

danielamorse (approved):

PR #2530, opened 2022-08-19
[6.x]: DS-823 add examples for linear and radial gradients backgrounds

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-823

## Summary

After the discussion, we decide to move ahead and deprecate a few of the background props.

## Details

- `fill_color` prop is deprecated
- `color` enum of the `fill` prop was renamed to `solid`
- `linear-gradient` and `radial-gradient` enums were deprecated (`solid` and `gradient` can be used)

**A solid overlay on the background component can be achieved by adding theming class `t-bolt-*` to the parent of a background component. The gradient option doesn't have support for this yet.**

## How to test

Pull the branch. Check if props are properly deprecated or renamed. Confirm that achieving solid overlay can be done by passing the theme class to the parent of the background component.

### Fill Prop Changes
- The `fill` prop option `color` was renamed to `solid`.
- The `fill` props options `linear-gradient` and `radial-gradient` were removed. These are now handled with a new `.u-bolt-bg-gradient` class, refer to the [Background Gradients](https://boltdesignsystem.com/pattern-lab/?p=viewall-visual-styles-background-gradients) docs page for more information.
- The `fill_color` and `fillColor` props are now removed from the schema. Solid overlay on the background component can be achieved by adding a [theming class](https://boltdesignsystem.com/pattern-lab/?p=visual-styles-theming-system) t-bolt-* to the parent of a background component.

### Shape Prop Changes
- The `shapeGroup` prop are now removed from the schema. Use `shape_group` instead.
- The `shapeAlignment` prop are now removed from the schema. Use `shape_alignment` instead.

### Overlay Prop Changes
- The `overlay` prop is now removed from the schema. Use the `opacity` prop instead.
- Change the `overlay: enabled` prop to `opacity: medium`
- Change the `overlay: disabled` prop to `opacity: none`

### Focal Point Prop Changes
- The `focalPoint` prop is now removed from the schema. Use the `focal_point` prop instead.

### Content Items Props Changes
- The `contentItems` and `content_items` props are now removed from the schema. Use the `items` prop instead (each item must be renderable, e.g. a string or render array. Arrays intended to be rendered with `pattern_template()` are no longer supported).


colbytcook (approved):

andrewmriv (approved):

packages/components/bolt-background/src/background-overlay.twig
```diff
@@ -5,6 +5,15 @@
   {% set focal_point = focalPoint %}
 {% endif %}
 
+{# DEPRECATED. Solid overlay on the background component can be achieved by adding theming class to the parent of a background component. #}
+{% if fill_color %}
+  {% if fill == 'solid' %}
+    {% set fill = 'solid' %}
+  {% elseif fill == 'gradient'  %}
+    {% set fill = 'gradient' %}
```
remydenton, line 13: I don't see a `fill` value used anywhere lower down in this file.  Was this intended to replace `this.data.fill.value`?  Note that logic on lines 21 and 22 depends on that value, but not `fill`.

I also don't really understand the logic here.  If the deprecated fill_color is present then... set fill to `solid` if it's already `solid`, or fill to `gradient` if it's already `gradient`?  Seems like it isn't doing anything that wouldn't already happen if this section were removed.

packages/components/bolt-background/src/background.scss
```diff
@@ -269,55 +269,59 @@ bolt-background {
   background-color: transparent !important;
 }
 
-.c-bolt-background__overlay--radial-gradient-fill {
```
remydenton, line 272: I found a couple places in the wild (on Academy) where radial gradient is currently in use.  Do you know what will happen to these when we remove this CSS?

- https://gitlab.com/pegadigital/pa/pegaacademy/-/blob/release/release-67/docroot/themes/custom/pegaacademy_theme/templates/layout/page--layout--landing-page.html.twig#L33
- https://gitlab.com/pegadigital/pa/pegaacademy/-/blob/release/release-67/docroot/themes/custom/pegaacademy_theme/templates/layout/page--layout--split-header-landing-page.html.twig#L33

remydenton (changes requested):
Overall, nice work on updating the schema and removing deprecated examples.  I have a few inline questions/comments to address.

packages/components/bolt-background/src/background-overlay.twig
MarcinMr: @remydenton 
I messed something up here...

At first, I thought to do it like this:

```
{% if fill_color %}
  {% set fill = fill_color %}
{% endif %}
```

But when someone uses `fill_color` and passes for example `pink` this will create a class `c-bolt-background__overlay--pink-fill` which we haven't got in our CSS.

Now I was thinking to add code like this:

```
{% if fill_color == 'default' or fill_color == 'navy' or fill_color == 'pink' or fill_color == 'black' %}
  {% set fill = 'solid' %}
{% endif %}
```

So whenever someone wants to use deprecated `fill_color` with its values, it automatically sets to `fill: 'solid'`.

I dont use `'c-bolt-background__overlay-fill-color--' ~ this.data.fill.value` because this is used for `linear-grdient` and `radial-gradient` which are removed now.

In relation to commented CSS, I added 

```
{% if fill == 'linear-gradient' or fill == 'radial-gradient' %}
  {% set fill = 'gradient' %}
{% endif %}
```

So whenever `radial-gradient` is used it will be overwritten to `gradient`, so that CSS can be commented/removed because `radial-gradient `will be replaced by a normal `gradient` for places like on Academy?

Does it make any sense or this should be done yet otherwise?

packages/components/bolt-background/src/background.scss
MarcinMr: I added 

```
{% if fill == 'linear-gradient' or fill == 'radial-gradient' %}
  {% set fill = 'gradient' %}
{% endif %}
```

So whenever `radial-gradient` is used it will be overwritten to `gradient`, so that CSS can be commented/removed because `radial-gradient `will be replaced by a normal `gradient` for places like on Academy?

Does it make any sense or this should be done yet otherwise?

packages/components/bolt-background/src/background-overlay.twig
```diff
@@ -6,14 +6,16 @@
...
-    {% set fill = 'gradient' %}
-  {% endif %}
+{% if fill_color == 'default' or fill_color == 'navy' or fill_color == 'pink' or fill_color == 'black' %}
+  {% set fill = 'solid' %}
 {% endif %}
 
+{# DEPRECATED. Whenever `linear-grdient` or `radial-gradient` is passed to the `fill` prop, set it to fill: 'gradient'. #}
+{% if fill == 'linear-gradient' or fill == 'radial-gradient' %}
+  {% set fill = 'gradient' %}
+{% endif %}
```
remydenton, line 16: I think I gave you bad advice here.  It seems that the example page I found that is theoretically using a radial gradient (https://academy.pega.com/exam/pega-certified-senior-system-architect-2) isn't actually working.  It looks like it's solid:
Screen Shot 2022-09-15 at 2 09 25 PM

And if I change the fill to `gradient`, it changes, probably not in an expected way:
Screen Shot 2022-09-15 at 2 10 15 PM

In Slack, we discuss maybe just making the fallback a solid color.

packages/components/bolt-background/src/background-overlay.twig
```diff
@@ -6,14 +6,16 @@
...
 
 {# DEPRECATED. Solid overlay on the background component can be achieved by adding theming class to the parent of a background component. #}
-{% if fill_color %}
-  {% if fill == 'solid' %}
-    {% set fill = 'solid' %}
-  {% elseif fill == 'gradient'  %}
-    {% set fill = 'gradient' %}
-  {% endif %}
+{% if fill_color == 'default' or fill_color == 'navy' or fill_color == 'pink' or fill_color == 'black' %}
+  {% set fill = 'solid' %}
```
remydenton, line 10: This still seems off to me.  `fill_color` will be passed when someone is expecting, for example, the overlay to be pink.  Just making the fill solid won't accomplish that (the user will need to add a theming class now instead).

I don't have any examples in the wild, but it just doesn't feel quite right.  Let's discuss in dev huddle.

packages/components/bolt-background/src/background-overlay.twig
MarcinMr: @remydenton 

I updated the code, could you check once again if there is a `radial-gradient` the fallback will be `gradient`?
I added back `radial-gradient` to the schema for testing and I got something like this:

gradient

Then I deleted the `radial-gradient` prop from the schema again so it won't be working now.

packages/components/bolt-background/src/background-overlay.twig
MarcinMr: @remydenton @colbytcook 

I thought that after deprecation the `fill_color` it is a must to add theming class to a parent based on the point 3 from here --> https://pegadigitalit.atlassian.net/browse/DS-244 . And I chose the option of adding the theme class to the parent to achieve overlay, and not update the `fill_color` to support theme classes

MarcinMr:
@colbytcook @remydenton I pushed some changes and comment Remy's remarks. I read the Slack discussion and comment in the thread. I don't know what we should do with this?. Leave this component as it is without merging this PR and focus on creating the new BG element? 

Let me know what you think

PR #2531, opened 2022-08-19
DS-839: Add Edit Button to Profile Avatar

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-839](https://pegadigitalit.atlassian.net/browse/DS-839)

## Summary

Add Edit Button to Profile Avatar.

## Details

This adds a conditional prop to the bolt-profile component, `avatar_editable`. Setting it to try and hovering over the avatar will show an edit button.

## How to test

Check the new doc page in the profile section of the doc site.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/40-components/profile/15-profile-editable-avatar.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/40-components/profile/15-profile-editable-avatar.twig
@@ -1,33 +1,19 @@
 {% set notes %}
     <bolt-ol>
-        <bolt-li></bolt-li>
+        <bolt-li>This is a conditional property, set `avatar_editable: true`, then <em>hover</em> to see the edit button.</bolt-li>
     </bolt-ol>
 {% endset %}
 
 {% set demo %}
-<bolt-list display="flex" spacing="small">
-    <bolt-list-item>
-        {% include '@bolt-components-profile/profile.twig' with {
-            name: {
-                content: '<strong>First Last</strong> (@username)',
-                tag: 'h2',
-            },
-            job_title: 'Lead System Architect',
-            location: 'United States',
-        } only %}
-    </bolt-list-item>
-    <bolt-list-item>
-        {% include '@bolt-components-profile/profile.twig' with {
-            name: {
-                content: '<strong>First Last</strong> (@username)',
-                tag: 'h2',
-            },
-            job_title: 'Lead System Architect',
-            location: 'United States',
-            avatar_editable: true,
-        } only %}
-    </bolt-list-item>
-</bolt-list>
+    {% include '@bolt-components-profile/profile.twig' with {
+        name: {
+            content: '<strong>First Last</strong> (@username)',
+            tag: 'h2',
+        },
+        job_title: 'Lead System Architect',
+        location: 'United States',
+        avatar_editable: true,
+    } only %}
 {% endset %}
 
 {% set twig_markup %}{% verbatim %}
--- a/docs-site/src/pages/pattern-lab/_patterns/40-components/profile/15-profile-editable-avatar.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/40-components/profile/15-profile-editable-avatar.twig
@@ -1,57 +1,57 @@
 {% set demo %}
-    {% set icon_pencil %}
-        {% include '@bolt-elements-icon/icon.twig' with {
-            name: 'pencil',
-        } only %}
-    {% endset %}
-    {% set edit_button %}
-        {% include '@bolt-elements-button/button.twig' with {
-            content: 'Edit Profile Image',
-            border_radius: 'full',
-            icon_only: icon_pencil,
-            size: 'large',
-            attributes: {
-                type: 'button'
-            }
-        } only %}
-    {% endset %}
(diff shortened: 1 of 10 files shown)


colbytcook (changes requested):
@cjwhitedev I think the hover animation still seems off (we have discussed this previously), fir you confirm this was due to the circle being behind the circle or the scoped element for hover?
Additionally, the behavior when you click is a little jarring. Currently when you click on the edit button it abruptly inverts with the background overlay. Also, icon color does not change so we are left with a white font on light grey background. Could we investigate making this alittle more subtle? The currently click animation of the [Button Element](https://boltdesignsystem.com/pattern-lab/?p=elements-button) just reverts the slight animation adjustment of the button. Is this something we can adjust?

docs-site/src/pages/pattern-lab/_patterns/40-components/profile/15-profile-editable-avatar.twig
```diff
@@ -0,0 +1,57 @@
+{% set demo %}
```
danielamorse, line 1: Is this 4-space indentation? Would you please check your VS Code settings if so?

packages/components/bolt-profile/profile.schema.js
```diff
@@ -12,6 +12,11 @@ module.exports = {
       type: 'any',
       description: 'Render a custom avatar image.',
     },
+    avatar_edit: {
```
danielamorse, line 15: @cjwhitedev since we want to tightly control the style of the button, I think this would make sense as a sub-template with its own schema. Then you could have props for `content` and `attributes` and the user wouldn't have to build the rest of the button themselves.

packages/components/bolt-profile/src/profile.scss
```diff
@@ -87,6 +87,57 @@
...
+      z-index: -1;
+      background-color: var(--bolt-color-white);
+      transition: opacity var(--bolt-transition);
+    }
+
+    &:hover,
+    &:focus {
+      > * {
+        opacity: 1;
+        transform: none;
```
danielamorse, line 133: Is `transform: none` to prevent the button from sliding up on hover? If so, would you please add a comment?

packages/components/bolt-profile/src/profile.scss
```diff
@@ -87,6 +87,57 @@
...
+    left: 0;
+    transform: translate3d(0, 0, 0);
+
+    > * {
+      opacity: 0;
+      position: static;
+      transform: none;
+      transition: opacity var(--bolt-transition);
+
+      &::after {
```
danielamorse, line 108: We usually do `&:after` instead of `&::after`.

danielamorse (changes requested):
@cjwhitedev let's talk more in the morning about maybe using a sub-template for the avatar instead. I think that would give us the control we need over the styling of this element and still give users the control they need to add attributes and content to the button.

MarcinMr (commented):
When it comes to the animation itself, maybe to make it smoother adding a `will-change: opacity` to the elements that have `transition: opacity var(--bolt-transition);` will help a little bit? Then worth checking if manipulating the speed of the transition and timing-function in this particular case will help. For example, override the `--bolt-transition-timing` like 350ms and maybe play with timing-function like `ease-in` or simply `linear`?

danielamorse (changes requested):
@cjwhitedev I created a separate branch with some minor edits. Please take a look and merge if it all works for you: `feature/DS-839-editable-profile-avatar--edits`. See my changes [here](https://github.com/boltdesignsystem/bolt/commit/44400997a66a25d034b26f57d26ade3b6f773654).

I renamed a few things for clarity and consistency:
- `@bolt-components-profile/profile-avatar-button.twig` => `@bolt-components-profile/profile-avatar-edit-button.twig`
- `avatar_button` => `avatar_edit_button`
- `profile-editable-avatar.twig` => `profile-avatar-edit-button.twig`

I also added a missing Drupal `|t` function and updated some copy.

Next steps:
- Any animation updates you had in mind
- Edit button should show on focus as well as on hover
- Add an initial class to the button and target that class instead of `> *`

danielamorse:
> When it comes to the animation itself, maybe to make it smoother...

@MarcinMr good suggestion. I tried this but `will-change` also affected the z-index stacking order which made it impossible to position the pseudo element below the button.

danielamorse:
@cjwhitedev @MarcinMr I pushed changes that smooth out the animation, [see commit](https://github.com/boltdesignsystem/bolt/pull/2531/commits/954f6562234422da31c629b231503665ff0f604a).

It uses the pseudo element to make the whole avatar hoverable and uses a sibling `<span>` to apply the animation. This was the only way I could get the animation to be smooth. In some cases it seems fine to animate an element _and_ its pseudo-element, and it runs smoothly. In this case, perhaps because of the z-index requirements, it gets choppy. So, this is a work around, but I'm ok with it.

Would you please review the code and cross-browser test? We will aim to release this on Monday 🙏

danielamorse (approved):

MarcinMr (approved):
I think this solution works well now :). The animation is smoother than before. I also tested using chrome tools. Slowing down the animation to 25% or 10% still looks smooth. Testing with dev tools performance monitor and rendering doesn't show any issues when hovering. Tested on Chrome, Safari, and FF.

PR #2534, opened 2022-08-30
DS-839: Fix avatar edit button template

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-839

## Summary

Fixes bug where button `type` is hardcoded and user classes are overwritten.

## Details

There was a regression in the avatar edit button template where `type` became hardcoded and user provided classnames were overwritten by base classnames. This PR fixes both and adds a test page verifying the required Drupal Ajax attributes will work when passed into the component.

## How to test

- Review the code
- Review the Test page


cjwhitedev (approved):
looks good to me!

PR #2536, opened 2022-08-31
DS-850: Fix edit button z-index bug

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-850

## Summary

Fixes bug where user can't hover over bottom half of the avatar edit button because the "action" block covers it up.

## Details

Fix by adding `z-index: 1` to avatar block.

## How to test

- Review code
- View demo


cjwhitedev (approved):
I can see the fix in action

PR #2541, opened 2022-09-13
[6.x] DS-244 cleanup and refactor background component

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-244

## Summary

Shema and docs were cleaned up. Most of the work was done in [DS-823](https://github.com/boltdesignsystem/bolt/pull/2530)

## Details

- Added CSS support for gradients (Gradients can be created by adding theming class to the parent of `background-componenet`)
- Background default `opacity` option is `none`
- `shape_group` and `shape_alignment` were marked as deprecated in the docs -->  `/pattern-lab/?p=components-background-shapes-variations`
- Added docs page "Background Theme Options Variations"

## How to test

Pull the branch. Confirm that achieving solid and gradient overlay can be done by passing the theme class to the parent of the background component. Review the docs.

## Release notes

- [DS-823](https://github.com/boltdesignsystem/bolt/pull/2530)
- `shape_group` and `shape_alignment` are deprecated
- gradient overlay can be achieved by adding one of the theming class to the parent of background component


danielamorse:
@remydenton @MarcinMr, I fixed the broken radial-gradient on Academy by adding the [missing support](https://github.com/boltdesignsystem/bolt/pull/2541/commits/33121c5410ab7c80a1032f281b9e17c2830c50be) for the deprecated prop `fillColor`. This is a production bug that only effects Academy and no one has reported it yet. We can probably just include this in the next minor release.

The broken Academy background image I mentioned in huddle was fixed by adding the [missing "pattern" prop](https://github.com/boltdesignsystem/bolt/pull/2541/commits/5cf8528a6741677688e7ddc0983f4fc93456f598) to the template. On the actual Academy site this "pattern" prop is present, so it actually _won't_ break there. I fixed it in Bolt just to avoid confusion (my own).

danielamorse:
@MarcinMr, @remydenton, now that I'm up to speed, my recommendation is to add support for `radial-gradient` and include an example in the docs. Let's discuss in huddle before starting work on it.

What is `linear-gradient` and how does it differ from `gradient`?

Some specific docs feedback:
- Under "Background Theme Colors Variations" let's just show two or three theme examples and link to [the docs](https://boltdesignsystem.com/pattern-lab/?p=visual-styles-theming-system).
- The Focal Point demo code uses `focal_point: { vertical: 'center' }` which is not a prop. Let's remove it.
  - Since horizontal is the only focal point axis, this whole prop should probably be rethought, but that's out of scope for this work.

I think it would be worth converting the Background docs to the latest format, like you see on [Notifications](https://boltdesignsystem.com/pattern-lab/?p=components-notifications). A bit of explanation for each prop would go a long way. While we're at it, we can remove "Variation" from every page title. Both can be done in a separate ticket since it's a decent amount of work.

For my own understanding, I jotted down some prop descriptions as a reviewed. Feel free to use this in the future when we update the Background docs:

- **Theme**: Theme colors are applied by adding a Bolt theme class to the background or its container. See the [theming system docs](https://boltdesignsystem.com/pattern-lab/?p=visual-styles-theming-system) for more info.
- **Opacity**: Opacity controls the layer of color between the background image and the content. The color itself is controlled by the theme class.
- **Focal Point**: Focal Point sets the horizontal alignment of the gradient (if applicable) to allow a point of interest in the background image to be seen.
- **Fill**: Fill sets the style of the layer of color between the background image and the content.
- **Valign**: Use the `--e-bolt-image-position` CSS var on the background image element to control its vertical alignment when the image overflows the background container. This cannot be used for horizontal alignment.

colbytcook (approved):

PR #2542, opened 2022-09-15
DS-837 animate on scroll

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-837

## Summary

JS triggers animations when they come within the viewport.

## Details

- After adding specific classes to wanted elements, animation triggers when elements are visible in the viewport. 
- When scrolling and elements are visible in the viewport they should be animated.

## How to test

Pull the branch. Go to the **Test Folder** with examples of animations. Confirm that elements appear when visible in the viewport and when scrolling. Check the cascade and no-cascade behavior.

## Release notes

The animation toolset is updated. Elements appear not only on the load event but they are observed. If they are on the viewport area the animation triggers.


danielamorse (approved):
@MarcinMr looks and works great.

FYI @andrewmriv and @remydenton (re: layout builder), animations will now start when an element is in the viewport rather than on load. If an element is in the viewport on load, it will animate immediately.

PR #2546, opened 2022-09-26
CC-959: Hide instead of destroy Tippy instance on disconnect

danielamorse:
## Jira

- https://pegadigitalit.atlassian.net/browse/DS-817
- https://pegadigitalit.atlassian.net/browse/CC-959

## Summary

Fixes a rare bug where a Popover/Tooltip inside of a Bolt List (or any other web component) fails to open.

## Details

When a Popover/Tooltip is inside another Bolt web component, and that other web component renders _after_ the Popover/Tooltip has rendered, it will trigger the Popover/Tooltip to disconnect. We are currently destroying the Tippy instance on disconnect. When the Popover/Tooltip reconnects, I expected it would reinitialize the Tippy instance, but it does not because the component's `firstUpdated()` lifecycle event only fires on the initial creation.

Instead of destroying the Tippy instance I am now just hiding it, which solves the original problem we sought to fix in https://github.com/boltdesignsystem/bolt/pull/2513, without creating this other bug.

## How to test
- Review code
- Open test page in PL: `/pattern-lab/?p=tests-popover-disconnected-callback`
- Replace the main Bolt List component JS file (`packages/components/bolt-list/index.js`) with this:
```
import { lazyQueue } from '@bolt/lazy-queue';

lazyQueue(['bolt-list'], async () => {
  // This forces the race condition to happen every time (for testing only!)
  setTimeout(async () => {
    await import(/*  webpackChunkName: 'bolt-list' */ './main');
  }, 100);
});
```
- Verify the Popover/Tooltip still works on the test page.

### DIFF
--- a/docs-site/src/pages/pattern-lab/_patterns/999-tests/popover/15-popover-disconnected-callback.twig
+++ b/docs-site/src/pages/pattern-lab/_patterns/999-tests/popover/15-popover-disconnected-callback.twig
@@ -1,5 +1,3 @@
-<p>When you click on the Popover or the Tooltip custom JS will remove the component from the DOM. This test verifies that the disconnected callback fires on removal and the Popover/Tooltip disappears and is destroyed.</p>
-
 {% set trigger %}
   {% include '@bolt-elements-button/button.twig' with {
     content: 'This triggers a popover on hover',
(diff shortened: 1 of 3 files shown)


adamszalapski (approved):
I can confirm, working as designed.

PR #2548, opened 2022-09-28
DS 835: Table Element

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-835

## Summary

The new element `table` is created that can replace the basic functionalities of the current Table Component

## Details
1. Much of the props and functionality are taken from the old table component, for example:

- only rows table
- table with top headers only
- table with side headers only
- table with both top and side headers
- caption of the table
- borderless table
- format table (regular, numeric)

2. The following methods of display

- Twig (appearance and behavior can be achieved via choosing props from the schema)
- HTML only (appearance and behavior can be achieved by adding appropriate classes to the `<table>`)
- HTML only (without the **tbody**) (appearance and behavior can be achieved by adding appropriate classes to the `<table>`)

3. JavaScript creates a wrapper and wraps the `<table>` if needed
4. There are props that set the first, second, or first two columns as fixed width
5. There is now a full width prop that will changed the default behavior (inline-block) to be full width
6. There are sticky header functionality now included

## How to test

(Focus on testing in all browsers)
* Pull the branch.
* Check the code, especially the CSS and JS. 
* Check the `table-element` docs.
* Go to test folder and find table test examples. 
* Play with them and confirm they work properly. 

## Release notes

There is a new Table Element that replaces the basic functionalities of the current Table Component


colbytcook (approved):

packages/components/bolt-table-basic/__tests__/table-basic.js
```diff
@@ -0,0 +1,206 @@
+import { render, stopServer } from '../../../testing/testing-helpers';
+import schema from '../table-basic.schema';
+const {
+  format,
+  borderless,
+  // eslint-disable-next-line camelcase
+  fixed_width_table,
```
adamszalapski, line 7: In other components, we use the camelCase approach, why we don't do this here?

packages/components/bolt-table-basic/__tests__/table-basic.js
```diff
@@ -0,0 +1,206 @@
...
+  // eslint-disable-next-line camelcase
+  fixed_width_columns,
```
adamszalapski, line 9: In other components, we use the camelCase approach, why we don't do this here?

packages/components/bolt-table-basic/src/table-basic.scss
```diff
@@ -0,0 +1,295 @@
...
+  thead td,
+  tfoot td {
+    padding: var(--bolt-spacing-y-small) var(--bolt-spacing-x-small);
+  }
+
+  td {
+    text-align: left;
+    vertical-align: top;
+
+    // &:first-child:empty ~ td {
```
adamszalapski: We should remove all commented/unused code

adamszalapski (approved):
Looks good, few comments but they are not mandatory.

docs-site/src/pages/pattern-lab/_patterns/20-elements/table/15-table-side-header.twig
```diff
@@ -0,0 +1,124 @@
+{% set notes %}
+{% spaceless %}
+  <bolt-ol>
+    <bolt-li>The Table Element is designed to work with both Twig and vanilla HTML (Twig is the recommended method).</bolt-li>
```
cjwhitedev, line 4: i dont think we need this line on every demo

docs-site/src/pages/pattern-lab/_patterns/20-elements/table/30-table-format-variations.twig
```diff
@@ -0,0 +1,213 @@
...
+  <bolt-ol>
+    <bolt-li>The Table Element is designed to work with both Twig and vanilla HTML (Twig is the recommended method).</bolt-li>
+    <bolt-li>Refer to the <a href="{{ link['elements-table-headers-footers'] }}" class="e-bolt-text-link">Table Element with Headers and Footer</a> page for information on adding the headers and footer.</bolt-li>
+    <bolt-li><strong>Twig:</strong> Use the <code>format</code> boolean prop to style the table in numeric format.</bolt-li>
+    <bolt-li><strong>HTML:</strong> Add a <code>e-bolt-table--numeric</code> class to the <code>&lt;table&gt;</code> to style the table in numeric format.</bolt-li>
+  </bolt-ol>
+{% endspaceless %}
+{% endset %}
+
+{% set demo %}
```
cjwhitedev, line 12: the demo isnt rendering for me, is anyone else having the same problem?

docs-site/src/pages/pattern-lab/_patterns/20-elements/table/35-table-borderless.twig
```diff
@@ -0,0 +1,124 @@
...
+        cells: [
+          'Optimus Prime',
+          'Bumblebee',
+        ]
+      }
+    },
+    rows: [
+      {
+        cells: [
+          'The awe-inspiring leader of the Autobot forces. Selfless and endlessly courageous, he is the complete opposite of his mortal enemy Megatron.',
```
cjwhitedev, line 33: this content seems weird for this demo, it doesnt show the removed borders as well as shorter content might

docs-site/src/pages/pattern-lab/_patterns/20-elements/table/45-table-column-fixed-width.twig
```diff
@@ -0,0 +1,395 @@
...
+    <bolt-li>The Table Element is designed to work with both Twig and vanilla HTML (Twig is the recommended method).</bolt-li>
+    <bolt-li>Refer to the <a href="{{ link['elements-table-top-header'] }}" class="e-bolt-text-link">Table Element with Top Header</a> page for information on adding a top header.</bolt-li>
+    <bolt-li><strong>Twig:</strong> Use the <code>fixed_width_columns</code> prop to adjust the column widths.</bolt-li>
+    <bolt-li><strong>HTML:</strong> Add a <code>e-bolt-table--fixed-width-columns--first</code>, <code>e-bolt-table--fixed-width-columns--second</code> or <code>e-bolt-table--fixed-width-columns--first-two</code> class to the <code>&lt;table&gt;</code> to adjust the column widths.</bolt-li>
+    <bolt-li>The <code>fixed_width_table</code> boolean prop will disable the <code>fixed_width_columns</code> column adjustments.</bolt-li>
+  </bolt-ol>
+{% endspaceless %}
+{% endset %}
+
+{% set demo %}
```
cjwhitedev, line 13: i cant see a difference between any of these demos until the last one.

cjwhitedev (changes requested):

danielamorse (changes requested):
As we discussed on Slack, we probably want to base the Table element on the work @MarcinMr did in https://github.com/boltdesignsystem/bolt/pull/2437 rather than the current Table component. That work has many nice improvements to CSS, the schema, etc. Since you're making a new "element" we don't have to worry about introducing breaking changes.

packages/components/bolt-table-basic/__tests__/table-basic.js
colbytcook: Same as the previous comment

docs-site/src/pages/pattern-lab/_patterns/20-elements/table/15-table-side-header.twig
colbytcook: Removed

docs-site/src/pages/pattern-lab/_patterns/20-elements/table/30-table-format-variations.twig
colbytcook: This should be fixed now

docs-site/src/pages/pattern-lab/_patterns/20-elements/table/45-table-column-fixed-width.twig
colbytcook: This is updated

cjwhitedev (commented):
- For the caption: is this what we want it to look like? it doesnt have a border in the component version. also the rounded corners in this version are abutted to the the table and i dont think thats what we're going for.
- the sticky header is not working for me in firefox dev, i know i saw it working on @colbytcook's screen share though.
- i like the red in the fix width section, just wondering if we should put a note that says its only there for demo purposes?

colbytcook:
@cjwhitedev Updated the "Fixed Width" page with the note.

danielamorse (changes requested):
@colbytcook, I reviewed the docs and noticed:
- Docs > Table Examples: The full-width tables aren't full width.
- Tests > Table Default html: The tables are full-width until the JS loads. I think @MarcinMr  works around this [on the 6.x branch](https://release-6-x.boltdesignsystem.com/pattern-lab/?p=components-table-column-width) by scoping the `width: 100%` rule to `.c-bolt-table > table`, or in your case `e-bolt-table-wrapper > .e-bolt-table`.

These are the only issues I found. Here are a couple other thoughts, not blockers:

**Full-width tables in wysiwyg**
If I'm a content author and I make a plain HTML table and I want it to be width 100% (and I'm not going to edit the raw markup), how do I do this? Since the "full-width" style comes from the wrapper and the wrapper is added by JS I don't think there's anything in place to support full-width tables for a non-power-user. To solve that we'd have to check for some additional class on the `<table>` itself. Is this a use case we are solving for? Any thoughts?

**Wrapper full-width classname**
To make the container full-width I'd actually go with `u-bolt-block`, like @MarcinMr did on the 6.x branch. I didn't realize that literally the only thing this wrapper had to do was set `display: block`. Alternatively, if you prefer a component class, maybe `e-bolt-table-wrapper--full-width` would tell the user more about what it does?

colbytcook:
@danielamorse I addressed the feedback you had

* Docs > Table Examples: the full width tables now render full width
* Tests > Table Default html: Updated the CSS to stop the flashes

Full Width Prop
I added a `full_width` prop into the schema that will do three things.
1. It will add a `e-bolt-table--full-width` class to the `<table>`. This class will apply a `width: 100%;` to the table to help prevent flash. 
2. The `table.twig` will now apply a `e-bolt-table-wrapper--full-width` class to the `e-bolt-table-wrapper` that will make the wrapper display block (and forcing it to be full width). 
3. The table JS will now poll the `<table>` for the `e-bolt-table--full-width` class and add the `e-bolt-table-wrapper--full-width` class to the wrapper when it is being rendered.

For WYSIWYG support, we can simply add a class (`e-bolt-table--full-width`) to the table element to achieve the desired effect.

I updated the existing docs pages with these changes (where applicable) and created a new documentation page for the full width prop. Let me know if you have any other questions regarding this.

cjwhitedev (commented):
- i think the PR description needs to be updated again, at least since adding the full_width prop, if not more than that.
- see inline comment as well

docs-site/src/pages/pattern-lab/_patterns/20-elements/table/00-table-docs.twig
```diff
@@ -0,0 +1,114 @@
+{% set usage %}{% verbatim %}
```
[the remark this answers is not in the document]
colbytcook: @cjwhitedev I don't mind this initial demo page being longer than Layouts because layout examples are larger scoped than a table.

colbytcook:
@cjwhitedev Updated the PR description and responded to your inline comment

cjwhitedev (approved):

packages/elements/bolt-table/table-cell.schema.js
```diff
@@ -0,0 +1,28 @@
...
+      description:
+        'Generates a table cell &lt;td&gt; element. To render a button which triggers expanding hidden rows, pass table-expand-button.twig as the content. To render a button which triggers sorting, pass table-sorting-button.twig as the content.',
+    },
+    header: {
+      type: 'boolean',
+      default: false,
+      description:
+        'Generates a table cell as a table header &lt;th&gt; element if set to true. &lt;th&gt; elements can be used in table head, table body, or table footer.',
+    },
+    filters: {
```
remydenton, line 22: Is `filters` a real prop?  I don't see it used in the twig template or any of the examples.

PR #2549, opened 2022-09-28
Update twig version requirement to not be hard-coded.

remydenton:
## Jira

N/A

## Summary

Changes fixed twig version requirement to range

## Details

Drupal is currently unable to update to a new version of twig since Bolt has a fixed version requirement.  This PR updates the composer requirement to be a semver range so that versions can be updated separately (within the same major version).

## How to test

Within the `docs-site` directory, run `composer update twig/twig`.  Confirm that a version later than 2.14.11 is installed.

### DIFF
--- a/docs-site/composer.json
+++ b/docs-site/composer.json
@@ -87,7 +87,6 @@
       "patternlab-plugin"
     ]
   },
-  "minimum-stability": "dev",
   "config": {
     "allow-plugins": {
       "cweagans/composer-patches": true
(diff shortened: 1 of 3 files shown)


colbytcook (changes requested):
@remydenton I cannot confirm that the latest version of Twig is being installed, i am still seeing `2.14.6`. Let me know if there is something else I should be doing to review this ticket besides what is in the testing instructions.

colbytcook (approved):
Approved, can confirm the twig version is updating appropriately

PR #2550, opened 2022-09-30
DS 833 list element

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-833

## Summary

A list element is created that will replace the list component.

## Details
- a new list element doesn't use JS
- all props will be CSS classes that will modify the markup to adjust the display appropriately
- the new element has a similar schema as the old list component
  -- the `flex` option was deleted from the `display` prop,
  -- the `inline` and the `inline@breakpoint` props were renamed to `horizontal` and `horizontal@breakpoint`,
  --  the `block` prop was renamed to `vertical`,

## How to test

Pull the branch. In the docs, there is various example of the list variations. Make sure that the new element behaves like the old list component.

## Release notes

There is a new List Element and The List Component was marked as deprecated.


docs-site/src/pages/pattern-lab/_patterns/20-elements/list/10-list-tag-variations.twig
```diff
@@ -0,0 +1,90 @@
...
+    } only %}
+  {% endset %}
+
+  {% for tag in schema.properties.tag.enum %}
+    <h3>{{ tag|capitalize }}</h3>
+    {% include '@bolt-elements-list/list.twig' with {
+      tag: tag,
+      items: list_items
+    } only %}
+  {% endfor %}
```
colbytcook, line 51: @MarcinMr Should we use a better example for the `ol`? Something that has an order to it?

MarcinMr:
I'm wondering if we should have examples of all `inline@breakpoint` (xxsmall, xsmall, small or medium) in particular docs pages? Right now in some examples, there is only an inline@breakpoint and on the other, there are all

danielamorse (changes requested):
@MarcinMr the List element looks and works great. I have some feedback on the docs and a few issues to discuss...

I made a number of minor copy edits. In general, here's what I did:
- I shortened a few of the "Important Notes".
- I removed some mentions of "List Element" and replaced with just "List". I only kept "element" when we make a distinction between List element and List component.
- I lowercased "element" which is how I see it elsewhere in the docs.
- I updated docs page titles, removing "Variations".

**Requested changes**
- Review the copy edits and merge if you approve: `feature/DS-833-List-Element--edits`. Let's discuss if you have any questions or I got something wrong.
- Please remove "Variations" from the demo file names and update the links to match. The "Variations" convention is legacy and our new elements/components don't use it.

**Comments**
- Per our convention, I think we should change the prop `items` to `content`, as it is a blob of content not an actual array of items.
- Our other elements have their default styles baked-in, i.e. you don't need to add `e-bolt-button--medium` to get a "medium" (default) button. This makes it easier to use plain HTML, as there are fewer classes required for default styles. I noticed the Twig is setting default classes for things like `spacing` and `alignment` and maybe some others. I think we should make List element like Button element and set default styles in the CSS rather than relying on additional classes. Classes can be used for non-default options.

> @remydenton would you agree with the above two points?

Lastly, just a reminder that the List element adds ~40k of new CSS. @colbytcook we should retire the List component ASAP to avoid further CSS bloat.

danielamorse:
> I'm wondering if we should have examples of all horizontal@breakpoint (xxsmall, xsmall, small or medium) in particular docs pages? Right now in some examples, there is only an horizontal@medium and on the other, there are all

@MarcinMr I don't think we have to show every breakpoint in each demo. IMO the docs should include the minimum number of examples needed to show the prop behavior. If necessary, you can make a test page with all prop variations for developer testing.

MarcinMr:
@danielamorse 

I merged your `--test` copy and did your change requests. Added one test page in the test folder with all display options.

Waiting for the @remydenton approval of the comments Dan wrote above.

@colbytcook  Should the list component be marked as deprecated?

remydenton:
I agree with both points @danielamorse, thanks for noticing those.  Specifically:

- We have docs about `items` vs `content`: https://github.com/boltdesignsystem/bolt/wiki/Standard-Props-for-Passing-Content-in-Twig
- Elements can be used as plain HTML (without twig), so having baked-in defaults is important to make them actually function as defaults.

MarcinMr:
> I agree with both points @danielamorse, thanks for noticing those. Specifically:
> 
> * We have docs about `items` vs `content`: https://github.com/boltdesignsystem/bolt/wiki/Standard-Props-for-Passing-Content-in-Twig
> * Elements can be used as plain HTML (without twig), so having baked-in defaults is important to make them actually function as defaults.

Thanks Remy, I'll do that :)

MarcinMr:
Change requests and two suggestions from the comments above are made

packages/elements/bolt-list/__tests__/list.js
```diff
@@ -0,0 +1,153 @@
...
+    expect(results.ok).toBe(true);
+    expect(results.html).toMatchSnapshot();
+  });
+});
+
+describe('Bolt List prop -', () => {
+  // Target each of the schema keys with the following pattern
+  tag.enum.forEach(async option => {
+    test(`tag: ${option}`, async () => {
+      const results = await render('@bolt-components-list/list.twig', {
```
danielamorse, line 47: @MarcinMr I missed this the first time - would you please update these tests to use the List element instead of the component?

packages/elements/bolt-list/src/list.scss
```diff
@@ -0,0 +1,395 @@
...
+   List
+\* ------------------------------------ */
+
+@import '@bolt/core-v3.x';
+
+/**
+ * List
+ *
+ * Dev notes:
+ * 1. Display prop.
```
danielamorse, line 11: Nitpick... If you're just labelling a section of SCSS, e.g. "Display prop", I think it's fine to use `// Display prop` above the section in the code. I usually see the numbered "Dev Notes" used for explaining code rather than a section label. [The Type element](https://github.com/boltdesignsystem/bolt/blob/master/packages/elements/bolt-type/src/type.scss) is a good reference.

That said, our comments are a little inconsistent throughout the codebase. I appreciate the attention to detail.

danielamorse (changes requested):
@MarcinMr thanks for addressing the feedback. I left a couple more comments. The main one is about updating the tests. Once the tests are updated, this is good to go 👌

packages/elements/bolt-list/__tests__/list.js
MarcinMr: Ah... I missed that too :) It should be ok now

packages/elements/bolt-list/src/list.scss
MarcinMr: I wasn't sure if should use "numbered comments" or "slash comments", this is why I decided to change them finally to "numbered comments". I update them now according to your feedback above :)

MarcinMr: Made change requests according Dan's feedback

packages/elements/bolt-list/__tests__/list.js
danielamorse: @MarcinMr I made one more update to the tests so that they use the List Item template: https://github.com/boltdesignsystem/bolt/pull/2550/commits/952e32fd22a1341feb1235fc4cf515198c210e9c

danielamorse:
@MarcinMr @remydenton @colbytcook one more thing (maybe the _actual_ last thing 😅)...

I'm thinking we should not require a classname on the List Items so that the HTML version of this element is super simple.

Before:
```
<ul class="e-bolt-list">
  <li class="e-bolt-list__item">
    This is a list item
  </li>
</ul>
```
After:
```
<ul class="e-bolt-list">
  <li>
    This is a list item
  </li>
</ul>
```

This would be similar to what we did for the Table element, which also does not require classnames on the `<tr>`, `<td>`, etc.

colbytcook:
@danielamorse I think this is correct way to do this, that way the display modifications that we provide can be managed with a single class addition to the `ol` or `ul`

danielamorse:
Ok, thanks for the feedback @colbytcook @remydenton re: https://github.com/boltdesignsystem/bolt/pull/2550#issuecomment-1307614568

@MarcinMr would you please go ahead and remove this class, update the CSS, and then update the test snapshots?

MarcinMr:
> Ok, thanks for the feedback @colbytcook @remydenton re: [#2550 (comment)](https://github.com/boltdesignsystem/bolt/pull/2550#issuecomment-1307614568)
> 
> @MarcinMr would you please go ahead and remove this class, update the CSS, and then update the test snapshots?

@danielamorse Updated item template, CSS, and snapshots

PR #2559, opened 2022-10-18
DS-835: Table Sticky Top Bugfix

colbytcook:
## Summary

Fixed a bug where the sticky top header would go behind the rows.

## Details

Removed a unneeded CSS on the `tr` that was causing this issue.

## How to test

Pull down the branch
Navigate to the `/pattern-lab/?p=tests-table-support-center` page
Scroll down on the table and observe that the top header is sticking

### DIFF
--- a/packages/elements/bolt-table/src/table.scss
+++ b/packages/elements/bolt-table/src/table.scss
@@ -112,10 +112,6 @@
       td {
         background-color: var(--m-bolt-table-highlight);
       }
-
-      th {
-        z-index: 1;
-      }
     }
   }
 
@@ -182,7 +178,6 @@
   thead th {
     position: sticky;
     top: 1px;
-    z-index: 2;
     box-shadow: 0 0 0 1px var(--m-bolt-table-border);
   }
 
--- a/packages/elements/bolt-table/src/table.scss
+++ b/packages/elements/bolt-table/src/table.scss
@@ -97,22 +97,17 @@
   tbody,
   tfoot {
     th:first-of-type {
-      position: relative;
-
-      &:before {
-        content: '';
-        opacity: 0;
-        position: absolute;
-        top: 0;
-        bottom: 0;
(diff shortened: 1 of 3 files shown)


packages/elements/bolt-table/src/table.scss
```diff
@@ -101,6 +97,8 @@
   tbody,
   tfoot {
     th:first-of-type {
+      position: relative;
```
danielamorse, line 100: This rule breaks the stickiness of the side headers.

packages/elements/bolt-table/src/table.scss
```diff
@@ -194,6 +191,7 @@
   thead th {
     position: sticky;
     top: 1px;
+    z-index: 2;
```
danielamorse, line 194: `z-index: 1` seems to work for me once I remove the `position: relative` rule above. Does that work for you or am I missing something?

danielamorse (changes requested):

packages/elements/bolt-table/src/table.scss
colbytcook: This is now removed

packages/elements/bolt-table/src/table.scss
colbytcook: The z-indexes are now removed

colbytcook:
@danielamorse @MarcinMr I added the `min-width :0` to the `.l-bolt-site > *:only-child` selector to fix the bug. Can you please review again?

MarcinMr (approved):
Seems like scrolling behavior works for the table

packages/elements/bolt-table/src/table.scss
danielamorse: Cool, I was saying just to remove this one but I actually don't see any bugs with them both removed.

danielamorse (approved):

PR #2565, opened 2022-10-20
[6.x] DS-872 Add bundle analyzer to build

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-872

## Summary

Add bundle analyzer to the build to monitor bundle sizes.

## Details

[Bundle analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) shows bundle sizes of the build in a list view and as an interactive visual tree.

Bundle analyzer can run in both dev and prod modes. Prod is most useful because it shows us the actual bundle sizes we ship. Unfortunately I was unable to "watch" in prod mode due to some build issues I do not have time to debug. So for now you can only watch in "dev" mode and build in "prod" mode.

This PR also adds `compiler.close()` to `packages/build-tools/tasks/webpack-tasks.js`. This is [recommended in the docs](https://webpack.js.org/api/node/#run). I added it while attempting to watch in "prod" mode - that didn't work but closing the compiler is a good idea regardless.

## How to test

- Run `yarn` to install bundle analyzer.
- Run `yarn analyze` to watch the files in dev mode. Verify the Bundle analyzer window pops up.
- Run `yarn analyze:prod` to build the files in prod mode. Verify the same and that the bundles are smaller, as they should be in prod mode.

### Release Notes
- Adds `yarn analyze` and `yarn analyze:prod` commands to the build to enable [Webpack bundle analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer).

### DIFF
--- a/packages/build-tools/tasks/webpack-tasks.js
+++ b/packages/build-tools/tasks/webpack-tasks.js
@@ -34,6 +34,7 @@ async function compile(customWebpackConfig) {
   return new Promise((resolve, reject) => {
     const compiler = boltWebpackMessages(webpack(webpackConfig));
     compiler.run((err, stats) => {
+      compiler.close();
       if (err) {
         return reject(err);
       } else {
--- a/packages/build-tools/tasks/webpack-tasks.js
+++ b/packages/build-tools/tasks/webpack-tasks.js
@@ -34,7 +34,6 @@ async function compile(customWebpackConfig) {
   return new Promise((resolve, reject) => {
     const compiler = boltWebpackMessages(webpack(webpackConfig));
     compiler.run((err, stats) => {
-      compiler.close();
       if (err) {
         return reject(err);
       } else {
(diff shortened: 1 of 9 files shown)


packages/build-tools/tasks/webpack-tasks.js
```diff
@@ -34,6 +34,7 @@ async function compile(customWebpackConfig) {
   return new Promise((resolve, reject) => {
     const compiler = boltWebpackMessages(webpack(webpackConfig));
     compiler.run((err, stats) => {
+      compiler.close();
```
colbytcook, line 37: @danielamorse why are we adding a `compiler.close()` here?

danielamorse: @colbytcook good catch. I should've called that out in the PR. I added a note about it and an inline comment.

> This PR also adds `compiler.close()` to `packages/build-tools/tasks/webpack-tasks.js`. This is [recommended in the docs](https://webpack.js.org/api/node/#run). I added it while attempting to watch in "prod" mode - that didn't work but closing the compiler is a good idea regardless.

colbytcook (approved):
@danielamorse approved

PR #2566, opened 2022-11-03
[6.x] DS-823 restore radial gradient

remydenton:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-823

## Summary

Restore and build out background radial gradient

## Details

This PR adds back `radial-gradient` as an option for `fill` in the background component.

A summary of the updates:
1. Added full support for the `focal_point` prop (across both horizontal and vertical axes), including full documentation.
2. Made `radial-gradient` fill always full opacity.  It looks like crap when used with any transparent opacity options, and would probably never have been used this way in any designs.
3. Integrated `radial-gradient` with the theming system.
  a. `radial-gradient` changes colors based on the the `t-bolt-*` class of the closest parent.
  b. `--m-bolt-bg-highlight` and `--m-bolt-bg-lowlight` CSS variables have been added to each theme.  These two variables indicate the two colors that should be used in a background gradient within this theme (based on design direction) and could be re-used in the future.
  c. Documentation for theme compatibility of both `gradient` and `radial-gradient` fill options can be seen at /pattern-lab/?p=components-background-theme-compatibility
  d. Documentation for all possible colors can be seen at /pattern-lab/?p=tests-background-themes

Specific questions for reviewers:
1. (easier question) Should we combine the two pages for background theme compatibility?  I had originally intended to keep the main doc page (/pattern-lab/?p=components-background-theme-compatibility) lean, but it honestly just feels more confusing having two pages.  My latest thought is to just keep the sizes as small as reasonable, then group the different fill types in different tabs, following the example of https://boltdesignsystem.com/pattern-lab/?p=layouts-layout-template.
2. (harder question) After building to this point, it feels kind of wrong to have `radial-gradient` as a background `fill` option.  Specifically:
- It is not like the other fill options, which are actually _overlay_ fill options.
- A background gradient could theoretically be applicable anywhere there is a background color (such as cards/teasers-- this is not just theoretical, but something Matt specifically requested).

I have an alternative to propose: Instead of the `fill` prop, we could use one class to add a gradient background to _any_ element, and a second class to modify the gradient focal point.  Something like the following:
```
<div class="t-bolt-navy u-bolt-gradient u-bolt-gradient--top-right">Hello world</div>
```
```
.u-bolt-gradient--top-right {
  --bolt-radial-gradient-focus: 100% 0%;
}
...

.u-bolt-gradient {
  background-repeat: no-repeat;
  background-size: 100% 100%;
  background-image:
    radial-gradient(
      circle at var(--bolt-radial-gradient-focus),
      var(--m-bolt-bg-highlight) 20%,
      var(--m-bolt-bg-lowlight) 80%
    );
}
```

## How to test

- Most simply, create a background with `fill: radial-gradient`.  A basic example can be seen at /pattern-lab/?p=components-background-fill
- For the new `focal_point` options, see /pattern-lab/?p=components-background-focal-point.
- For theme compatibility, see the documentation at /pattern-lab/?p=components-background-theme-compatibility and /pattern-lab/?p=tests-background-themes should make it clear.

## Release notes

- Added full support for the `focal_point` prop
- Added a gradient and gradient focus utility class that works with the Bolt Theming system
- Added `--m-bolt-bg-highlight` and `--m-bolt-bg-lowlight` CSS variables to the Bolt themes

### Visual changes

- Existing Background Comopoent gradients will display as navy instead of black


MarcinMr (commented):
@remydenton 

I tested backgrounds with `solid`, `gradient`, and `radial-gradient`, and seems like everything works great.

Also tested all variations of `focal_point` and it behaves as expected.

Applying theme class to the parent also works.

I can agree with what you wrote in point two about transparent `radial-gradient`.

For example, when `focal_point` is `right-left` I can barely say if this is a `gradient` or a `radial-gradient`. 

The only variation I can notice in a transparent `radial gradient` is when `focal_point` is set to `center-center` and the `opacity` is `heavy`. Below heavy is not easy to notice the radial for me.

When it comes to questions...

1. The idea of having tabs seems like a good concept to group everything nicely.

2. The second question... I cannot decide what would be better. Since now gradients work very well. I think the main factor to decide if we should create these "utility gradient classes" is if they will be used anywhere there is a background color. As you wrote, Matt requested that, so maybe we should?

danielamorse (approved):
1. Tabs would be nice but not necessary. I _would_ add a radial-gradient example to the Twig demo code on the background themes compatibility page so that both solid and radial-gradient options are shown in code. I just found it a little confusing to only see one of them.

2. I like the radial gradient utility class idea from a code-reuse standpoint. Would an author need to know the utility classname or would that be abstracted somehow in the layout builder UI? I'd be fine with this shipping as is or refactoring later into a utility class.

packages/components/bolt-background/src/background.scss
```diff
@@ -269,36 +245,70 @@ bolt-background {
   background-color: transparent !important;
 }
 
-/* DEPRECATED. Radial gradient is deprecated. Commented for future reference */
+:root,
```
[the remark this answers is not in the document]
remydenton: I left `:root` here so that these two CSS variables would never be undefined when they get used below.

I suppose it's redundant with setting a schema default for the focal point.  It does seem like a nice safety check to avoid undefined CSS variable bugs in the future though, and matches what's being done for [the themes in _vars-mode.scss](https://github.com/boltdesignsystem/bolt/pull/2566/files/e3cb2da5c8c542e66e94828257613f585c318901#diff-68acd9d490eb66a0f6a82a67c6c3fe1d1df233177fff6ef35f3c9422bb890709R48) and elsewhere as far as I know.

remydenton:
Thanks for the feedback guys.  I am going to try to update this to use the utility classes for radial gradient-- it is part way done already, and really does feel like a better way the more I think about it.  To your question @danielamorse, the plan would be to abstract it in the admin UI behind a checkbox or select list or similar.

packages/components/bolt-background/src/background-overlay.twig
```diff
@@ -9,13 +9,17 @@
 {% set this = init(schema) %}
 {% set attributes = create_attribute({}) %}
 
-{% set _is_reversed = focal_point and focal_point.horizontal == 'left' and this.data.fill.value == 'gradient' %}
-{% set fill_value = _is_reversed ? 'gradient-reversed' : this.data.fill.value %}
+{# Validate focal_point manually since it is an object. #}
+{% set schema_horizontal = schema.properties.focal_point.properties.horizontal %}
+{% set schema_vertical = schema.properties.focal_point.properties.vertical %}
```
colbytcook, line 14: @remydenton I am going to rename this `focal_point_horizontal` and `focal_point_vertical` since that would be morre accurate of the data it is storing

colbytcook (approved):

packages/components/bolt-background/src/background-overlay.twig
remydenton: Great choice, thanks @colbytcook

PR #2568, opened 2022-11-04
DS-887: Remove ja/en lang logic

danielamorse:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-887

## Summary

Remove Japanese/English language logic from Bolt.

## Details

Bolt used to output separate files for Japanese and English webpages. These files controlled the spacing and font-families used for each. There were also configuration options for the webpack build that handled languages. All of this is no longer used as language differences are [handled by CSS vars](https://github.com/boltdesignsystem/bolt/blob/master/packages/global/styles/00-vars/_vars-japanese.scss).

This PR removes all remnants of the original language logic from Bolt. It leaves the CSS vars in place.

## How to test

- Review code.
- Run this branch locally and verify the Japanese homepage template looks the same [on prod](https://boltdesignsystem.com/pattern-lab/patterns/999-archive-homepage-homepage-japanese/999-archive-homepage-homepage-japanese.html) as it does locally.

## Release notes

- The `$bolt-lang` SCSS variable has been removed.
- The `lang` and `i18n` boltrc config options have been removed.
- None of these options are needed anymore.


adamszalapski (approved):
I compared both pages and they look identical. I can't find any occurrences of `i18n` and `lang` in the code base. Great job, @danielamorse

packages/build-tools/create-webpack-config.js
```diff
@@ -358,18 +345,16 @@ async function createWebpackConfig(buildConfig) {
     return globalJsData;
   }
 
-  const webpackConfig = merge(sharedWebpackConfig, {
-    entry: await buildWebpackEntry(true),
```
colbytcook, line 362: @danielamorse quick question, is this argument removal just a clean up task? `buildWebpackEntry` doesn't seem to have any paramters (want to make sure I am not missing anything).

colbytcook (approved):

remydenton (approved):
Looks great!

Regarding the Drupal side of things, I _think_ it will still continue to function (using the non-language js and css) even after this update.  But it's obviously best if we can clean it up there as well ASAP.  Which reminds me, I have some work to do!

packages/build-tools/create-webpack-config.js
danielamorse: Yes, nice catch. It's just clean up. The argument didn't do anything. Maybe it once did but it doesn't anymore.

PR #2577, opened 2022-11-23
DS-774: Close Popover when inner action is clicked

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-774](https://pegadigitalit.atlassian.net/browse/DS-774)

## Summary

Add event listener to close popover on click for anchors and buttons.

## Details

Add event listener to close popover on click for anchors and buttons.

## How to test

Check the test page in browser and verify the popover closes when clicking on a link or a button.

### DIFF
--- a/packages/components/bolt-popover/src/popover.js
+++ b/packages/components/bolt-popover/src/popover.js
@@ -170,6 +170,13 @@ class BoltPopover extends BoltElement {
     this.initTippy();
     this.setAttribute('ready', '');
 
+    // close popover when inner action is clicked
+    for (const innerAction of this.content.querySelectorAll('a, button')) {
+      innerAction.addEventListener('click', () => {
+        this.popover?.hide();
+      });
+    }
+
     // Call super and dispatch "ready" event _after_ Tippy is setup
     super.firstUpdated && super.firstUpdated();
   }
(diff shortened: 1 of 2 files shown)


packages/components/bolt-popover/src/popover.js
```diff
@@ -170,6 +170,13 @@ class BoltPopover extends BoltElement {
     this.initTippy();
     this.setAttribute('ready', '');
 
+    // close popover when inner action is clicked
+    for (const innerAction of this.content.querySelectorAll('a, button')) {
```
danielamorse, line 174: Interesting... I pretty much always use `forEach` and seeing your `for...of` led me to read up on the differences.

There are times when you definitely want to use `for...of`, like when you want to `break` early, use `await`, or loop over a huge data set. Otherwise, [the performance](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/) optimization of `for...of` seems to be negligible -- though `for...of` _is_ faster.

I go with `forEach` because it's more readable to me, but I could get used to `for..of`. I may start using that more.

To the team: any thoughts on best practices around `forEach` vs. `for...of`?

danielamorse (approved):
@cjwhitedev tested locally and works as expected. Nice work!

packages/components/bolt-popover/src/popover.js
colbytcook: @danielamorse I don't really have an opinion either way. I mean to be fair I still like my classic `for` loop so I don't count.

colbytcook (approved):

PR #2580, opened 2022-12-01
DS-868: Add description_clamp prop to Listing Teaser

cjwhitedev:
## Jira

[https://pegadigitalit.atlassian.net/browse/DS-868](https://pegadigitalit.atlassian.net/browse/DS-868)

## Summary

Add a new prop for Listing Teaser to toggle the line clamp on the description.

## Details

Add a new prop for Listing Teaser to toggle the line clamp on the description.

## How to test

Check out the new demo page.


danielamorse (approved):
Looks good 👍  Approving with one comment...

At first, I thought we should call it "description_truncation" because most people don't know the term "clamp". On second thought, I like "clamp" because it's more accurate. It clamps to a number of lines - it doesn't truncate a number of words. It's probably good to familiarize users with this advanced feature by name.

So, I'm good with it as is, open to alternatives if you can think of anything better.

colbytcook (approved):

PR #2586, opened 2022-12-07
DS-902: Icon in the menu comp doesn't display in proper place

MarcinMr:
## Jira

https://pegadigitalit.atlassian.net/browse/DS-902

## Summary

When the `icon_after` prop is used, the icon displays after the menu's text.

## Details

- The `_menu-item.twig` template was updated to make the `icon_after` prop work as expected
- The `_menu-item.scss` was updated to remove unnecessary grid gap at the end of the menu item
- The doc page **menu use case icons** was updated with the `icon_after` examples

## How to test

Pull the branch. Go to `pattern-lab/?p=components-menu-use-case-icons` and confirm that icons display properly before and after the text.

## Release notes

The bug in the menu component was fixed to make the icon display properly after the menu's text.


MarcinMr:
The only thing I'm wondering is which example is better for displaying the `icon_after` variant? Should the icon be aligned to the right like in the **menu item 3** or it should stuck to the text like in the **menu item 2**?

icon after examples

colbytcook (approved):
@MarcinMr approved

remydenton (approved):
Per Matt via Slack:
> Item 3 hands down

Thanks for the fix @MarcinMr.

